PR #685, opened 2014-12-18
Allow drag-and-drop of charms/bundles

makyo:
The inability to drag-and-drop charms/bundles from the sidebar was simply a side-effect of the switch to the newer data structures from charmworld to the charmstore.  These have been updated accordingly.


makyo:
Will reopen when complete

PR #695, opened 2015-02-17
Add apiv4 bundle id support to deploy-target

hatched:
The jujucharms.com website allows users to "add to demo" bundles to play around with. It has had to guess at the apiv3 bundle id for some time now which has caused a number of bugs. This adds apiv4 bundle support to the GUI's deploy-target handler so that jujucharms.com can now use the apiv4 bundle id's.

### DIFF
--- a/app/store/charmstore-api.js
+++ b/app/store/charmstore-api.js
@@ -345,6 +345,43 @@ YUI.add('charmstore-api', function(Y) {
           this.apiPath, charmId, 'icon.svg'].join('/');
       }
       return path;
+    },
+
+    /**
+      Takes the bundle id and fetches the bundle YAML contents. Required for
+      deploying a bundle via the deployer.
+
+      @method getBundleYAML
+      @param {String} id Bundle id in apiv4 format.
+      @param {Function} successCallback The success callback.
+      @param {Function} failureCallback The failure callback.
+    */
+    getBundleYAML: function(id, successCallback, failureCallback) {
+      this.getEntity(
+          id, this._getBundleYAMLResponse.bind(
+              this, successCallback, failureCallback), failureCallback);
+    },
+
+    /**
+      getEntity success response handler which grabs the deployerFileUrl from
+      the recieved bundle details and requests the YAML.
+
+      @method _getBundleYAMLResponse
+      @param {Function} successCallback The success callback.
+      @param {Function} failureCallback The failure callback.
+      @param {Array} bundle An array containing the requested bundle model.
+    */
+    _getBundleYAMLResponse: function(successCallback, failureCallback, bundle) {
+      // XXX Jeff 17-02-15 The deployerFileUrl is generated after the response
+      // in the _processEntityQueryData method. Once the deployer supports the
+      // new bundle format this can be updated to grab the real yaml not the
+      // 'orig' version.
+      this._makeRequest(
+          bundle[0].get('deployerFileUrl'),
+          function(resp) {
+            successCallback(resp.currentTarget.responseText);
+          },
+          failureCallback);
     }
   };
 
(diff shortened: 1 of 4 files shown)


hatched:
#### To QA
`make devel`
visit `<host>?deploy-target=bundle/django`
It should deploy the bundle.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/2372/
Test FAILed.

hatched:
Spurious failure, running manually.

makyo:
:+1: QA okay - works like a dream, tried a few different bundles.

hatched:
Passed when ran manually http://ci.jujugui.org:8080/job/juju-gui/2373/

app/store/charmstore-api.js
```diff
@@ -345,6 +345,43 @@ YUI.add('charmstore-api', function(Y) {
...
       }
       return path;
+    },
+
+    /**
+      Takes the bundle id and fetches the bundle YAML contents. Required for
+      deploying a bundle via the deployer.
+
+      @method getBundleYAML
+      @param {String} id Bundle id in apiv4 format.
```
jaycee: I might make it clear we're talking charmstore here, e.g. `... id in v4 charmstore api format`.

app/store/charmstore-api.js
```diff
@@ -345,6 +345,43 @@ YUI.add('charmstore-api', function(Y) {
...
+    /**
+      getEntity success response handler which grabs the deployerFileUrl from
+      the recieved bundle details and requests the YAML.
+
+      @method _getBundleYAMLResponse
+      @param {Function} successCallback The success callback.
+      @param {Function} failureCallback The failure callback.
+      @param {Array} bundle An array containing the requested bundle model.
+    */
+    _getBundleYAMLResponse: function(successCallback, failureCallback, bundle) {
```
jaycee: Aren't callbacks usually the last arguments, except maybe context/scope? This reads weirdly to me. I feel like it should be `_getBundleYAMLResponse: function(bundle, success, failure)`.

app/store/charmstore-api.js
```diff
@@ -345,6 +345,43 @@ YUI.add('charmstore-api', function(Y) {
...
+
+      @method getBundleYAML
+      @param {String} id Bundle id in apiv4 format.
+      @param {Function} successCallback The success callback.
+      @param {Function} failureCallback The failure callback.
+    */
+    getBundleYAML: function(id, successCallback, failureCallback) {
+      this.getEntity(
+          id, this._getBundleYAMLResponse.bind(
+              this, successCallback, failureCallback), failureCallback);
```
jaycee: I'm not sure this call is right, based on the def below. It looks like your passing in context first, the callbacks, and no bundle at all to the `.bind` call on `_getBundleYAMLResponse`. Am I misreading this?

app/store/charmstore-api.js
hatched: Good idea

app/store/charmstore-api.js
hatched: `bind` returns a function which executes in the context of the object passed as the first argument. The call signature is `bind(context, [arg [, arg[, ...])`

app/store/charmstore-api.js
hatched: The arguments which are passed to a bound function are passed first to the function, the function is essentially being curried. This is why when `getEntity` calls the `_getBundleYAMLResponse` method the arguments it passes to it get appended after the two callbacks.

jaycee:
:+1: @hatched, thanks for the explanations.

hatched:

jujugui:

PR #701, opened 2015-03-06
Fix bug #1428751: prevent incorrect lowercasing.

kadams54:
Fixes https://bugs.launchpad.net/juju-gui/+bug/1428751, which was caused by overzealous recursive lowercasing of all keys returned by the API. Some keys, such as the charm author-provided ones in the charm config, should not be lowercased. Adds a way to exclude (i.e., no lowercasing) certain levels when recursing through an object.

### DIFF
--- a/app/store/charmstore-api.js
+++ b/app/store/charmstore-api.js
@@ -143,25 +143,37 @@ YUI.add('charmstore-api', function(Y) {
       @method lowerCaseKeys
       @param {Object} obj The source object with the uppercase keys.
       @param {Object} host The host object in which the keys will be assigned.
+      @param {Object} exclude Exclude a particular level from lowercasing when
+        recursing.
       @return {Undefined} Does not return a value, modifies the supplied host
         object in place.
     */
-    _lowerCaseKeys: function(obj, host) {
+    _lowerCaseKeys: function(obj, host, exclude) {
       if (!obj) {
         return;
       }
       Object.keys(obj).forEach(function(key) {
-        host[key.toLowerCase()] =
+        // An exclude of 0 means "don't lowercase this level".
+        var newKey = key;
+        if (exclude !== 0) {
+          newKey = key.toLowerCase();
+        }
+        host[newKey] =
             (typeof obj[key] === 'object' && obj[key] !== null) ?
                 Y.merge(obj[key]) : obj[key];
         if (typeof obj[key] === 'object' && obj[key] !== null) {
-          this._lowerCaseKeys(host[key.toLowerCase()], host[key.toLowerCase()]);
+          // Decrement exclude by one if it exists.
+          var newExclude = exclude;
+          if (newExclude !== undefined) {
+            newExclude = exclude - 1;
+          }
+          this._lowerCaseKeys(host[newKey], host[newKey], newExclude);
         }
         // This technique will create a version with a capitalized key so we
         // need to delete it from the host object. To protect against keys
         // which are already lower case then we test to make sure we don't
         // delete those.
-        if (key.toLowerCase() !== key) {
+        if (newKey !== key) {
           delete host[key];
         }
       }, this);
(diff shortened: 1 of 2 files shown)


kadams54:
## QA

On the `develop` branch, in your sandbox:

1. Drag-and-drop the Mongo charm. Don't commit it.
2. In the console, run these commands and verify their output:
~~~~
> window.app.db.services.item(0).get('config').opidmem
<- "default"
> window.app.db.services.item(0).get('config').opIdMem
<- undefined
~~~~

That's the reverse of what it should be. `opIdMem` should be defined and set to "default". Checkout this PR and then repeat the steps above. This time you should see:

~~~~
> window.app.db.services.item(0).get('config').opidmem
<- undefined
> window.app.db.services.item(0).get('config').opIdMem
<- "default"
~~~~

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/2388/
Test FAILed.

makyo:
:+1: no QA yet - may need to investigate the saucelabs failure

kadams54:
The SauceLabs failure seems to be spurious - it's a notifier widget test failure.

app/store/charmstore-api.js
```diff
@@ -143,25 +143,37 @@ YUI.add('charmstore-api', function(Y) {
...
         object in place.
     */
-    _lowerCaseKeys: function(obj, host) {
+    _lowerCaseKeys: function(obj, host, exclude) {
       if (!obj) {
         return;
       }
       Object.keys(obj).forEach(function(key) {
-        host[key.toLowerCase()] =
+        // An exclude of 0 means "don't lowercase this level".
```
bac: This definition should be part of the parameter explanation above.

test/test_charmstore_apiv4.js
```diff
@@ -166,6 +166,13 @@ describe('Charmstore API v4', function() {
       charmstore._lowerCaseKeys(uppercase, host);
       assert.deepEqual(host, { baz: '1', foo: { bar: { baz: '1'}}});
     });
+
+    it('can skip one level of keys in an object', function() {
+      var uppercase = { Baz: '1', Foo: { Bar: { Baz: '1' }}, Fee: '2'};
+      var host = {};
+      charmstore._lowerCaseKeys(uppercase, host, 0);
```
bac: I'm surprised by this usage. I expected 'exclude' to be a positive integer that counts down.

bac:
:+1: Though I find the semantics of the exclude parameter unintuitive and perhaps overly complicated. Or maybe it is brilliant.

Do you really need to be able to pick any arbitrary depth to be the one excluded from lowercasing?  If we only really need to exclude the first level, then why not send in exclude=true for it and then always recurse with exclude=false?  Seems less confusing than a counter that adds arbitrary flexibility that YAGNI.

No QA yet.

bac:
QA-:ok:

app/store/charmstore-api.js
kadams54: It isn't because that wouldn't be correct. If you pass in an exclude of 3, it would mean "exclude the fourth level down from lowercasing". As the function recurses, it decrements the exclude such that, when it hits the fourth level, `exclude == 0`. But that's from an internal perspective. From an external perspective, you specify 3 and can expect the fourth level to be excluded.

kadams54:
Yeah, the exclude parameter is not very straightforward. @mitechie was not sure if we might need to exclude other levels within the API objects from lowercasing, hence the flexibility of an integer vs. a boolean.

test/test_charmstore_apiv4.js
kadams54: It is a positive integer that counts downward, just functioning on a 0-based index. So 0 is the first level of recursion, 1 the second level down, etc.

kadams54:
Thanks all! :shipit:

jujugui:

jujugui:
        build url: http://ci.jujugui.org:8080/job/juju-gui-merge/835

jujugui:

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/2391/
Test FAILed.

PR #707, opened 2015-03-16
Fix service icon bounceback when dragging on canvas

hatched:
This fixes an issue where the service icons would bounce back to their original positions while the user was dragging them around.

### DIFF
--- a/app/views/topology/service.js
+++ b/app/views/topology/service.js
@@ -1248,7 +1248,7 @@ YUI.add('juju-topology-service', function(Y) {
             var existing = box.model.get('annotations') || {};
             if (!existing || !existing['gui-x']) {
               vertices.push([box.x || 0, box.y || 0]);
-              topo.annotateBoxPosition(box, false);
+              topo.annotateBoxPosition(box);
             } else {
               if (vertices.length > 0) {
                 vertices.push([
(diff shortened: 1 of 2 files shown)


hatched:
Note: Tests already existed for this functionality but did not exist for the timer functionality so no new tests are needed.

#### To QA
Deploy to a real environment - LXC is fine but the slower the env the better.
On the canvas, click and drag the service icon around in short bursts ~1s for 30s. If it doesn't snap back to its original position then you're ok.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/2411/
Test PASSed.

bac:
Code is :+1: but I have my machine tied up right now and cannot do QA.

makyo:
Code LGTM, QAing now.

makyo:
QA okay - deployed the GUI and NFS, dragged them around for a while, individually, then alternating.  :+1:

One thing I noticed is that, with the deltas coming back, the service health graph on an unplaced service goes up and down as I'm dragging it about in short bursts.  It's almost as if each time a delta comes back, it decides whether or not to display as unplaced.  Not for this branch and low priority, just something to keep an eye on.

hatched:
Thanks for the reviews and qa! @makyo can you create a low priority bug just so that we can track it? Thanks again!

:shipit:

jujugui:

chuckbutler:
dear @hatched, thank you for this

hatched:
@chuckbutler No Problem! :) A release with this fix (and others) will be coming soon as well.

PR #711, opened 2015-03-23
Update change version to use apiv4

hatched:
Old change-version generated the version numbers but charmstore apiv4 now has them available.

### DIFF
--- a/app/store/charmstore-api.js
+++ b/app/store/charmstore-api.js
@@ -407,6 +407,53 @@ YUI.add('charmstore-api', function(Y) {
       var bundle = jsyaml.safeLoad(bundleYAML);
       var wrapped = { 'bundle-deploy': bundle };
       return jsyaml.safeDump(wrapped);
+    },
+
+    /**
+      Gets the list of available versions of the supplied charm id.
+
+      @method getAvailableVersions
+      @param {String} charmId The charm id to fetch all of the versions for.
+      @param {Function} successCallback The success callback.
+      @param {Function} failureCallback The failure callback.
+    */
+    getAvailableVersions: function(charmId, successCallback, failureCallback) {
+      charmId = charmId.replace('cs:', '');
+      var series = charmId.split('/')[0];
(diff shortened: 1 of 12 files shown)


hatched:
#### To QA

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/2427/
Test PASSed.

app/store/charmstore-api.js
```diff
@@ -407,6 +407,53 @@ YUI.add('charmstore-api', function(Y) {
...
+      @method getAvailableVersions
+      @param {String} charmId The charm id to fetch all of the versions for.
+      @param {Function} successCallback The success callback.
+      @param {Function} failureCallback The failure callback.
+    */
+    getAvailableVersions: function(charmId, successCallback, failureCallback) {
+      charmId = charmId.replace('cs:', '');
+      var series = charmId.split('/')[0];
+      this._makeRequest(
+          this._generatePath(charmId, null, '/expand-id'),
```
huwshimi: Unless I'm misunderstanding how this function is to be used it looks like you're not actually using the result of this function call, could that be right?

huwshimi:
When doing a QA in the sandbox environment I have the following issue:
- add a service, deploy
- click "Change version"
- click "Upgrade" for any version
- in the inspector header the version won't change
- closing and opening the inspector or clicking on the service block will update the version number correctly
- now click "Upgrade" for a different version and the header will update correctly

app/store/charmstore-api.js
hatched: The return value of this function is being passed to the first argument of the `_makeRequest` function

hatched:
Thanks for the review and qa @huwshimi That bug you found also appears in trunk so I'll make a bug for a follow-up

makyo:
:+1: much simpler than before, thank you!  No QA yet.

hatched:

jujugui:

PR #714, opened 2015-03-30
React inspector spike (SPIKE) 

mitechie:
Added so comments can be made inline

### DIFF
--- a/app/store/env/go.js
+++ b/app/store/env/go.js
@@ -1166,6 +1166,29 @@ YUI.add('juju-env-go', function(Y) {
       userCallback(transformedData);
     },
 
+    setCharm: function(serviceName, charmUrl, force, callback) {
+      this._setCharm(serviceName, charmUrl, force, function(result) {
+        var db = this.get('db');
+        if (result.err) {
+          db.notifications.add({
+            title: 'Error setting charm.',
+            message: result.err,
+            level: 'error'
+          });
+          return;
+        }
+        this.get_charm(charmUrl, function(data) {
+          if (data.err) {
+            db.notifications.add({
+              title: 'Error retrieving charm.',
+              message: data.err,
+              level: 'error'
+            });
+          }
+        });
+      }.bind(this));
+    },
+
     /**
        Set a service's charm.
 
@@ -1177,7 +1200,7 @@ YUI.add('juju-env-go', function(Y) {
          operation is performed.
        @return {undefined} Sends a message to the server only.
      */
-    setCharm: function(service_name, charm_url, force, callback) {
+    _setCharm: function(service_name, charm_url, force, callback) {
       var intermediateCallback = null;
       if (callback) {
         intermediateCallback = Y.bind(this.handleSetCharm, this,
--- a/app/views/viewlet-manager.js
+++ b/app/views/viewlet-manager.js
@@ -353,7 +353,10 @@ YUI.add('juju-viewlet-manager', function(Y) {
         // In order to destroy this component when we change viewlets we need
         // to store it somewhere we'll have access to it at that stage.
         view.renderedComponent = React.render(
-            React.createElement(view, model.getAttrs()),
+            React.createElement(view, {
+              model: model.getAttrs(),
+              viewletManager: view.viewletManager.getAttrs()
+            }),
             view.wrapperElement);
       }
     },
--- a/app/views/viewlet-manager.js
+++ b/app/views/viewlet-manager.js
@@ -257,7 +257,9 @@ YUI.add('juju-viewlet-manager', function(Y) {
         should be rendered.
     */
     renderViewlet: function(view, name, model, viewContainer) {
-      if (!view.name) { view.name = name; }
+      // A React component is just a function so you cannot set the read-only
+      // name property.
+      if (!view.name && view instanceof Y.View) { view.name = name; }
       if (view.slot) { return; }
       if (view instanceof Y.View) {
         if (!view.model) {
@@ -266,13 +268,21 @@ YUI.add('juju-viewlet-manager', function(Y) {
         view.render(this.getAttrs());
         viewContainer.append(view.get('container'));
         this.databindingBind(model, view);
-      } else {
+      } else if (typeof view.render === 'function') {
+        // We only pre-render if it's a YUI view.
         var result = view.render(model, this.getAttrs());
         if (result && typeof result === 'string') {
           view.container = Y.Node.create(result);
         }
         viewContainer.append(view.container);
         this.databindingBind(model, view);
+      } else {
+        // If this is a React component we need to create a container so that it
+        // will play nice with the Y.View's and then we need to store that
+        // container with the React component constructor.
+        var wrapperNode = Y.Node.create('<div/>');
+        viewContainer.append(wrapperNode);
+        view.wrapperElement = wrapperNode.getDOMNode();
       }
 
     },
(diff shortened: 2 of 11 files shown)


jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/2435/
Test FAILed.

app/store/env/go.js
```diff
@@ -1166,6 +1166,29 @@ YUI.add('juju-env-go', function(Y) {
       userCallback(transformedData);
     },
 
+    setCharm: function(serviceName, charmUrl, force, callback) {
```
mitechie: Any comments here at all?  So we added a new api method to the Go API, that talks to the notification system which seems very intertwined. Is this pure spike code then?

app/views/viewlet-manager.js
```diff
@@ -266,13 +268,21 @@ YUI.add('juju-viewlet-manager', function(Y) {
         view.render(this.getAttrs());
         viewContainer.append(view.get('container'));
         this.databindingBind(model, view);
-      } else {
+      } else if (typeof view.render === 'function') {
```
mitechie: is there any other sniffing method vs just 'it's a function'?

app/views/viewlet-manager.js
```diff
@@ -266,13 +268,21 @@ YUI.add('juju-viewlet-manager', function(Y) {
...
         if (result && typeof result === 'string') {
           view.container = Y.Node.create(result);
         }
         viewContainer.append(view.container);
         this.databindingBind(model, view);
+      } else {
+        // If this is a React component we need to create a container so that it
+        // will play nice with the Y.View's and then we need to store that
+        // container with the React component constructor.
+        var wrapperNode = Y.Node.create('<div/>');
```
mitechie: so this is begging for some sort of Y.View->component middle helper I think. That takes the container information or something and helps to build stuff and tear it down properly.

app/views/viewlet-manager.js
```diff
@@ -266,13 +268,21 @@ YUI.add('juju-viewlet-manager', function(Y) {
...
+        viewContainer.append(wrapperNode);
+        view.wrapperElement = wrapperNode.getDOMNode();
```
mitechie: does this help us cleanup/remove this when required?

app/views/viewlet-manager.js
```diff
@@ -316,12 +326,39 @@ YUI.add('juju-viewlet-manager', function(Y) {
             // parameter needed to be added so that multiple views could
             // be rendered into a single container
             if (!viewToCheck.slot) {
-              viewToCheck.hide();
+              // If this is a React component then it doesn't have a hide
+              if (typeof viewToCheck.hide === 'function') {
```
mitechie: yea, we need something like a Y.ReactMgr that we can check so these are more explicit. Then we'd just check that if we had a ReactMgr we'd do X vs Y

app/views/viewlet-manager.js
```diff
@@ -316,12 +326,39 @@ YUI.add('juju-viewlet-manager', function(Y) {
...
+                // It's a React component but it may not be rendered so we need
+                // to check if it has a DOM node.
+                var domNode;
+                try {
+                  domNode = React.findDOMNode(viewToCheck.renderedComponent);
+                } catch (e) {
+                  // do nothing if it's not in the DOM.
+                }
+                if (domNode) {
+                  React.unmountComponentAtNode(domNode.parentNode);
```
mitechie: Y.ReactMgr.destroy() could even give a nice Y.View api so that the code can be kept the same perhaps. show/hide on the container div or the like.

app/views/viewlets/change-version-jsx.jsx
```diff
@@ -0,0 +1,84 @@
...
+          </div>
+          <ul className="unit-list">
+            {items}
+          </ul>
+        </div>
+      );
+    }
+  });
+
+  ns.ChangeVersionItem = React.createClass({
```
mitechie: does this need to be on the namespace then?

app/views/viewlets/change-version-jsx.jsx
```diff
@@ -0,0 +1,84 @@
...
+          setCharm = env.setCharm;
+      setCharm.call(env, props.model.id, upgradeTo, false);
+    },
+
+    render: function() {
+      var items = [],
+          availableVersions = this.props.model.available_versions;
+      if (availableVersions) {
+        availableVersions.forEach(function(version) {
+          items.push(<ns.ChangeVersionItem version={version} upgradeService={this._upgradeService} />)
```
mitechie: oh the horror this sets off in my brain lol. No quotes at all?

app/views/viewlets/change-version-jsx.jsx
```diff
@@ -0,0 +1,84 @@
...
+
+  var ns = Y.namespace('juju.viewlets');
+
+  ns.ChangeVersionList = React.createClass({
+    // React is not MVC use utility classes and pass them into views.
+    // If it doesn't involve the UI do not put it in the component.
+    _upgradeService: function(upgradeTo) {
+      var props = this.props,
+          env = props.viewletManager.env,
+          setCharm = env.setCharm;
```
mitechie: any chance the env could be something just passed in?

mitechie:
Really interesting spike @hatched. A couple of questions on the code itself. I think this looks like a Y.View wrapper module to help meet API expectations would be really useful or else we risk a ton of if/else checks throughout the code that could lead to some hard to debug breakages in my opinion.

I also wonder about how these would nest a bit and the future of templates (handlebars) vs this world. 

I'm not sure if I can make the call or not today but will try to catch some time to hear everyone's thoughts and see what we want to look at from here. 

Maybe the next spike or real test case for feasibility would be doing something like the charm details breakout panel.

app/store/env/go.js
hatched: Yes this is pure spike code I wasn't sure where to put it. React views can be thought of as components and because they can be destroyed and modified just as DOM elements they aren't supposed to do anything other than 'call out' to get things done. So this was just a place to put the 'controller' bits of the change-version view.

app/views/viewlet-manager.js
hatched: Yeah there are probably more robust techniques. This was just the easiest to get moving

app/views/viewlet-manager.js
hatched: Yeah I agree - components don't render a wrapper node whereas Y.Views do so this so the middle-man might be pretty trivial.

app/views/viewlet-manager.js
hatched: not in the the viewlet manager takes care of cleaning all these elements up when its destroyed - for places like the machine view this funkyness wouldn't be necessary.

app/views/viewlets/change-version-jsx.jsx
hatched: Yeah it was a toss up - the current viewlets get the `viewletManager` passed in so I wanted to keep it the same.

app/views/viewlets/change-version-jsx.jsx
hatched: Heh yeah it was also tough for me to accept at first but then I had a realization. That JSX isn't HTML it just looks similar, it's actually a DSL for React. This can just as well be written in straight JS but it's a lot more typing.

app/views/viewlets/change-version-jsx.jsx
hatched: Putting every component in this namespace makes it trivial to test. We can even test these components in nodejs without phantom. But this is level of division isn't necessary I only did it as such to illustrate how we would do things like the charm lists, machine lists etc. Where the line items have a little bit more functionality.

hatched:
I think that the easier places to convert, and places where we will see the most advantage, would be places which don't use the Viewlet Manager. Charmbrowser, Added-Services, Machine View, etc. These areas will allow us to evolve the best techniques for the data flow when using these components without having to work with the complexity of the Viewlet Manager. I was thinking I'd look at doing the added services pane with React as it has a lot to gain from this architecture.

hatched:
Closing this PR but will keep the branch around

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/2436/
Test FAILed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/2437/
Test FAILed.

PR #715, opened 2015-03-31
Add React tools to makefile

hatched:
On build compile the *.jsx to *.js files and then watch for changes.

Note there are a few bugs with the jsx tool which cause multiple compilations on change but have no negative side effects on the output.
https://github.com/facebook/react/issues/3550
https://github.com/facebook/react/issues/3551


hatched:
#### To QA

Create a `foo.jsx` file somewhere in the app/ directory and then run `make devel` It should compile it to `foo.js`. Make a change to the jsx file and it should recompile.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/2438/
Test FAILed.

install-sysdeps.sh
```diff
@@ -7,4 +7,5 @@ sudo apt-get install -y nodejs imagemagick python-sphinx python-yaml \
   python-tz python-virtualenv python-shelltoolbox python-tornado \
   python-gflags g++ xvfb git firefox
 
-sudo npm install -g jshint@2.1.3 mocha-phantomjs@3.3.1 phantomjs@1.9.2-6
+sudo npm install -g jshint@2.1.3 mocha-phantomjs@3.3.1 phantomjs@1.9.2-6 \
+  react-tools
```
jrwren: I think this breaks the charm since react-tools is not in download cache, or if it is in download cache, will the version match? Should the version be pinned like the other packages?

Makefile
```diff
@@ -25,6 +25,7 @@ JSFILES=$(shell find . -wholename './node_modules*' -prune \
 	| sort | sed -e 's/^\.\///' \
 	| grep -Ev \
 		-e '^app/assets/javascripts/d3(\.min)?\.js$$' \
+		-e '^app/assets/javascripts/react-[0-9]+\.[0-9]+\.[0-9]+(\.min)*\.js$$' \
```
makyo: Does `\d` work here, rather than `[0-9]`.  Just curious, can never remember off the top of my head.

install-sysdeps.sh
hatched: The charm contains a built rollup archive of the GUI so it doesn't actually have to build anything. This new change may break the `specify a branch` functionality in the charm however. I'll make a card to make the necessary modifications to the charm. 

It's not pinned at a version because they update the tools very rapidly and we need a few bug fixes :)

makyo:
:+1: QA :ok:  Created and modified a jsx file a few times just to see how the watcher works; everything else works as expected. though I did need to run `make sysdeps` - not surprising, just something we'll all have to do.

Makefile
hatched: It doesn't but I could make it a little different if you'd prefer

`^app/assets/javascripts/react-[0-9]+(\.[0-9]+){2}(\.min)*\.js$$`

jrwren:
:+1:

hatched:
Thanks for the reviews and QA! :shipit:

jujugui:

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/2439/
Test PASSed.

PR #721, opened 2015-04-28
Skip spurious failing simulator tests

hatched:


makyo:
:+1:  The simulator was pretty useful, so we may want to keep it around, but I agree on making our tests more representative.

hatched:
Yes I agree the simulator was very useful however I think it should live on the other side of a simulated websocket so that it actually tests the entire stack.

hatched:
Thanks for the review!

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/2471/
Test FAILed.

PR #737, opened 2015-05-18
Clean state for uncommitted bundle importer

makyo:
A counter was not reset when a changeset was completed executing, so an attempt to deploy another bundle led to nothing happening (if the counter was > the number of records) or a broken bundle (if the number of records > the counter, but the counter was still not reset).

Addresses https://bugs.launchpad.net/juju-gui/+bug/1454750 and https://bugs.launchpad.net/juju-gui/+bug/1455652

### DIFF
--- a/app/utils/bundle-importer.js
+++ b/app/utils/bundle-importer.js
@@ -232,6 +232,7 @@ YUI.add('bundle-importer', function(Y) {
           message: 'ChangeSet import complete.',
           level: 'important'
         });
+        this._dryRunIndex = -1;
         return;
       }
       this._dryRunIndex += 1;
--- a/app/views/topology/service.js
+++ b/app/views/topology/service.js
@@ -1191,6 +1191,9 @@ YUI.add('juju-topology-service', function(Y) {
       // annotations.
       var new_service_boxes = Y.Object.values(topo.service_boxes)
       .filter(function(boundingBox) {
+            // In the case where a model has been removed from the database
+            // and update runs before exit, boundingBox.model will be empty;
+            // these can automatically be ignored.
             if (boundingBox.model) {
               var annotations = boundingBox.model.get('annotations');
               return ((!Y.Lang.isNumber(boundingBox.x) &&
--- a/app/views/topology/service.js
+++ b/app/views/topology/service.js
@@ -1191,9 +1191,12 @@ YUI.add('juju-topology-service', function(Y) {
       // annotations.
       var new_service_boxes = Y.Object.values(topo.service_boxes)
       .filter(function(boundingBox) {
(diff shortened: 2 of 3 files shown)


makyo:
# QA

* `juju set juju-gui juju-gui-source="https://github.com/makyo/juju-gui.git uncommitted-bundles-ecs"`

mitechie:
:+1: code no QA.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/2528/
Test PASSed.

app/views/topology/service.js
```diff
@@ -1191,9 +1191,12 @@ YUI.add('juju-topology-service', function(Y) {
...
       .filter(function(boundingBox) {
-            var annotations = boundingBox.model.get('annotations');
-            return ((!Y.Lang.isNumber(boundingBox.x) &&
-                !(annotations && annotations['gui-x'])));
+            if (boundingBox.model) {
+              var annotations = boundingBox.model.get('annotations');
+              return ((!Y.Lang.isNumber(boundingBox.x) &&
+                  !(annotations && annotations['gui-x'])));
+            }
+            return false;
```
hatched: What's going on here? Is this the fix for #1454750 ? Could you add a comment to the code as to why this is necessary

app/utils/bundle-importer.js
```diff
@@ -232,6 +232,7 @@ YUI.add('bundle-importer', function(Y) {
           message: 'ChangeSet import complete.',
           level: 'important'
         });
+        this._dryRunIndex = -1;
```
hatched: Yay! Maybe a quick comment here as to why this is required too plz.

hatched:
:+1: No QA yet

hatched:
It took me three tries but I eventually got a QA OK on this. I don't think that these issues are related to this fix but something we should look at none the less.

Notes:
I followed the QA steps.

The first time I ended up with the following error which is quite odd as I've never seen it before:
screenshot from 2015-05-18 18 35 10

The second time I got this error, which I have seen before, maybe we should put a conditional here. Any idea why this is even possible? Some how the model is undefined when trying to get the annotations.
screenshot from 2015-05-18 18 38 31

Third time worked as advertised.

app/views/topology/service.js
makyo: This was a driveby, meant to put a GH comment on it but spaced it.  Gets rid of one of the three regularly occurring errors with the ECS (the other two you describe below)

makyo:

jujugui:

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/2532/
Test PASSed.

PR #739, opened 2015-05-19
Do not truncate the display name of a service in the model

hatched:
Fixes https://bugs.launchpad.net/juju-gui/+bug/1456748


test/test_model.js
```diff
@@ -384,7 +384,7 @@ describe('test_model.js', function() {
           shortId = '12345678901234567';
       db.services.add([{id: longId}, {id: shortId}]);
       var longName = db.services.getById(longId).get('displayName');
-      assert.equal(longName.length, 18, 'name is not trucated');
+      assert.equal(longName.length, 19, 'name is not trucated');
```
hatched: We had a bit more space to take advantage of than we were using.

undocumented
```diff
@@ -1,11 +1,11 @@
-app/app.js:238 "callback"
```
hatched: Removed an undocumented function so this is just a generated file change.

makyo:
:+1: No QA yet

hatched:
#### To QA

```
mkdir -p ~/charms/trusty
cd ~/charms/trusty 
bzr branch lp:~juju-gui/charms/trusty/juju-gui/trunk juju-gui
juju bootstrap local
juju deploy --repository=~/charms local:trusty/juju-gui
juju set juju-gui juju-gui-source="https://github.com/hatched/juju-gui.git fix-truncated-names"
juju set juju-gui sandbox=true
```

Drag a service with a long name to the canvas, say nova-open-controller.
Change it's configuration values and then 'save' them (their values don't matter)
You should notice 3 records in the deployer bar
Now commit and auto place. 
It should commit and start to deploy.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/2536/
Test PASSed.

app/views/topology/service.js
```diff
@@ -210,7 +210,10 @@ YUI.add('juju-topology-service', function(Y) {
         }
         });
 
-    node.select('.name').text(function(d) { return d.displayName; });
+    node.select('.name').text(
```
mitechie: but these are now acting on the service name vs the display name correct?

app/views/topology/service.js
```diff
@@ -1346,6 +1349,29 @@ YUI.add('juju-topology-service', function(Y) {
...
 
     /**
+      Returns a service name which is a maximum of 10 characters wide.
+
+      @method truncateServiceName
+      @param {Object} service The service model
+      @return {String} The truncated service display name.
+    */
+    truncateServiceName: function(service) {
+      var name = service.displayName;
```
mitechie: ah no, ok so this is the service. I wonder about not passing the whole model here.

mitechie: yea, I think this would be nicer as either a method on the service so you can just do d.truncateServiceName which would then use self.displayName and self.pending or that it was a helper that took the two properties name, pending as arguments.

app/views/topology/service.js
```diff
@@ -1346,6 +1349,29 @@ YUI.add('juju-topology-service', function(Y) {
...
+      if (service.pending) {
```
mitechie: does this even need to be put only when .pending? I mean if we're going to use () as a trigger for things should we just not allow names that have that from users?

app/views/viewlets/inspector-header.js
```diff
@@ -78,13 +78,13 @@ YUI.add('inspector-header-view', function(Y) {
...
+      var name = pojoModel.displayName;
       if (pojoModel.pending) {
         // Check if there is already a service using the default name to
         // trigger the name ux.
         // This regex simply removes the outer parentheses from the
-        // displayName that is set in the ghost-inspector.js updateGhostName
-        // method.  If the regex doesn't match, blow up.  It should match.
-        var name = pojoModel.displayName.match(/^\(([^)]*)\)$/)[1];
+        // displayName that might be set in the model getter.
+        name = name.replace(/^\(/, '').replace(/$\)/, '');
```
mitechie: and then this can just use the same truncate helper vs redoing the logic. Just pass the string in and get it truncated.

mitechie:
Suggestion for making the function a bit more reusable and less duplication. Let me know what you think.

app/views/topology/service.js
hatched: I also had originally thought about putting it as a method on the model but at this point it's no longer a model. Just the attribute values. I could move it to be an attribute on the model `truncatedDisplayName` which is set via the setter of the `displayName`

app/views/topology/service.js
hatched: I don't understand. We wrap the name of the service in parens to indicate on the canvas that this service is not deployed.

app/views/viewlets/inspector-header.js
hatched: See above.

mitechie: I agree making it a method doesn't work but as a generic helper it can be
write once vs the duplicate code currently.

hatched:
I'm not sure that moving the method elsewhere will help at all with dedupe as it's already deduped in the topology. The inspector header code requires that the displayName not be truncated to check for an existing service then truncated for display in the deployed inspector

hatched:
As per discussion :shipit:

jujugui:

PR #776, opened 2015-08-05
Correct asset routing.

kadams54:
Asset routing was getting swallowed up by the more generalized app routing, because the app routing came first. Switch them around to correctly handle assets.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -21,6 +21,8 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
 
   config.vm.provider :virtualbox do |vb|
     vb.memory = 1024
-    vb.name = "Juju GUI Vagrant (Release)"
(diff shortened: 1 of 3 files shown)


jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/2676/
Test FAILed.

kadams54:
# QA

Open Network Inspector and watch the front page load. All assets *not* from `api.jujucharms.com` should be loaded properly. That is, before they had a 200 code, but when you clicked on the Preview tab in the request details pane you wouldn't see anything and the image file would show as broken/missing (because it was actually HTML). Now when you click on the Preview tab you should see the image.

hatched:
:+1: QA OK Thanks for this fix. The original bugs appear to have been resolved.

kadams54:
Thanks!

PR #787, opened 2015-08-18
Fix several bits of js creation/minification.

jaycee:


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/2735/
Test PASSed.

jaycee:
QA:

Run `make clean-all modules-js`
Verify that app/modules.js does not have any files ending in -min-min.js (ignore d3.min-min.js).
Verify that app/assets/javascripts/ has -min.js files for all of the files except unscaled-pack and the d3-wrapper files.

Run `make clean-all`
Verify that jujugui/static/gui/build is deleted.

Makefile
```diff
@@ -100,9 +99,10 @@ $(GUIBUILD)/app/%-min.js: $(NODE_MODULES)
 	cp -r $(GUISRC)/app/$*.js $(GUIBUILD)/app/$*.js
 	$(NODE_MODULES)/.bin/uglifyjs --screw-ie8 $(GUISRC)/app/$*.js -o $@
 
-$(BUILT_JS_ASSETS):
+$(BUILT_JS_ASSETS): $(NODE_MODULES)
 	mkdir -p $(GUIBUILD)/app/assets
 	cp -Lr $(JS_ASSETS) $(GUIBUILD)/app/assets/
+	find $(BUILT_JS_ASSETS) -type f -name "*.js" -not -name "*d3-wrapper*" -not -name "*unscaled-pack*" | sed s/\.js$$//g | xargs -I {} node_modules/.bin/uglifyjs --screw-ie8 {}.js -o {}-min.js
```
hatched: I'm not really sure what's going on here...?

Makefile
```diff
@@ -247,7 +248,7 @@ clean-pyc:
 	find . -name '*~' -exec rm -f {} +
 
 .PHONY: clean-gui
-clean-gui-build:
+clean-gui:
```
hatched: heh yeah I had this fixed locally too :)

Makefile
jaycee: This is generating a list of all the valid js files in assets and running them through uglify. The output of the find is the list of js files, which is piped through sed as a way to get the basename and then piped to xargs to do the uglify.

You might argue that we could just generate this list above as a var, but since it's a list of built files we can't guarantee the list exists at the time the target is run and we're back to all the old ugliness we had with importing src files etc.

There are probably better ways to structure this, but they involve restructuring much of the assets dir, and I would rather do that once we actually have things working since I think it's likely to be yak shave-y.

hatched:
:+1: QA OK!

jaycee:
Thanks @hatched! :shipit:

jujugui:

PR #790, opened 2015-08-20
Add babel compiler to the make steps

hatched:
To support ES6 and JSX the babel compiler needed to be added. This builds all js outside of the assets directory before having minified files generated.

- babel added to npm deps
- unused npm deps removed
- shrinkwrap regenerated
- makefile updated to build js with babel
- js files which weren't third party assets were moved

Notes: rebase caused some issues and included 31a7ee8 again, you can ignore this file.


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/2745/
Test PASSed.

hatched:
#### To QA
```
make clean-gui gui
```
- Open up `app.js` in the build directory, the YUI definition line should be longer than 80 characters (artifact of the build process that's not in the source)
- Make sure the app still works as expected.

Makefile
```diff
@@ -87,18 +87,26 @@ venv: $(PY)
...
+# The build-js target is used to build and minify only the js files that have
+# changed since the last time they were built.
+.PHONY: build-js
+build-js: $(BUILT_RAWJSFILES) $(MIN_JS_FILES)
+
+$(GUIBUILD)/app/%-min.js: $(GUIBUILD)/app/%.js
+	@#cp -r $(GUISRC)/app/$*.js $(GUIBUILD)/app/$*.js
 	$(NODE_MODULES)/.bin/uglifyjs --screw-ie8 $(GUISRC)/app/$*.js -o $@
 
+$(GUIBUILD)/app/%.js: $(GUISRC)/app/%.js
```
jaycee: This uses NODE_MODULES, so it should depend on that as well.

jaycee:
One note about deps for a make target, but otherwise :+1: 

QA to follow.

jaycee:
QA notes:

I ran `make clean-all gui` -- it built without errors.

Running `make run` worked as well, however when I opened the gui it stuck at the spinner and the console showed the error "TypeError: Templates is undefined combo:41:4" in firefox. It does appear to work in chrome.

I did `touch jujugui/static/gui/src/app/store/notifications.js && make build-js` and saw that it only redid notifications.js.

Makefile
hatched: I removed the dependency on NODE_MODULES because when it's written in this way it's depending on that for every file which I felt was a little overkill. I wasn't able to find a way where these could be triggered without another target which didn't already require NODE_MODULES. 

Of course now I see `build-js` doesn't have it, and should :)

jaycee:
I'm also seeing this message in the ff console.

```
SyntaxError: in strict mode code, functions may be declared only at top level or immediately within another function
```

Makefile
jaycee: I think it should still have it as a dep--if it's there, nothing happens, so there's no problem. If it isn't there this target fails unless another target is run first.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/2746/
Test PASSed.

jaycee:
As we determined the issue I'm seeing now exists in pyramid-fork, so it's not a QA error on this branch. In chrome, the gui runs normally.

QA OK. Thanks @hatched!

hatched:
Thanks for the qa's! :shipit:

jujugui:

jujugui:
        build url: http://ci.jujugui.org:8080/job/juju-gui-merge/975

jujugui:

PR #791, opened 2015-08-20
Add eslint

hatched:
Eslint added to parse all pre-built js files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -219,6 +219,10 @@ test-deps: $(PY)
 lint: $(FLAKE8)
 	$(FLAKE8) jujugui
 
+.PHONY: lint-js
+lint-js:
+	$(NODE_MODULES)/.bin/eslint $(RAWJSFILES)
+
 .PHONY: test
 test: $(PYTEST)
 	$(PYTEST) -s jujugui/tests
--- a/Makefile
+++ b/Makefile
@@ -220,15 +220,15 @@ lint: $(FLAKE8)
 	$(FLAKE8) jujugui
 
 .PHONY: lint-js
-lint-js:
-	$(NODE_MODULES)/.bin/eslint $(RAWJSFILES)
+lint-js: $(NODE_MODULES)
+	$(NODE_MODULES)/.bin/eslint $(GUISRC)
 
 .PHONY: test
 test: $(PYTEST)
 	$(PYTEST) -s jujugui/tests
 
 .PHONY: check
-check: clean-pyc lint test
+check: clean-pyc lint lint-js test
 
 # ci-check is the target run by CI.
 .PHONY: ci-check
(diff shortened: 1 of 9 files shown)


hatched:
#### To QA

```
make gui lint-js
```

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/2748/
Test PASSed.

jaycee:
Looks good. Can you create a lint-all target to do js and Python and update the check target to use the lint all?

jaycee:
:+1: w/ the requested change.

Makefile
```diff
@@ -219,6 +219,10 @@ test-deps: $(PY)
 lint: $(FLAKE8)
 	$(FLAKE8) jujugui
 
+.PHONY: lint-js
+lint-js:
```
jaycee: Let's make $(NODE_MODULES) a pre-req for this. We're linting source files, not built files, so we shouldn't have to build the gui to run this.

jaycee:
Playing around with this some more this morning, I'm curious why we're using $(RAWJSFILES). I had thought it was so we only linted files that have changed, but when I edited one file to introduce a lint error eslint was still run against everything. It's fast enough that I'm fine with that, but it does produce *massive* output blocks.

How about changing the lint command to `node_modules/.bin/eslint $(GUISRC)` and adding a .eslintignore file the app/assets path? The effect appears to be the same and the output from the command is much smaller--one line echoed by make and then only the lint output.

package.json
```diff
@@ -13,6 +13,7 @@
     "chai": ">=1.2.0",
     "cryptojs": ">= 2.5.3",
     "d3": "3.2.7",
+    "eslint": "^1.2.0",
```
bac: What's the caret do?

Makefile
bac: I don't understand that request.

bac:
:+1:

Makefile
hatched: done

package.json
hatched: The caret means that that is essentially `1.*.*`

In our case it doesn't really matter because it's shrinkwrapped so it'll always install 1.2.0 anyways but I don't really see any need to restrict this package when we update the shrinkwrap.

hatched:
@jaycee I made the requested changes, much better output now. Good idea.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/2749/
Test FAILed.

jaycee:
@hatched Awesome, thanks. :+1: 

Looks like the test failure is an npm install fail--likely spurious. Yay.

hatched:
CI failure was an npm spurious error. Passed successfully http://ci.jujugui.org:8080/job/juju-gui/2750/console

Thanks for the reviews and qa's! :shipit:

jujugui:

PR #796, opened 2015-08-25
Remove the crosshatched background of the canvas

anthonydillon:
Replaced with the new #f5f5f5 flat background color.


jujugui:
please test this

hatched:
I feel like this branch should be done against `pyramid-fork` because it's not really a required fix for the develop branch. It will also save us rebasing pyramid-fork. 

Also, I don't see any of the assets being removed.

makyo:
The QA is okay once the app gets up and running, but I agree with @hatched that this should be done against the pyramid fork because the modifications to `app/index.html` won't make it over to `jujugui/templates/index.html.mako` automatically, leading to the crosshatch showing up on load.

anthonydillon:
Thank you both. Should I close this PR and only work on the `pyramid-fork` branch? Good catch @hatched I'll remove the asset.

makyo:
@anthonydillon Yes, I think that'd be easiest.

PR #814, opened 2015-09-01
Base url fix

bac:
Rather than just deriving the baseUrl based on whether we are standalone or jujucharms.com we allow it to be explicitly set. The use case is embedding into a containing web application where we are placed at some arbitrary route.


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/2843/
Test PASSed.

PR #853, opened 2015-09-23
Remove units

huwshimi:
Add removing of a unit from the unit details or removing multiple units from the unit list.

### DIFF
--- a/jujugui/static/gui/src/app/components/unit-details/test-unit-details.js
+++ b/jujugui/static/gui/src/app/components/unit-details/test-unit-details.js
@@ -63,7 +63,7 @@ describe('UnitDetails', function() {
     var buttons = [{
       title: 'Remove',
       action: output.props.children[1].props.buttons[0].action
-      }];
+    }];
     assert.deepEqual(output.props.children[1],
       <juju.components.ButtonRow
         buttons={buttons} />);
--- a/jujugui/static/gui/src/app/components/unit-details/test-unit-details.js
+++ b/jujugui/static/gui/src/app/components/unit-details/test-unit-details.js
@@ -59,7 +59,10 @@ describe('UnitDetails', function() {
     var output = jsTestUtils.shallowRender(
       <juju.components.UnitDetails
         unit={fakeUnit} />);
-    var buttons = [{title: 'Remove'}];
+    var buttons = [{
+      title: 'Remove',
+      action: output.props.children[1].props.buttons[0].action
+      }];
     assert.deepEqual(output.props.children[1],
       <juju.components.ButtonRow
         buttons={buttons} />);
--- a/jujugui/static/gui/src/app/components/unit-details/test-unit-details.js
+++ b/jujugui/static/gui/src/app/components/unit-details/test-unit-details.js
@@ -33,7 +33,8 @@ describe('UnitDetails', function() {
     fakeUnit = {
       private_address: '192.168.0.1',
       public_address: '93.20.93.20',
-      agent_state: 'started'
+      agent_state: 'started',
+      id: 'unit1'
     };
   });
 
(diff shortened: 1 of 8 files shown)


huwshimi:
QA:
- visit `http://0.0.0.0:6543/:flags:/react/?deploy-target=cs%3Atrusty%2Fwordpress`
- click on wordpress and then Units
- click on the unit `wordpress/0`
- click on Remove
- the deployer bar should update that a unit will be removed
- visit `http://0.0.0.0:6543/:flags:/react/?deploy-target=cs%3Atrusty%2Fwordpress`
- click on wordpress and then Units
- check the box next to `wordpress/0`
- click Remove
- the deployer bar should update that a unit will be removed

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3009/
Test PASSed.

anthonydillon:
Code looks good but in QA I had some problems:
- In the units list with 1 unit. I select the check box next to `wordpress/0` then hit remove. The change is added to the deployer bar but the unit remains in the unit list
- With multiple units they remove correctly apart from unit `/0`
- Also when committed and then removed they remain in the list until re-committed. I believe this is a UX issue.

jujugui/static/gui/src/app/components/unit-details/test-unit-details.js
```diff
@@ -59,9 +60,45 @@ describe('UnitDetails', function() {
     var output = jsTestUtils.shallowRender(
       <juju.components.UnitDetails
         unit={fakeUnit} />);
-    var buttons = [{title: 'Remove'}];
+    var buttons = [{
+      title: 'Remove',
+      action: output.props.children[1].props.buttons[0].action
+      }];
```
hatched: indent

jujugui/static/gui/src/app/components/unit-list/unit-list.js
```diff
@@ -84,6 +84,24 @@ YUI.add('unit-list', function() {
...
     /**
+      Remove the selected units.
+
+      @method _handleRemoveUnits
+    */
+    _handleRemoveUnits: function() {
+      var units = [];
+      var refs = this.refs;
+      Object.keys(refs).forEach(function (ref) {
+        var isInstance = refs[ref] instanceof juju.components.UnitListItem;
```
hatched: This check is fairly resource intensive and may cause perf issues in a large environment. Maybe instead we should set a naming pattern for refs which we could compare against... `UnitListItem-mysql/0` for example

hatched:
QA Notes:

- From the unit pane when I remove a unit it changes back to the unit list but the unit is still in the list - we probably need some type of a 'status indicator' on those units. The db change event appears to fire but the unit data we have from the service unit list doesn't contain the 'deleted' flag which is available in the `app.db.units.item(0)` object. We will need to get that data into the service unit list and then update the unit lists accordingly.
- If I select 'select all' then 'remove' it deselects all the units checkboxes but the select all box is still checked.
- Trying to remove an uncommitted unit from the unit list throws a type error because the id it's trying to remove doesn't exist (ghost id vs real id) clicking on the unit and removing appears to work.

anthonydillon:
Also I noticed that if you auto place units 1 per machine and then remove them it removes the unit but not the machine. So if you scale 20 units and then select all the remove it removed the units but there are still 20 machines. This seems odd to me.

hatched:
@anthonydillon that is the intended functionality as creating a machine is a separate action so there is no way for us to know that that machine was created specifically by 'autoplace' - also by default juju leaves the machines around so you have to take explicit steps to remove.

huwshimi:
@hatched, @anthonydillon I completely agree, we need some kind of indicator that a committed service will be removed. At the moment we have the same problem with services.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3026/
Test PASSed.

hatched:
Thanks for the update @huwshimi The previous issues have been resolved but there is now a new one.

To reproduce:
- Load up the GUI and run `app.get('charmstore').getEntity('precise/python-django', function(entity) { app.deployService(entity[0]); });` 
- Scale up the service by a couple units then commit
- Navigate to the units list and click "Select all"
- Then remove.. it thinks the unit name is 'python'

hatched:
Looks to be working great now!

:+1: QA OK!

huwshimi:
Thanks @anthonydillon and @hatched :shipit:

jujugui:

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3027/
Test PASSed.

PR #855, opened 2015-09-23
Introduce the socket_path option.

frankban:
The socket_path config option can be used to explicitly set the path to use for the WebSocket connection to the Juju API.

Also allow user and password to be set via the wsgi app ini configuration file.
Bump version up to 0.0.3.

### DIFF
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -445,9 +445,11 @@ YUI.add('juju-gui', function(Y) {
         });
         // Instantiate the environment specified in the configuration, choosing
         // between the available implementations, currently Go and Python.
+        var socketUrl = this._generateSocketUrl();
+        this.set('socket_url', socketUrl);
         var envOptions = {
           ecs: ecs,
-          socket_url: this._generateSocketUrl(),
+          socket_url: socketUrl,
           user: this.get('user'),
           password: this.get('password'),
           readOnly: this.get('readOnly'),
@@ -832,11 +834,11 @@ YUI.add('juju-gui', function(Y) {
     },
 
     /**
-      Composes the various socket paths and protocols and returns the corret
-      url that the GUI should use to communicate with the environment.
+      Composes the various socket paths and protocols and returns the correct
+      URL that the GUI should use to communicate with the environment.
 
       @method _generateSocketUrl
-      @return {String} The fully qualified socket url.
+      @return {String} The fully qualified WebSocket URL.
     */
     _generateSocketUrl: function() {
       var socketProtocol = this.get('socket_protocol');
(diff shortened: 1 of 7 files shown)


frankban:
There is no easy way to QA this for now. The socket_path feature will be QAed later as part of a separate pull request.

jujugui/views.py
```diff
@@ -69,8 +69,13 @@ def config(request):
...
     request.response.content_type = 'application/javascript'
     sandbox_enabled = settings['jujugui.sandbox']
+    # If sandbox is enabled then set the password to "admin" so that the
+    # Juju GUI will automatically log in.
+    user, password = 'user-admin', 'admin'
+    if not sandbox_enabled:
+        user, password = settings['jujugui.user'], settings['jujugui.password']
     env_uuid = request.matchdict.get('uuid', 'sandbox')
-    baseUrl = settings.get('jujugui.baseUrl')
+    baseUrl = settings['jujugui.baseUrl']
```
fabricematrat: Why not a get anymore ?

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3013/
Test PASSed.

jujugui/static/gui/src/app/app.js
```diff
@@ -847,6 +849,12 @@ YUI.add('juju-gui', function(Y) {
       if (loc.port) {
         socketUrl += ':' + loc.port;
       }
+      // If a WebSocket path is explicitly provided, it gets precedence over
+      // all the other methods to automatically calculate it.
+      var path = this.get('socket_path');
```
fabricematrat: Where is socket_path coming from ?

jujugui/views.py
frankban: Because `jujugui.baseUrl` is always included in the dict. So using get was quite confusing.

jujugui/static/gui/src/app/app.js
```diff
@@ -847,6 +849,12 @@ YUI.add('juju-gui', function(Y) {
...
+      if (path) {
```
fabricematrat: I'd use path !== null

fabricematrat:
:+1:  code :ok:  no QA

jujugui/static/gui/src/app/app.js
frankban: From the config file.

jujugui/static/gui/src/app/app.js
frankban: We want to handle the case wher path is undefined or empty string as well.

fabricematrat: it can't be undefined in reality from the call above but it's more for cosmetic.

jaycee:
:+1: thanks @frankban.

frankban:
Thanks for the reviews!
:shipit:

jujugui:

PR #860, opened 2015-09-28
Add pending and error states to service blocks

anthonydillon:


anthonydillon:
## QA
- Deploy a few units of a service
- Simulate running with `CTRL + S`
- Watch for an pending unit which changes to border colour and width of the service block
- Check that when there is an error it trumps the pending and makes the border red

Design: 
service-blocks_states01 1

anthonydillon:
This branch only covers the Normal/Normal, Normal/Uncommitted, Normal/Pending and Normal/Error.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3055/
Test PASSed.

jujugui/static/gui/src/app/views/topology/service.js
```diff
@@ -151,10 +155,32 @@ YUI.add('juju-topology-service', function(Y) {
       } else if (d.highlighted) {
         rerenderRelations = true;
       } else {
-        curr_node.attr({
-          'stroke': '#888888',
-          'stroke-width': 1
-        });
+        var units = d.units._items;
+        for (var i = 0; i < units.length; i++) {
```
hatched: This could be a pretty big performance bottleneck in large environments. We should have a flag on the units modellist which indicates when a unit's agent state goes into error, pending, etc so that the multiple places around the app which require this information can use it. 

Could you take a look at this, and if it looks like too much work for this branch create a follow-up card. thx!

hatched:
:+1: QA OK

I'm ok with pushing that performance improvement into another branch/card if necessary.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3069/
Test PASSed.

jujugui/static/gui/src/app/views/topology/service.js
```diff
@@ -151,10 +152,32 @@ YUI.add('juju-topology-service', function(Y) {
       } else if (d.highlighted) {
         rerenderRelations = true;
       } else {
-        curr_node.attr({
-          'stroke': '#888888',
-          'stroke-width': 1
+        var units = d.units._items;
+        units.forEach(function(unit) {
```
anthonydillon: @hatched noted this would result in a performance hit in large bundles. @makyo do you think it would be possible to use the unitServiceList to determine state?

jujugui/static/gui/src/app/views/topology/service.js
```diff
@@ -151,10 +152,32 @@ YUI.add('juju-topology-service', function(Y) {
...
+          if (unit.agent_state === 'installing'
+              || unit.agent_state === 'pending') {
+            is_pending = true;
+          } else if (unit.agent_state === 'error') {
+            is_erroring = true;
+          }
         });
+
+        if (is_erroring) {
```
makyo: I think the logic is pretty good for this.  I wonder if it would be easier to add a class to the service blocks rather than setting the stroke color/width as attributes?  That would make it a little easier to test, because then you can just test for the presence of the classes (normal + status-normal, etc).

makyo:
:+1: with a suggestion to help testing.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/3090/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3091/
Test PASSed.

anthonydillon:
@hatched @makyo This look ok now?

anthonydillon:
:shipit:

anthonydillon:
:shipit:

jujugui:

PR #890, opened 2015-10-08
Implement charm details header as a React component.

kadams54:


kadams54:
## QA

Click on the Wordpress charm in the midpoint. The entity header will load; verify it looks similar to what's at https://jujucharms.com/wordpress/trusty/3. Note that the copy-n-paste command and the "Add to demo" button are not yet present; right now those are planned as separate subcomponents that will be added in subsequent PRs.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/3198/
Test FAILed.

jujugui/static/gui/src/app/components/charmbrowser/test-charmbrowser.js
```diff
@@ -43,14 +43,12 @@ describe('Charmbrowser', function() {
         }
       }};
     var changeState = sinon.stub();
-    var addService = sinon.stub();
```
kadams54: Cleanup so we're only stubbing out what actually gets passed to/checked in the child component.

jujugui/static/gui/src/app/components/charmbrowser/test-charmbrowser.js
```diff
@@ -69,7 +67,6 @@ describe('Charmbrowser', function() {
           activeComponent: 'mid-point'
         }
       }};
-    var clearState = sinon.stub();
```
kadams54: More cleanup.

jujugui/static/gui/src/app/components/charmbrowser/test-charmbrowser.js
```diff
@@ -119,17 +116,17 @@ describe('Charmbrowser', function() {
           id: id
         }
       }};
-    var clearState = sinon.stub();
-    var addService = sinon.stub();
```
kadams54: Last bit of test cleanup.

jujugui/static/gui/src/app/models/entity-extension.js
```diff
@@ -18,19 +18,20 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 'use strict';
 
-YUI.add('search-result-extension', function(Y) {
+YUI.add('entity-extension', function(Y) {
```
kadams54: Realized that this extension is more generically re-usable beyond just search results. We need pretty much the same POJO display object in the details page. Renamed it to `entity-extension` and clarified that it holds functionality (not just producing the display object) common to both charms and bundles.

jujugui/static/gui/src/app/components/entity-details/entity-details.js
```diff
@@ -18,14 +18,111 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+            <header className="twelve-col">
+              <div className="inner-wrapper">
+                <div className="eight-col no-margin-bottom">
+                  <img src={entity.iconPath} alt="" width="96" id="the-icon"/>
+                  <div className="header__details">
+                    <h1 className="header__title" itemprop="name">{entity.displayName}</h1>
+                    <p className="header__by">By <a href={'https://launchpad.net/~' + entity.owner} target="_blank">{entity.owner}</a></p>
+                    <ul className="bullets inline">
+                      <li className="revisions-item">
+                        <a href="#revisions" className="revisions-link">{entity.revisions.length} {utils.pluralize('revision', entity.revisions.length)}</a>
```
kadams54: Note that I broke from the pattern of passing functions in for `pluralize`; this instance seems like a case where requiring in the functionality makes more sense than injecting it.

jujugui/static/gui/src/app/components/entity-details/entity-details.js
```diff
@@ -18,14 +18,111 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 'use strict';
 
-YUI.add('entity-details', function() {
+YUI.add('entity-details', function(Y) {
+
+  var utils = Y.namespace('juju.views.utils');
```
kadams54: Comment below explaining why utils is being brought in.

jujugui/static/gui/src/app/components/search-results/search-results.js
```diff
@@ -103,7 +103,7 @@ YUI.add('search-results', function(Y) {
     searchSuccess: function(rawResults) {
       // Parse the raw results.
       var results = rawResults.map(function(model) {
-        return model.toSearchResult();
+        return model.toEntity();
```
kadams54: Comment below on this change.

jujugui/static/gui/src/app/assets/css/_mixins.scss
```diff
@@ -203,3 +203,18 @@
     overflow-x: hidden;
     overflow-y: auto;
 }
+%shadow-header {
+    overflow: hidden;
+    padding-top: 0;
+    background: #fff;
+    color: $cool-grey;
+
+    header {
```
hatched: switching to the new BEM workflows we don't target tags anymore because they influence too broadly.

jujugui/static/gui/src/app/components/entity-details/_entity-details.scss
```diff
@@ -0,0 +1,613 @@
...
+            }
+        }
+
+        .featured {
+            float: right;
+            font-size: 0.875em;
+            color: #BABABA;
+        }
+
+        #the-icon {
```
hatched: id?

jujugui/static/gui/src/app/components/entity-details/_entity-details.scss
```diff
@@ -0,0 +1,613 @@
...
+        @media only screen and (min-width : 769px) {
+            border: 1px solid $light-mid-grey;
+        }
+    }
+
+    .bullet {
+        color: $warm-grey;
+    }
+
+    main {
```
hatched: global tag styling

jujugui/static/gui/src/app/views/utils.js
```diff
@@ -2010,29 +2010,30 @@ YUI.add('juju-view-utils', function(Y) {
...
-          if (typeof(plural_word) === 'string') {
-            return plural_word;
-          } else {
-            return word + 's';
-          }
-        } else {
-          return word;
-        }
-      });
+  utils.pluralize = function(word, object, plural_word, options) {
```
hatched: Lets create a component utils file void of all YUIness, wrapped in a YUI.add() for convoy and put this function in there. I couldn't find anywhere that actually used this helper but if required you can then leave the hbs helper code in here and we can delete it when we upgrade whatever requires it.

jujugui/static/gui/src/app/components/entity-details/entity-details.js
hatched: No YUI.

I commented below in the utils file about creating a component utils file. Because it would be a utils file it would probably be nice to pass it to every top level component.

jujugui/static/gui/src/app/components/entity-details/entity-details.js
```diff
@@ -18,14 +18,111 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
...
 
   juju.components.EntityDetails = React.createClass({
+    fetchSuccess: function(models) {
```
hatched: all of these methods are missing docblocks

jujugui/static/gui/src/app/components/entity-details/entity-details.js
```diff
@@ -18,14 +18,111 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+      if (!entity) {
+        return (
+          <div className="spinner-loader">Loading...</div>
+        );
+      }
+      if (entity.tags) {
+        tags = entity.tags.map(function(tag) {
+          return(
+            <li className="tag-list--item">
+              <a target="_blank" href={'/q?tags=' + tag}>{tag}</a>
```
hatched: This will need to be changed to modify state `onClick` instead of changing the url.

jujugui/static/gui/src/app/components/entity-details/entity-details.js
hatched: This will also need to update state when `onClick` so right now we could just leave off the hash support.

jujugui/static/gui/src/app/components/entity-details/test-entity-details.js
```diff
@@ -34,7 +34,7 @@ describe('EntityDetails', function() {
   it('can be rendered', function() {
```
hatched: This suite is missing the tests for the multiple entity details rendering outputs.

jujugui/static/gui/src/app/models/entity-extension.js
```diff
@@ -18,19 +18,20 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-   * Class extension that allows models to be displayed as search results.
+   * Class extension containing functionality common to both charms and
+   * bundles.
    *
    * @namespace juju.models
-   * @class SearchResult
+   * @class Entity
    */
-  function SearchResult() {}
+  function Entity() {}
```
hatched: I wonder if this shouldn't be EntityExtension?

jujugui/static/gui/src/app/components/charmbrowser/test-charmbrowser.js
```diff
@@ -43,14 +43,12 @@ describe('Charmbrowser', function() {
...
       }};
     var changeState = sinon.stub();
-    var addService = sinon.stub();
     var charmstoreSearch = sinon.stub();
     var output = jsTestUtils.shallowRender(
       <juju.components.Charmbrowser
         appState={appState}
         changeState={changeState}
-        charmstoreSearch={charmstoreSearch}
-        addService={addService} />);
```
huwshimi: You're correct in that `addService` is not currently being used, but when you go to make the 'Add to canvas' button work you'll want to use that method. Up to you when you want to add that back in :).

kadams54: It's more that `addService` is not being used in this particular test. We really only need to mockup (and eventually assert on) the services that are being used by that particular test. For example, the test to make sure the entity details are loaded properly doesn't need to mockup charmstoreSearch, because search isn't used. It just ends up being an undefined prop, which is OK because it never gets referenced in the entity details branch of code.

That said, we also need to talk about `addService`. I took a look at the underlying code in utils and it's ugly. The full function call is `utils.addService(app, charmId)`, so any component calling it would need app passed in. I don't think we want to pass the entire application down through each component. We need an lightweight (or at leas lighter) implementation of `addService`.

jujugui/static/gui/src/app/components/entity-details/entity-details.js
kadams54: I'll add them, but docblocks are not at all consistent in the React code right now. They aren't in `search-results.js`, which is on me, but it was never noted in any of the reviews on that code. They're in `charmbrowser.js`, but only for the functions that aren't lifecycle-related (which seems very reasonable to me).

We should settle on some guidelines for moving foward and we may want to think about a different format post-YUI, e.g., JSDoc.

jujugui/static/gui/src/app/components/charmbrowser/test-charmbrowser.js
huwshimi: Well, remember that you can just bind the app at the top level when you're passing in the `addService` method, so no need to pass app around (that's the whole point). We're not using `addService` anywhere at the moment so feel free to modify it as you see fit.

huwshimi: Having had another look at `addService` you could update it to pass in the `getEntity` and `deployService` methods if that was easier.

kadams54: I'm not sure what the impact would be, but based on my other refactor PR (to remove the charmstore dependency from `getIconPath`) I'm guessing changing `addService` would not be trivial. We already have the `getEntity` method in the `EntityDetails` component; I'd rather just pass in `deployService` and bypass `addService` altogether.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/3207/
Test FAILed.

jujugui/static/gui/src/app/components/charmbrowser/test-charmbrowser.js
huwshimi: I added `addService` last week, you're fine to modify it as you see fit, if you don't end up using it, please remove it.

jujugui/static/gui/src/app/views/utils.js
kadams54: I may or may not need more functions from the existing `utils`, so I'm leaving this in here for the moment and will revisit once the entire entity page is implemented.

jujugui/static/gui/src/app/components/entity-details/test-entity-details.js
kadams54: All logic paths are now covered by tests.

jujugui/static/gui/src/app/components/entity-details/entity-details.js
kadams54: Utils is now passed into Charmbrowser, and it passes pluralize into `EntityDetails`. That may change as `EntityDetails` grows to render the rest of the details page, but it's a start.

jujugui/static/gui/src/app/components/charmbrowser/test-charmbrowser.js
kadams54: Will do!

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/3208/
Test FAILed.

jujugui/static/gui/src/app/components/entity-details/entity-details.js
hatched: Why is this still left here if it's now being passed in?

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3209/
Test PASSed.

jujugui/static/gui/src/app/components/entity-details/entity-details.js
```diff
@@ -18,11 +18,162 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 'use strict';
 
-YUI.add('entity-details', function() {
+YUI.add('entity-details', function(Y) {
```
hatched: unused Y, odd that linter didn't pick this up.

jujugui/static/gui/src/app/components/entity-details/entity-details.js
```diff
@@ -18,11 +18,162 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+              {tag}
+            </a>
+          </li>
+        );
+      }, this);
+      var ownerUrl = 'https://launchpad.net/~' + entity.owner;
+      return (
+        <div className={'entity-details ' + entity.type}>
+          <div className="row-hero">
+            <header className="twelve-col header">
```
hatched: I feel like this header should be broken out into a separate component because otherwise this file is going to get massive and difficult to test.

jujugui/static/gui/src/app/components/entity-details/test-entity-details.js
hatched: I don't think so, there are three possible outputs from the render function and only a single test below. Take a look at how Huw did the tests in his branch here https://github.com/juju/juju-gui/pull/891/files he tests the full output splitting up the components into many smaller components to make it easier to test.

jujugui/static/gui/src/app/views/utils.js
hatched: Alright, can you create a card so it doesn't get lost?

jujugui/static/gui/src/app/components/entity-details/entity-details.js
kadams54: I'm planning on breaking the header out in my next card, which deals with pulling the main content in.

jujugui/static/gui/src/app/views/utils.js
kadams54: Card created.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3214/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3215/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3216/
Test PASSed.

hatched:
:+1: Thanks for sticking through this one :)

QA :ok:

QA Notes:
- It would be nice if the loading indicator was inside a whitebox as well, I see there is a card to normalize the loading indicator though.
- I got the following warning when playing around with the interactions but I wasn't able to reliably
reproduce it so we'll have to keep an eye on it. `Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op.`

kadams54:
Excellent! :shipit:

kadams54:
Excellent! :shipit:

jujugui:

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/3217/
Test FAILed.

PR #906, opened 2015-10-15
Fix id mismatch for units.

huwshimi:
Fix lookup of units if they were left uncommitted when a service was committed.

### DIFF
--- a/jujugui/static/gui/src/app/models/models.js
+++ b/jujugui/static/gui/src/app/models/models.js
@@ -1012,6 +1012,25 @@ YUI.add('juju-models', function(Y) {
     },
 
     /**
+      Get a unit by displayName.
+
+      @method getByDisplayName
+      @param {String} displayName The display name for a unit.
+      @return {Object} The matching unit with the matching display name.
+    */
+    getByDisplayName: function(displayName) {
+      var units = this.filter(function(unit) {
+        if (unit.displayName === displayName) {
+          return unit;
+        }
+      });
+      if (units.length > 0) {
+        return units[0];
+      }
+      return;
+    },
+
+    /**
       Returns information about the state of the set of units for a given
       service in the form of a map of agent states.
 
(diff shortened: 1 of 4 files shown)


huwshimi:
QA:
- add a service
- navigate to the uncommitted unit details
- deploy the service (leaving the unit uncommitted)
- navigate to the uncommitted unit details
- check that there are no errors in your console

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3257/
Test PASSed.

jujugui/static/gui/src/app/models/models.js
```diff
@@ -1012,6 +1012,25 @@ YUI.add('juju-models', function(Y) {
...
+      @param {String} displayName The display name for a unit.
+      @return {Object} The matching unit with the matching display name.
+    */
+    getByDisplayName: function(displayName) {
+      var units = this.filter(function(unit) {
+        if (unit.displayName === displayName) {
+          return unit;
+        }
+      });
+      if (units.length > 0) {
```
hatched: curious why there would be more than one unit which matches the display name? Could you add a comment to this effect.

jujugui/static/gui/src/test/test_model.js
```diff
@@ -1084,6 +1084,14 @@ describe('test_model.js', function() {
       });
     });
 
+    it('can get a unit but displayName', function() {
```
hatched: by

hatched:
I didn't know this was actually a problem, I had to go and test it without the patch first :) Good find!

I'm wondering if this is actually just covering up the problem though. I think that the real fix is to update these unit id's on deploy so that this problem doesn't exist at all.

jujugui/static/gui/src/app/models/models.js
huwshimi: I should probably just update that to be `=== 0`

huwshimi:
A better fix for this issue is over at: https://github.com/juju/juju-gui/pull/933

PR #910, opened 2015-10-16
Bundle add to canvas

huwshimi:
Added functionality to add a bundle to the canvas.

### DIFF
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -887,11 +887,14 @@ YUI.add('juju-gui', function(Y) {
     _renderCharmbrowser: function(metadata) {
       var state = this.state;
       var utils = views.utils;
-      var cs = this.get('charmstore');
+      var charmstore = this.get('charmstore');
       React.render(
         <components.Charmbrowser
-          charmstoreSearch={cs.search.bind(cs)}
-          getEntity={cs.getEntity.bind(cs)}
+          charmstoreSearch={charmstore.search.bind(charmstore)}
+          importBundleYAML={this.bundleImporter.importBundleYAML.bind(
+              this.bundleImporter)}
+          getBundleYAML={charmstore.getBundleYAML.bind(charmstore)}
+          getEntity={charmstore.getEntity.bind(charmstore)}
           deployService={this.deployService.bind(this)}
           appState={state.get('current')}
           changeState={this.changeState.bind(this)}
--- a/jujugui/static/gui/src/app/components/entity-header/entity-header.js
+++ b/jujugui/static/gui/src/app/components/entity-header/entity-header.js
@@ -55,7 +55,24 @@ YUI.add('entity-header', function() {
       @param {Object} e The click event
     */
     _handleDeployClick: function(e) {
-      this.props.deployService(this.props.entityModel);
+      var entityModel = this.props.entityModel;
+      var entity = entityModel.toEntity();
+      if (entity.type === 'charm') {
+        this.props.deployService(entityModel);
+        this._closeEntityDetails();
+      } else {
+        var id = entity.id.replace('cs:', '');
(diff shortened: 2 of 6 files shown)


huwshimi:
QA:
- search for a bundle
- view the details
- click "Add to canvas"
- the bundle should be added

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3272/
Test PASSed.

jujugui/static/gui/src/app/app.js
```diff
@@ -887,11 +887,14 @@ YUI.add('juju-gui', function(Y) {
...
       var state = this.state;
       var utils = views.utils;
-      var cs = this.get('charmstore');
+      var charmstore = this.get('charmstore');
       React.render(
         <components.Charmbrowser
-          charmstoreSearch={cs.search.bind(cs)}
-          getEntity={cs.getEntity.bind(cs)}
+          charmstoreSearch={charmstore.search.bind(charmstore)}
+          importBundleYAML={this.bundleImporter.importBundleYAML.bind(
```
urosj: how are errors handled here?

jujugui/static/gui/src/app/components/entity-header/entity-header.js
```diff
@@ -55,7 +55,24 @@ YUI.add('entity-header', function() {
...
     */
     _handleDeployClick: function(e) {
-      this.props.deployService(this.props.entityModel);
+      var entityModel = this.props.entityModel;
+      var entity = entityModel.toEntity();
+      if (entity.type === 'charm') {
+        this.props.deployService(entityModel);
+        this._closeEntityDetails();
+      } else {
+        var id = entity.id.replace('cs:', '');
```
urosj: why do we need to drop cs: with bundles?

jujugui/static/gui/src/app/components/entity-header/test-entity-header.js
```diff
@@ -120,11 +139,15 @@ describe('EntityHeader', function() {
         title="Add to canvas" />);
   });
 
-  it('adds the service when the add button is clicked', function() {
+  it('adds a charm when the add button is clicked', function() {
```
urosj: a charm or a bundle here?

urosj:
Not seen a nicer animation of a deploy of a bundle, yet ... sweeeet!

jujugui/static/gui/src/app/app.js
huwshimi: Here we're just passing the reference to the `importBundleYAML` method. If need be we can do some error handling where we call the method here: https://github.com/juju/juju-gui/pull/910/files#diff-7cf7695007c40f36e1ca92951612a296R91

jujugui/static/gui/src/app/components/entity-header/entity-header.js
huwshimi: The entity id is something like 'cs:~charmers/django' but when we use that to do a lookup in the api we insert it into the url like this `https://api.jujucharms.com/charmstore/v4/~charmers/django...` hence needing to drop the `cs:` before generating the URL.

jujugui/static/gui/src/app/components/entity-header/test-entity-header.js
huwshimi: The is the deploy charm test. The bundle test was added below.

urosj:
LGTM and +1 on QA

jujugui/static/gui/src/app/components/entity-header/entity-header.js
urosj: I'd suggest to add some url resolving function in one place and use that in the code. If URLs change, it'll be hard to find all such examples. But not for this PR.

hatched:
:+1: QA :ok:

:shipit:

jujugui:

PR #926, opened 2015-10-21
Create a standalone Spinner component.

kadams54:


kadams54:
## QA

Load up entity details for some entity and make sure everything looks right in the brief moments the spinner is displayed. If you want, modify `entity-details.js` around line 90 so that the spinner is never replaced by content.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3309/
Test PASSed.

huwshimi:
Not sure what's up but (even after a clean) this branch can't load the component for the entity details. `'spinner'` will need to be added to the `requires` array, but that doesn't fix it.

kadams54:
Added the dependency in. I'm not sure why it worked locally for me without it, so I did a `make clean-all gui` before running it again. The spinner still loads: Spinner working

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3310/
Test PASSed.

jujugui/static/gui/src/app/components/entity-details/test-entity-details.js
```diff
@@ -85,7 +85,7 @@ describe('EntityDetails', function() {
         changeState={sinon.spy()}
         getEntity={sinon.spy()}
         pluralize={sinon.spy()} />);
-    assert.equal(output.props.children, 'Loading...');
+    assert.equal(output.props.className, 'entity-details');
```
hatched: You can now do a deep equals on this

```JavaScript
assert.deepEqual(output, 
           <div className="entity-details">
            <juju.components.Spinner/>
          </div>);
```

hatched:
It would be nice if you updated the search results to use this one as well.

jujugui/static/gui/src/app/components/spinner/spinner.js
```diff
@@ -0,0 +1,33 @@
...
+*/
+
+'use strict';
+
+YUI.add('spinner', function() {
+
+  juju.components.Spinner = React.createClass({
+    render: function() {
+      return (
+        <div className="spinner_loading">Loading...</div>
```
hatched: Why does it still have 'Loading' ? Is this a fallback?

kadams54:
It would indeed be nice… if search results wasn't rendering the spinner in the handlebars template ;-)

jujugui/static/gui/src/app/components/spinner/spinner.js
kadams54: Correct, in case the CSS spinner doesn't show.

jujugui/static/gui/src/app/components/entity-details/test-entity-details.js
kadams54: `deepEqual` was giving me all sorts of fits in `EntityDetails` - when I visually inspected/drilled down into both objects they looked equal in every regard. The `Spinner` itself had its own tests, so I didn't need to cover that. All that the test in `EntityDetails` really needed to verify was that the `Spinner` was wrapped in something with the `entity-details` class. So I streamlined the test to just check that.

hatched:
:+1: QA :ok:

kadams54:
Thanks all! :shipit:

jujugui:

PR #988, opened 2015-11-16
Update node and dependencies

hatched:
After a number of previous branches fixed larger issues with the node upgrade we can finally upgrade to the latest node and the latest dependencies. This branch switches our reliance on the nodejs installation in the yellow ppa and instead pulls node directly from the nodesource deb.


Makefile
```diff
@@ -213,7 +213,7 @@ $(BUILT_D3):
 	  node_modules/d3/src/svg/line.js \
 	  node_modules/d3/src/svg/arc.js \
 	  node_modules/d3/src/event/drag.js \
-	  $(GUISRC)/app/assets/javascripts/unscaled-pack.js \
+		node_modules/d3/src/layout/pack.js \
```
hatched: We will need to qa this once we have a charm installable GUI again to make sure that this functionality is either a) good enough or b) we have to write a new unscaled-pack. The methods our unscaled-pack relies on are no longer available in the latest version of d3

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/3593/
Test FAILed.

jujugui/static/gui/src/test/test_app.js
```diff
@@ -82,13 +82,12 @@ function injectData(app, data) {
 
     });
 
-    afterEach(function(done) {
```
hatched: The updated mocha doesn't like it when done() is called multiple times per test

package.json
```diff
@@ -10,34 +10,33 @@
   },
   "dependencies": {
     "babel": "^5.8.29",
-    "chai": "^3.2.0",
```
hatched: Updates almost all of the dependencies. babel still needs to be updated to the 6.x line but the changes are quite big so I will do that on another friday

bac:
Please make sure the makefile help includes all the targets. I noticed fast-babel is missing.

bac:
So QA was not a clear win. @hatched please advise about the resource issues.

PR #997, opened 2015-11-19
Add service visibility toggles to the services list

hatched:
Clicking the two new buttons in the added services list allows the user to fade and highlight specific services.

### DIFF
--- a/jujugui/static/gui/src/app/components/added-services-list-item/added-services-list-item.js
+++ b/jujugui/static/gui/src/app/components/added-services-list-item/added-services-list-item.js
@@ -139,7 +139,7 @@ YUI.add('added-services-list-item', function() {
       var statusData = this._getPriorityUnits(service.units.toArray());
       var statusIndicator = this._renderStatusIndicator(statusData);
       var focusIcon = state.focus ? 'focused_16' : 'unfocused_16';
-      var highlightIcon = state.highlight ? 'highlight_16' : 'unhighlight_16';
+      var highlightIcon = state.highlight ? 'show_16' : 'hide_16';
       return (
         <li className="inspector-view__list-item"
             data-serviceid={service.id}
--- a/jujugui/static/gui/src/app/components/added-services-list-item/added-services-list-item.js
+++ b/jujugui/static/gui/src/app/components/added-services-list-item/added-services-list-item.js
@@ -22,6 +22,13 @@ YUI.add('added-services-list-item', function() {
 
   juju.components.AddedServicesListItem = React.createClass({
 
+    propTypes: {
+      focusService: React.PropTypes.func.isRequired,
+      getUnitStatusCounts: React.PropTypes.func.isRequired,
+      changeState: React.PropTypes.func.isRequired,
+      service: React.PropTypes.object.isRequired
+    },
+
     getInitialState: function() {
       return {
         focus: false,
@@ -100,7 +107,12 @@ YUI.add('added-services-list-item', function() {
       // We need to stop the propagation so that the click event doesn't
       // bubble up to the list item and navigate away.
       e.stopPropagation();
-      this.setState({focus: !this.state.focus});
+      var focus = !this.state.focus;
+      this.setState({focus: focus});
+      if (focus) {
+        this.setState({highlight: false});
+      }
+      this.props.focusService(this.props.service.get('id'));
     },
 
     /**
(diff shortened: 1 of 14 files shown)


hatched:
#### To QA

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/3628/
Test FAILed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/3632/
Test FAILed.

huwshimi:
QA looks good, a couple of small things:
- the alignment of the icons is a bit off when there are status badges: 
alignment
- the new icons mean we'll need to change the cut-off point for long charm names (maybe about 110px): https://github.com/juju/juju-gui/blob/develop/jujugui/static/gui/src/app/components/added-services-list/_added-services-list.scss#L41

hatched:
screenshot from 2015-11-19 22 15 10

hatched:
@huwshimi The mockups actually have the visibility toggles hidden except on hover so once I added that it looked a lot better.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/3639/
Test FAILed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/3640/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3641/
Test PASSed.

jujugui/static/gui/src/app/components/added-services-list-item/added-services-list-item.js
```diff
@@ -22,6 +22,30 @@ YUI.add('added-services-list-item', function() {
...
+      var service = this.props.service;
+      return {
+        focus: service.get('highlight') || false,
+        fade: service.get('fade') || false
+      };
+    },
+
+    componentWillReceiveProps: function(nextProps) {
+      var service = this.props.service;
+      this.setState({focus: service.get('highlight')});
```
huwshimi: Is there a reason these don't need `|| false` like above?

huwshimi:
I like the hover stuff, much better. one comment, if you've hidden/focussed services it can be hard to know which one it was and you end up hovering over all the services looking for the ones that have been hidden. Maybe if you've hidden/focussed as service the icon should be visible while it is active?

QA OK :+1:

jujugui/static/gui/src/app/components/added-services-list-item/test-added-services-list-item.js
```diff
@@ -248,4 +387,185 @@ describe('AddedServicesListItem', function() {
...
+        service={service} />, true);
+
+    var instance = renderer.getMountedInstance();
+    assert.equal(instance.state.focus, true);
+    assert.equal(instance.state.fade, true);
+
+    // This is ugly but we have to check that the proper name prop was passed
+    // to the SvgIcon component.
+    var output = renderer.getRenderOutput();
+    assert.deepEqual(
```
kadams54: Would teaspoon (or having some sort of JSX query utility) help out in situations where we want to shallow render but do deep introspection?

jujugui/static/gui/src/app/components/entity-content/test-entity-content.js
```diff
@@ -163,7 +163,6 @@ describe('EntityContent', function() {
         {undefined}
       </div>
     );
-    debugger;
```
kadams54: Whoops, sorry! Thanks for cleaning up after me…

kadams54:
Code LGTM, :+1:. QA is in progress.

jujugui/static/gui/src/app/components/added-services-list-item/test-added-services-list-item.js
hatched: Yeah, this test in particular was quite a struggle because I have to use shallowRender so that the sub components don't render but I also wanted to make sure that the proper props are passed to the sub component so the only way I could figure that out is by doing that deep object check

kadams54:
Something still seems off when there's a status indicator in the row with a long service name:

Screenshot

Also: when some rows have service indicators and others don't, it causes the icons to "jump" around - as click targets they're not in a consistent location:

Screencast

We either need to pad on the right side when there is no status indicator or move the status indicator to the far left (before the service name).

kadams54:
One final bit: if we leave the status indicator on the right side, the spacing between it and the icons should be consistent. Right now the space between the icons is wider than the space between the icon and status indicator. I think it would also help if the status indicator was the same size as the icons; that is, slightly smaller than what it currently is.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3658/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3677/
Test PASSed.

hatched:
Thanks for the reviews and qa's! :shipit:

jujugui:

kadams54:
My official QA is :ok: :-)

PR #998, opened 2015-11-19
Resize on subordinate status change

makyo:
This appropriately deals with all cases of a service moving between subordinate and not subordinate status except one (moving from subordinate back to non-subordinate service, there is a rendering issue with the add-relation button showing up off to the side, which appears in chrome).  As this situation has not cropped up (would be impossible in core\*), I'm leaving it until we find a way to fix that rendering

\* Why we're receiving subordinate information separate from environment status is the root of the problem, but not addressed in this card.  This card only touches on the display of services.

### DIFF
--- a/jujugui/static/gui/src/app/views/topology/service.js
+++ b/jujugui/static/gui/src/app/views/topology/service.js
@@ -146,7 +146,7 @@ YUI.add('juju-topology-service', function(Y) {
       });
     node.select('.service-icon')
       .attr('transform', function(d) {
-        return d.subordinate ? 'translate(17, 17)' : 'translate(47, 47)'
+        return d.subordinate ? 'translate(17, 17)' : 'translate(47, 47)';
       });
     node.select('.relation-button')
       .attr('transform', function(d) {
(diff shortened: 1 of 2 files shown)


makyo:
# QA

* If you're Uros, deploy a subordinate service (or a bundle containing one) and ensure that it shows as subordinate
* All: deploy a non subordinate service (displays as normal), open the console and enter `app.db.services.item(0).set('subordinate', true)` (service shows as a subordinate)

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/3631/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3638/
Test PASSed.

hatched:
QA Issue:
 I know the case we're trying to fix here is the case where the service goes from full to subordinate. But when I do that, then try and run it in reverse it puts the relation link in the incorrect spot.

hatched:
:+1: QA :ok: Thanks for stickin with this and getting it solved!

huwshimi:
Thanks @makyo QA OK :+1:!

makyo:
:shipit:

jujugui:

PR #1007, opened 2015-11-23
Add a files widget to the entity details.

kadams54:


kadams54:
## QA

1. Pull up the entity details for a charm or bundle.
2. Verify that the files are listed to the right of the README.
3. Check that they match the mock here: https://cloud.githubusercontent.com/assets/5873394/8907463/5fc27a72-3476-11e5-9c42-aaff407985b0.png

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/3659/
Test FAILed.

jujugui/static/gui/src/app/components/entity-content/test-entity-content.js
```diff
@@ -156,26 +162,29 @@ describe('EntityContent', function() {
                 getFile={getFile} />
             </div>
             <div className="four-col">
-              <p>{' '}</p>
+              <juju.components.EntityFiles
+                entityModel={mockEntity}
+                pluralize={pluralize} />
```
hatched: I think this will have to be `pluralize={instance.pluralize}`

jujugui/static/gui/src/app/components/entity-content/test-entity-content.js
```diff
@@ -189,7 +198,9 @@ describe('EntityContent', function() {
...
```
hatched: see above about instance if you get failures here.

jujugui/static/gui/src/app/components/entity-files/_entity-files.scss
```diff
@@ -0,0 +1,15 @@
+.entity-files {
+    ul {
```
hatched: no entity level styles please. target a class instead.

jujugui/static/gui/src/app/components/entity-files/entity-files.js
```diff
@@ -0,0 +1,96 @@
...
+    propTypes: {
+      entityModel: React.PropTypes.object.isRequired,
+      pluralize: React.PropTypes.func.isRequired
+    },
+
+    render: function() {
+      var entityModel = this.props.entityModel;
+      var files = entityModel.get('files');
+      var archiveUrl = 'https://api.jujucharms.com/charmstore/v4/' +
+                       `${entityModel.get('full_name')}/archive`;
```
hatched: Interesting use of template strings...I'm not sure how I feel about combining the two...what do you think of
```JavaScript
var baseUrl = 'https://api.jujucharms.com/charmstore/v4';
var archiveUrl = `${baseUrl}/${entityModel.get('full_name')}/archive`;
```

jujugui/static/gui/src/app/components/entity-files/entity-files.js
```diff
@@ -0,0 +1,96 @@
...
+        );
+        */
+        return (
+          <li key={file} className="section__list-item">
+            {file}
+          </li>
+        );
+      });
+      return (
+        <div className="entity-files section" id="files">
```
hatched: use BEM style classes throughout plz

jujugui/static/gui/src/app/components/entity-files/entity-files.js
```diff
@@ -0,0 +1,96 @@
...
+      entityModel: React.PropTypes.object.isRequired,
+      pluralize: React.PropTypes.func.isRequired
+    },
+
+    render: function() {
+      var entityModel = this.props.entityModel;
+      var files = entityModel.get('files');
+      var archiveUrl = 'https://api.jujucharms.com/charmstore/v4/' +
+                       `${entityModel.get('full_name')}/archive`;
+      debugger;
```
hatched: debugger statement

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/3660/
Test FAILed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/3661/
Test FAILed.

jujugui/static/gui/src/app/components/entity-files/entity-files.js
kadams54: There are two things going on here:

1. We need a generic class that will be on all of the blocks on the right side, i.e., "section". These styles are in `_entity-content.scss`.
2. We need a specific class for the file widgets block, i.e., "entity-files". These styles are in `_entity-files.scss`.

I follow BEM for elements under these; the generic ones, like the header `.section__title`, are prefixed with "section". The specific ones are prefixed with "entity-files".

jujugui/static/gui/src/app/components/entity-files/_entity-files.scss
```diff
@@ -0,0 +1,15 @@
...
+        margin: 0;
+    }
+    li {
```
huwshimi: Spaces between classes.

jujugui/static/gui/src/app/components/entity-files/_entity-files.scss
kadams54: I'm not sure I buy that 100%. We have two lists, `entity-files__links` and `entity-files__files`. I could setup a class (separate from their BEM designation) that is shared by both, but that's duplicative of the HTML element they already share… that is, you'd end up with `<ul className="list">`. The benefit (more resilient to markup changes) doesn't seem strong enough to warrant the duplication. Furthermore, the styles specified in these blocks are very specific to the HTML element here - that is, they're overriding default styles. If we changed markup, we likely wouldn't need these styles any more.

kadams54: All of which is to say: I'm all for BEM. I just want to be pragmatic about it rather than dogmatic.

jujugui/static/gui/src/app/components/entity-content/test-entity-content.js
kadams54: I thought the `instance` trick was only necessary for callbacks like event handlers, because they get bound. Utility methods aren't bound so no need to use that approach.

jujugui/static/gui/src/app/components/entity-files/entity-files.js
kadams54: That's fine - will do.

jujugui/static/gui/src/app/components/entity-files/entity-files.js
```diff
@@ -0,0 +1,95 @@
...
+      var archiveUrl = 'https://api.jujucharms.com/charmstore/v4/' +
+                       `${entityModel.get('full_name')}/archive`;
+      var codeUrl = entityModel.get('code_source').location;
+      var codeLink;
+      if (codeUrl) {
+        codeUrl = codeUrl.replace('lp:', 'https://code.launchpad.net/');
+        codeLink = (
+          <li>
+            <a target="_blank"
+              className="section__actions--launchpad"
```
huwshimi: Where are we using these classes? I couldn't see them in the css...

jujugui/static/gui/src/app/components/entity-files/entity-files.js
```diff
@@ -0,0 +1,95 @@
...
+            </a>
+          </li>
+        );
+      } else {
+        codeLink = '';
+      }
+      var fileItems = files.map(function(file) {
+        /*
+        XXX Verify whether or not these should be links. Mock seems to
+        show them as unlinked.
```
huwshimi: I'm pretty sure these are supposed to be links. If you look at jujucharms.com you'll see that they link to the file, but the text is grey like it is in the mockup.

jujugui/static/gui/src/app/components/entity-files/entity-files.js
```diff
@@ -0,0 +1,95 @@
...
+    render: function() {
+      var entityModel = this.props.entityModel;
+      var files = entityModel.get('files');
+      var archiveUrl = 'https://api.jujucharms.com/charmstore/v4/' +
+                       `${entityModel.get('full_name')}/archive`;
+      var codeUrl = entityModel.get('code_source').location;
+      var codeLink;
+      if (codeUrl) {
+        codeUrl = codeUrl.replace('lp:', 'https://code.launchpad.net/');
+        codeLink = (
```
huwshimi: Instead of this render method being so big you could move this to a separate generate method.

jujugui/static/gui/src/app/components/entity-files/entity-files.js
```diff
@@ -0,0 +1,95 @@
...
+              className="section__actions--launchpad"
+              href={codeUrl}>
+              View code
+            </a>
+          </li>
+        );
+      } else {
+        codeLink = '';
+      }
+      var fileItems = files.map(function(file) {
```
huwshimi: These could also get moved to a generate method to prevent this render from having too much logic in it.

jujugui/static/gui/src/app/components/entity-files/test-entity-files.js
```diff
@@ -0,0 +1,81 @@
...
+
+  beforeEach(function() {
+    mockEntity = jsTestUtils.makeEntity(false, ['foo.zip']);
+  });
+
+  afterEach(function() {
+    mockEntity = undefined;
+  });
+
+  it('renders a list of files', function() {
```
huwshimi: There are a couple of other code paths for when there are no files, no code link etc. that you could add tests for.

jujugui/static/gui/src/app/components/entity-files/test-entity-files.js
```diff
@@ -0,0 +1,81 @@
...
+              Download .zip
+            </a>
+          </li>
+        </ul>
+        <ul className="section__list">
+          {fileItems}
+        </ul>
+      </div>
+    );
+    jsTestUtils.compare(output, expected);
```
huwshimi: We should probably add this to our linter so we don't forget to remove it before commit. I'll add a card.

huwshimi:
Looks good @kadams54 with a few minor things:
- I think the box is supposed to have a 2px border radius
- looks like the top and bottom padding on the box is a bit small
- the heading font size looks a bit small
- the file names should be links (there's probably not a whole point listing the files if they're not links)

:+1: with those changes.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3674/
Test PASSed.

jujugui/static/gui/src/app/components/entity-files/_entity-files.scss
hatched: While one of the benefits is markup changes I think that's secondary to the removal of the cascade effect. Because of these styles if there were any sub components which implemented a ul/li they would inherit these styles. I've spent so much time overwriting styles that are much to generic that I always try to er on the side of caution.

jujugui/static/gui/src/app/components/entity-files/entity-files.js
```diff
@@ -0,0 +1,110 @@
...
+    },
+
+    /**
+      If able, generates a link to the entity's source code.
+
+      @method _generateCodeLink
+      @param {Object} codeSource metadata about the entity's source code.
+      @return {Object} The markup for the link to the code.
+    */
+    _generateCodeLink: function(codeSource) {
```
hatched: you can use default params here to avoid line 38

`_generateCodeLink: function(codeSource = {}) {`

jujugui/static/gui/src/app/components/entity-files/entity-files.js
```diff
@@ -0,0 +1,110 @@
...
+      var apiUrl = 'https://api.jujucharms.com/charmstore/v4';
+      var archiveUrl = `${apiUrl}/${entityModel.get('full_name')}/archive`;
+      var codeSource = entityModel.get('code_source');
+      return (
+        <div className="entity-files section" id="files">
+          <h3 className="section__title">
+            {files.length + ' ' + this.props.pluralize('file', files.length)}
+          </h3>
+          <ul className="entity-files__links">
+            {this._generateCodeLink(codeSource)}
```
hatched: `this._generateCodeLink(entityModel.get('code_source'));`

jujugui/static/gui/src/app/components/entity-files/test-entity-files.js
```diff
@@ -0,0 +1,103 @@
...
+  });
+
+  it('excludes the code link when url is not present', function() {
+    mockEntity.set('code_source', null);
+    var output = testUtils.renderIntoDocument(
+      <juju.components.EntityFiles
+        entityModel={mockEntity}
+        pluralize={sinon.spy()} />
+    );
+    assert.notOk(output.refs.codeLink);
```
hatched: I think that this should assert on `undefined` because `notOk` will still pass for `false` which would actually be invalid in this case.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3675/
Test PASSed.

jujugui/static/gui/src/app/components/entity-files/entity-files.js
kadams54: The point here is to make sure that `codeSource.location` never fails because codeSource is `null`; even if `null` were explicitly passed in, we need to override it to `{}`.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3676/
Test PASSed.

jujugui/static/gui/src/app/components/entity-content/test-entity-content.js
hatched: ahh right right right, sorry

jujugui/static/gui/src/app/components/entity-files/entity-files.js
hatched: ahh ok

hatched:
:+1: QA :ok: Thanks!

kadams54:
:shipit:

jujugui:

PR #1017, opened 2015-11-26
Fix svg recolour

huwshimi:
Fixed the close icon to allow it to be coloured as required.


huwshimi:
QA:
- open the mid-point
- check that the close icon is grey
- run `app.db.notifications.add({ title: 'yeah', description: 'here is a really really really really really long status message', level: 'error' })`
- check that the close icon is white

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/3700/
Test FAILed.

PR #1024, opened 2015-11-30
update npm to the latest in the 2.x line

hatched:
The 2.x line of npm allows us to use the download cache instead of going out to npm to download the dependencies. This _may_ also have the side effect of solving the `cb() not called` issue.


hatched:
#### To QA

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/3725/
Test FAILed.

huwshimi:
QA OK :+1: nice work.

jrwren:
This is npm 2.4.13 on node 0.10.37 from theblues/ppa, will it help our move to nodejs 4.2.x ? I'm afraid of sudo npm install -g, which is why I wonder if moving to nodejs 4.2.x would make that sudo npm install -g unnecessary.

hatched:
@jrwren yes it's definitely better if we can move to the 4.2.x series but that means that we'll have to keep the yellow ppa up to date with all of the node releases for all of the supported versions of Ubuntu which we don't have a plan for yet. I felt that this was the easiest next step.

Moving to 4.2.x will make the npm install unnecessary but only because it's already being installed globally by default. Essentially what that command is doing is using npm to update itself.

bac:
QA-OK

I'm also bothered by the npm update.  But, as we discussed, a deploy via the charm will never see npm so it only affects dev environments.  Code is :+1:

hatched:
Thanks for the qa's everyone! A successful passing CI run http://ci.jujugui.org:8080/job/juju-gui/3734/

:shipit:

jujugui:

jujugui:
        build url: http://ci.jujugui.org:8080/job/juju-gui-merge/1271

jujugui:

PR #1043, opened 2015-12-03
Machine removal

huwshimi:
Allow machines and containers to be removed.

### DIFF
--- a/jujugui/static/gui/src/app/models/models.js
+++ b/jujugui/static/gui/src/app/models/models.js
@@ -2757,13 +2757,7 @@ YUI.add('juju-models', function(Y) {
 
       var changedMachines = [];
       this.machines.each(function(machine) {
-        var units = machine.units;
-        if (!units) {
-          // If the machine view hasn't been opened yet then there won't be a
-          // units collection on the machines. So we need to get the 'real'
-          // unit list.
-          units = this.units.filterByMachine(machine.id, true);
-        }
+        var units = this.units.filterByMachine(machine.id, true);
         var keepVisible = this._highlightedServices.some(
             function(highlightedService) {
               return units.some(function(unit) {
--- a/jujugui/static/gui/src/app/utils/environment-change-set.js
+++ b/jujugui/static/gui/src/app/utils/environment-change-set.js
@@ -711,7 +711,8 @@ YUI.add('environment-change-set', function(Y) {
       var db = this.get('db');
       var machine = db.machines.getById(command.args[0]);
       var removedUnits = [];
-      machine.units.forEach(function(unit) {
+      var units = db.units.filterByMachine(machine.id);
+      units.forEach(function(unit) {
         if (!unit.agent_state) {
           // Remove the unit's machine, making it an unplaced unit.
           delete unit.machine;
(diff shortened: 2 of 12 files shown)


huwshimi:
QA:
- use the flag /:flags:/mv/
- add a service
- open the machine view
- auto place the unit
- open the context menu on the machine and click destroy
- the machine should be destroyed

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/3786/
Test FAILed.

jujugui/static/gui/src/app/models/models.js
```diff
@@ -2757,13 +2757,7 @@ YUI.add('juju-models', function(Y) {
...
       var changedMachines = [];
       this.machines.each(function(machine) {
-        var units = machine.units;
-        if (!units) {
-          // If the machine view hasn't been opened yet then there won't be a
-          // units collection on the machines. So we need to get the 'real'
-          // unit list.
-          units = this.units.filterByMachine(machine.id, true);
-        }
+        var units = this.units.filterByMachine(machine.id, true);
```
huwshimi: Removed this kinda hacky unit list, we can just do a lookup using the machine id.

jujugui/static/gui/src/app/utils/environment-change-set.js
```diff
@@ -711,7 +711,8 @@ YUI.add('environment-change-set', function(Y) {
       var db = this.get('db');
       var machine = db.machines.getById(command.args[0]);
       var removedUnits = [];
-      machine.units.forEach(function(unit) {
+      var units = db.units.filterByMachine(machine.id);
+      units.forEach(function(unit) {
```
huwshimi: As above, removed this kinda hacky unit list which relied on the machine view having been opened before a machine was destroyed, we can just do a lookup using the machine id.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3787/
Test PASSed.

hatched:
QA Issue:
 
- I auto placed a unit and then before committing destroyed the machine. It left the UI representation for the 'root container' in the right column. Closing and re-opening the mv fixed it.

- There is no UI update to the machine when it is pending destroy. I assumed this would turn blue or something.

- If I commit the machines and then destroy it. It leaves both the machine and root container visible but with no units in it. Closing and re-opening the mv does not fix this. It appears that it does not get deleted.

huwshimi:
@hatched it looks like we were never listening to machine events correctly. I've fixed that first QA issue. There is a card to add uncommitted states in the MV. The last issue is related to the email thread about properly destroying machines from the GUI.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3810/
Test PASSed.

jujugui/static/gui/src/app/widgets/autodeploy-extension.js
```diff
@@ -75,7 +75,6 @@ YUI.add('autodeploy-extension', function(Y) {
         parentId: parentId,
         constraints: constraints || {}
       }], callback, {modelId: machine.id});
-      db.fire('update');
```
huwshimi: This is a change I had previously made to make the UI update on machine creation, this has now been replaced with the listener in app.js.

hatched:
QA :ok:
Thanks for making that fix!

hatched:
:+1:

huwshimi:
Thanks for the review @hatched :shipit:!

jujugui:

PR #1065, opened 2015-12-10
Fix switch env

jaycee:


jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/3859/
Test FAILed.

jaycee:
QA:

This requires use of the juju gui charm as well as this branch.

Checkout this branch and package it with `make clean-all fast-babel dist`.
Checkout juju/juju-gui-charm, and prep it for deploy with `make clean package`.
Copy the file created by make juju-gui (dist/jujugui-1.9.2.tar.gz) into the releases folder in the charm directory.
Bootstrap your environment of choice (not local).
In the charm directory, run `make deploy`.
Verify in the gui (once it comes up) that you can login, create a new env, and switch between environments.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3860/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3861/
Test PASSed.

jujugui/static/gui/src/app/app.js
```diff
@@ -1637,12 +1637,15 @@ YUI.add('juju-gui', function(Y) {
...
       }
-      // Set the credentials so the GUI will automatically log in when we
-      // switch the environments.
-      this.env.setCredentials({
-        user: username,
-        password: password
-      });
+      if (this.jem) {
+        // Set the credentials so the GUI will automatically log in when we
+        // switch the environments. We only switch if using JEM where user/pass
```
frankban: Maybe " We only set credentials"?
Otherwise this is quite confusing, as we can also switch between models in the controller scenario without jem.

jujugui/static/gui/src/app/app.js
```diff
@@ -1654,12 +1657,14 @@ YUI.add('juju-gui', function(Y) {
       }
 
       var socketUrl = this.env.get('socket_url');
-      // XXX frankban: this is not generic, and very specific for how the
-      // socket URL is composed in the GUI embedded in OpenStack. Therefore,
-      // the logic here for calculating the new socket URL for the given UUID
-      // must be considered temporary demo code.
-      var baseUrl = socketUrl.substring(0, socketUrl.lastIndexOf('/'));
-      var newSocketUrl = baseUrl + '/' + uuid;
+      var newSocketUrl, baseUrl;
```
frankban: I guess this is ok for now, but it will likely need to be reworked later.
The switchEnv method does not yet support multiple controllers (it does not receive an address).
The embedded flag semantic is more similar to embeddedInHorizon than embeddedInGeneral, as we don't know about available URLs in public clouds. What I am thinking is something in which we pass a template for the WebSocket URL in config. Like `config.socketTemplate = '/juju/api/$address/$port/$uuid'` or `config.socketTemplate = '/ws/environment/$uuid/api'`. In both cases those are paths of 'wss://$currenthost:$currentport`, which can be easily retrieved. In config, the GUI takes the socket template and the initial URL (either as the current socket URL or as separate options like socketAddress and socketUUID). The switchEnv is changed to receive a "host:port" address. At that point you can use host, port and uuid as a context to render the socket URL based on the socketTemplate in a generic way, and the embedded is no longer required. This would also solve the current issue in _generateSocketUrl: if there is a jem, then /juju/api is always used, but a jem URL could be provided even if we are not embedded in horizon. That function needs to change as well, and having a separate renderSocketUrl(template, host, port, uuid) would be great. What do you think? I am not saying this must be done now, and what we have here is surely a step forward. But if you agree with my plan, please let's add at least an XXX describing it.

jujugui/static/gui/src/test/test_app.js
```diff
@@ -1056,6 +1056,136 @@ describe('File drag over notification system', function() {
...
+      container = Y.Node.create('<div id="test" class="container"></div>');
+    });
+
+    afterEach(function() {
+      if (app) {
+        app.destroy({remove: true});
+      }
+    });
+
+    it('does not switch envs in sandbox', function() {
```
frankban: Thanks for adding tests for switchEnv! And sorry that you didn't find them already: the switchEnv was initially intended as demo code, thanks for making it production ready code.

frankban:
:+1: with a couple of suggestions, including a non-minor one. Thanks for fixing this JC!
QA ok. I remember we used to switch to the new model right after it was created. This is no longer the case: any reason for that? UX decision or just a regression on the widget?

frankban:
Out of curiosity, why did you suggest to avoid local envs? I used local to QA without problems.

urosj:
Models can be created and switched to/from. GUI doesn't switch to them automatically anymore. 

Possible bug: I had two browsers open on the same model. The first one had uncommitted state. The second one was empty. Once i committed the changes, the environment switcher in the second one failed to respond or switch between models until i reloaded the gui in the browser.

This PR looks good, but we need to switch to env directly and check more on the "concurrent" multi user work in a model.

frankban:
Great QA Uros, and interesting bug. I wonder if that's related to the fact that we never close the mega-watcher, and so there is a race between switching connections and incoming data from the mega-watcher (as a result of a change in the state, like committing a bundle). By closing the mega-watcher I mean sending a request like `{Type: 'AllWatcher', Request: 'Stop', Id: ...}`. Anyway, this is just an idea, might be totally unrelated.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/3874/
Test FAILed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/3875/
Test FAILed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/3876/
Test FAILed.

jaycee:
Passing run at http://ci.jujugui.org:8080/job/juju-gui/3879/console

jujugui/static/gui/src/app/app.js
jaycee: Comment added.

jujugui/static/gui/src/app/app.js
jaycee: THat is clearer, thanks. Fixed.

jaycee:
:shipit:

jujugui:

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/3882/
Test FAILed.

jujugui:
        build url: http://ci.jujugui.org:8080/job/juju-gui-merge/1328

jujugui:

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3883/
Test PASSed.

PR #1091, opened 2015-12-16
Cross model relations: offer details and remote service info.

frankban:
Implement the missing ServiceOffers API call on the CrossModelRelations
facade. It is used to retrieve details on a offered remote service.

Also implement the model layer for remote services. It is represented
in the database by a RemoteService model, and its corresponding
model list.

A mega-watcher handler is also introduced in order to populate the
new models. As juju-core does not yet implement relating to remote
services it is not possible to really test the mega-watcher handler
at this time, but I followed to Go code and that logic should be
ready very soon anyway.

Finally, added an addDetails method on the RemoteService that can be
use to enhance the model instance (created on delta) with info
obtained from calling env.getOffer().

Note that sandbox mode is not supported yet.

I expect small tweaks will be required when the functionality is
completed in core, but for now this should be enough for proceeding
with visualization steps.

### DIFF
--- a/jujugui/static/gui/src/app/models/handlers.js
+++ b/jujugui/static/gui/src/app/models/handlers.js
@@ -236,6 +236,34 @@ YUI.add('juju-delta-handlers', function(Y) {
     },
 
     /**
+      Handle remote service info coming from the juju-core delta, updating the
+      relevant database models.
+
+      @method remoteserviceInfo
+      @param {Object} db The app.models.models.Database instance.
+      @param {String} action The operation to be performed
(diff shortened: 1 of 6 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3926/
Test PASSed.

hatched:
:+1:  Looks great! Can't wait to actually use it :)

frankban:
Thanks for the review!
:shipit:

jujugui:

PR #1114, opened 2015-12-18
Fix visibility toggles

kadams54:


kadams54:
# QA

1. Deploy the MongoDB cluster bundle, as well as another service (e.g., the Wordpress charm).
2. Go all chaos monkey on the visibility toggles. At no point should you be able to make the "fade" and "focus" modes active for the same service; turning one on should turn the other off.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3968/
Test PASSed.

huwshimi:
QA OK :+1: nicely done @kadams54!

jujugui/static/gui/src/app/components/added-services-list-item/test-added-services-list-item.js
```diff
@@ -25,12 +25,44 @@ chai.config.includeStack = true;
...
 describe('AddedServicesListItem', function() {
+  var mockService;
 
   beforeAll(function(done) {
     // By loading this file it adds the component to the juju components.
     YUI().use('added-services-list-item', function() { done(); });
   });
 
+  beforeEach(function() {
+    mockService = {
```
hatched: Lets move this to `component-test-utils.js` so that we can enhance/reuse elsewhere.

jujugui/static/gui/src/app/components/added-services-list-item/test-added-services-list-item.js
```diff
@@ -435,22 +445,13 @@ describe('AddedServicesListItem', function() {
       <juju.components.SvgIcon name="focused_16" size="16"/>);
     assert.deepEqual(
       output.props.children[4].props.children[1].props.children,
-      <juju.components.SvgIcon name="hide_16" size="16"/>);
+      <juju.components.SvgIcon name="show_16" size="16"/>);
```
hatched: was the old test incorrect?

jujugui/static/gui/src/app/components/added-services-list-item/test-added-services-list-item.js
```diff
@@ -563,29 +530,17 @@ describe('AddedServicesListItem', function() {
         unfadeService={unfadeService}
         changeState={sinon.stub()}
         getUnitStatusCounts={sinon.stub()}
-        service={service} />);
+        service={mockService} />);
 
-    // Check that the container has the proper classes when no visibility
-    // toggle is active.
-    assert.equal(
```
hatched: Why are we no longer checking for these classes?

jujugui/static/gui/src/app/components/added-services-list/test-added-services-list.js
```diff
@@ -103,35 +103,52 @@ describe('AddedServicesList', () => {
...
-      get: () => 'trusty/apache',
-      set: sinon.stub(),
-      setAttrs: sinon.stub()
-    }, {
-      get: () => 'trusty/mysql',
-      set: sinon.stub(),
-      setAttrs: sinon.stub()
-    }];
-    var services = {
+  function createService(attrs) {
```
hatched: Maybe you can combine this and the test util above and move it into the component test utils.

jujugui/static/gui/src/app/components/added-services-list/test-added-services-list.js
```diff
@@ -103,35 +103,52 @@ describe('AddedServicesList', () => {
...
+        }
+      }
+    };
+  }
+
+  function createModelList(models){
+    return {
       each: (cb) => {
-        allServices.forEach(cb);
+        models.forEach(cb);
```
hatched: Because an 'environment' is now a 'model' I've been trying to avoid calling things models :)

jujugui/static/gui/src/app/components/added-services-list/test-added-services-list.js
```diff
@@ -308,26 +287,23 @@ describe('AddedServicesList', () => {
...
-    instance.fadeService('trusty/wordpress');
-    assert.equal(serviceSet.callCount, 1);
-    assert.deepEqual(serviceSet.args[0], ['fade', true]);
+    var unfocusService = sinon.stub(instance, 'unfocusService');
+    instance.fadeService(id);
+    // Set the fade switch appropriately.
+    assert.equal(allServices[0].get('fade'), true);
+    // Unfocused the service.
+    assert.isTrue(unfocusService.calledWith(id));
+    unfocusService.restore();
```
hatched: I'm not sure if this is actually a problem with these tests....but if there is a failure between when this is stubbed out and when it's restored, it will never be restored. Could this be tested without stubbing that fn to avoid any potential issues?

jujugui/static/gui/src/app/components/added-services-list/test-added-services-list.js
kadams54: I called them "models" because that's what they are, architecturally - objects that model and provide access to our datastore. It has nothing to do (and should have nothing to do) with whatever terminology is used in the UI.

jujugui/static/gui/src/app/components/added-services-list-item/test-added-services-list-item.js
kadams54: Because the classes don't change anymore within the scope of these tests. All the AddedServicesListItem object does now is call the injected methods (focusService, unfocusService, etc.). These methods are now stubbed out and asserted to make sure they're being called with the right data, but since they never execute, the ListItem component doesn't re-render, doesn't get new classes, etc.

That sort of testing now takes place in `test-added-services-list.js`, in the "…performs the necessary work to…" tests.

jujugui/static/gui/src/app/components/added-services-list-item/test-added-services-list-item.js
kadams54: No, but I changed the test slightly (see lines 427 and 428) to test one toggle (focus/highlight) on true and the other (fade) on false.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4007/
Test PASSed.

jujugui/static/gui/src/app/components/added-services-list/test-added-services-list.js
kadams54: If it isn't stubbed (i.e., just use `sinon.spy()`) then a bunch of stuff happens in order to unfocused the service, which would in turn need to be mocked/stubbed in order for the test to pass. Best to stub. I removed the restore as I don't think it's actually a problem.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4008/
Test PASSed.

jujugui/static/gui/src/app/components/added-services-list/test-added-services-list.js
hatched: discussed on irc

hatched:
:+1: QA :ok:

kadams54:
Thanks all! :shipit:

jujugui:

PR #1125, opened 2015-12-19
Display the unit icons as circles in the machine view

huwshimi:


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/3984/
Test PASSed.

hatched:
:+1: :ok: :shipit: 
It's unfortunate that every icon has that little black line at the top of the circle now. Hopefully we can update those icons soon.

jujugui:

jujugui/static/gui/src/app/components/machine-view/machine/_machine.scss
```diff
@@ -95,6 +89,14 @@
         margin-right: 10px;
         padding: 2px;
         border: 1px solid $canvas-background;
-        border-radius: 6px;
+        border-radius: 50%;
+
+        img {
```
anthonydillon: We shouldn't target the element here. As a component we shouldn't only target classes as the element may change but we don't want the styling to be lost. For example this element maybe be changed to picture in the future which with require two edits.

PR #1143, opened 2015-12-21
Force machine removal when destroying from MV

### LINKED ISSUE #1104: Add force functionality to remove units when destroying machine
When using force to destroy a machine we have to then cascade that down to mark all units and containers on the machine recursively in the ecs and then add the ability to revert those changes.

hatched:
When deleting machines from the machine view it's assumed that the user understands that they will destroy all child machines and units. This implements that functionality. (Fixes #1104)

### DIFF
--- a/jujugui/static/gui/src/app/utils/environment-change-set.js
+++ b/jujugui/static/gui/src/app/utils/environment-change-set.js
@@ -711,18 +711,21 @@ YUI.add('environment-change-set', function(Y) {
       var db = this.get('db');
       var machine = db.machines.getById(command.args[0]);
       var removedUnits = [];
-      var units = db.units.filterByMachine(machine.id);
+      var units = db.units.filterByMachine(machine.id, true);
       units.forEach(function(unit) {
+        // Update the revived model to trigger events.
+        var unitModel = db.units.revive(unit);
         if (!unit.agent_state) {
           // Remove the unit's machine, making it an unplaced unit.
           delete unit.machine;
           removedUnits.push(unit);
-
-          // Update the revived model to trigger events.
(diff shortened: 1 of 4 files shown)


hatched:
#### To QA

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4004/
Test PASSed.

jujugui/static/gui/src/app/utils/environment-change-set.js
```diff
@@ -505,7 +506,16 @@ YUI.add('environment-change-set', function(Y) {
...
         case '_destroyMachines':
+          var machineId = command.args[0][0];
+          // Set the parent machine to not be deleted.
           machines.getById(command.args[0]).deleted = false;
+          // Set the containers of the parent machine to not be deleted.
+          machines.filterByAncestor(machineId).forEach(machine => {
+            machine.deleted = false;
+          });
+          units.filterByMachine(machineId, true).forEach(unit => {
+            unit.deleted = false;
```
huwshimi: Is it OK to update the units/machines etc. here without reviving them? I guess the db events must get fired correctly otherwise this wouldn't work...

hatched: Reviving makes every attribute change fire, getters, setters run etc. So this is actually an optimization because it only fires the change event (above L485) once all have been updated.

huwshimi: Cool, nice.

bac:
QA-OK -- thanks

hatched:
Thanks! :shipit:

jujugui:

PR #1162, opened 2016-01-05
Do not hide the service you're creating a relation from

### LINKED ISSUE #1153: On relation making, all units disappear
When a unit doesn't have any other unit to relate to, the unit from which we want to create a relation disappears as well.

hatched:
When creating a relation it should hide services you cannot relate to but leave the source service still visible. Fixes #1153

### DIFF
--- a/jujugui/static/gui/src/app/views/topology/relation.js
+++ b/jujugui/static/gui/src/app/views/topology/relation.js
@@ -872,7 +872,7 @@ YUI.add('juju-topology-relation', function(Y) {
               .classed('selectable-service', true)
               .filter(function(d) {
                 return (d.id in invalidRelationTargets &&
-                          d.id !== service.id);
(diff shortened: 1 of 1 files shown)


hatched:
#### To QA

Deploy a few charms that can and can not relate. Create a relation, it should hide the ones that are not relate-able. 

Note: 
This is missing a test update because the issue was caused by data being sent from a long ways away so the proper test would be a selenium test. We'll have to add that at a later time.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4066/
Test PASSed.

kadams54:
:+1: on the one line of code. Whew!

QA is :ok: - nicely done!

hatched:
Thanks! :shipit:

jujugui:

PR #1183, opened 2016-01-08
Get username from macaroon for breadcrumb and jem url setting.

jaycee:


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4095/
Test PASSed.

jujugui/static/gui/src/app/app.js
```diff
@@ -1011,11 +1011,82 @@ YUI.add('juju-gui', function(Y) {
...
+        });
+      });
+      return username;
+    },
+
+    /**
+      Get auth object, or if unavailable get mocked object with username from
+      cookie.
+
+      @method _getAuth
```
frankban: @return parameter? [shrug]

jujugui/static/gui/src/app/app.js
```diff
@@ -1011,11 +1011,82 @@ YUI.add('juju-gui', function(Y) {
     },
 
     /**
+      Get user name from cookie.
+
+      @method _getUsernameFromCookie
+    */
+    _getUsernameFromCookie: function() {
+      // XXX jcsackett 2016-01-08: We should be using the whoami endpoint on
+      // charmstore to get the username of a logged in jem user rather than
```
frankban: the whoami endpoint on jem

jujugui/static/gui/src/app/app.js
```diff
@@ -1011,11 +1011,82 @@ YUI.add('juju-gui', function(Y) {
...
+      var cookie = document.cookie;
+      var data = cookie.split(';');
+      var macaroon = null;
+      var prefix = 'Macaroons-jem=';
+      var found = data.some(function(item) {
+        macaroon = item;
+        return (item.indexOf(prefix) !== -1);
+      });
+      if (!found) {
+        console.error('No macaroon found to get username from.');
```
frankban: Is this an error or just a log info?

frankban:
:+1: with minor suggestions.
Thank you!

jujugui/static/gui/src/app/app.js
```diff
@@ -1011,11 +1011,82 @@ YUI.add('juju-gui', function(Y) {
...
+      if (this.get('sandbox')) {
+        return null;
+      }
+      var cookie = document.cookie;
+      var data = cookie.split(';');
+      var macaroon = null;
+      var prefix = 'Macaroons-jem=';
+      var found = data.some(function(item) {
+        macaroon = item;
+        return (item.indexOf(prefix) !== -1);
```
rogpeppe: If it's a prefix, shouldn't it be === 0 ?

Also, I'm thinking that the outer brackets are probably unnecessary here.

jujugui/static/gui/src/app/app.js
```diff
@@ -1011,11 +1011,82 @@ YUI.add('juju-gui', function(Y) {
...
+      // Don't look for cookies if in sandbox.
+      if (this.get('sandbox')) {
+        return null;
+      }
+      var cookie = document.cookie;
+      var data = cookie.split(';');
+      var macaroon = null;
+      var prefix = 'Macaroons-jem=';
+      var found = data.some(function(item) {
+        macaroon = item;
```
rogpeppe: Why not just use data.find ?

jujugui/static/gui/src/app/app.js
```diff
@@ -1011,11 +1011,82 @@ YUI.add('juju-gui', function(Y) {
...
+      macaroon = macaroon.replace('Macaroons-jem=', '');
+      macaroon = JSON.parse(atob(macaroon));
+      var username = null;
+      // We check all the caveats of each piece of the macaroon for the username
+      // data. If we find more than one username we abandon the macaroon as
+      // invalid.
+      macaroon.forEach(function(piece) {
+        var caveats = piece.caveats;
+        caveats.forEach(function(caveat) {
+          var declaration = 'declared username ';
```
hatched: This is going to need a bunch more comments :) I see what is happening here but I have no idea why...

jujugui/static/gui/src/app/app.js
```diff
@@ -1011,11 +1011,82 @@ YUI.add('juju-gui', function(Y) {
...
+      var prefix = 'Macaroons-jem=';
+      var found = data.some(function(item) {
+        macaroon = item;
+        return (item.indexOf(prefix) !== -1);
+      });
+      if (!found) {
+        console.error('No macaroon found to get username from.');
+        return null;
+      }
+      var index = macaroon.indexOf(prefix) + prefix.length;
```
rogpeppe: I'm probably not understanding this, but ISTM you're calculating the index only to throw
it away. (macaroon.slice on the line below is a no-op, right?)

Couldn't you do:

    macaroon = macaroon.slice(index)

and lose the macaroon.replace line?

jujugui/static/gui/src/app/app.js
```diff
@@ -1011,11 +1011,82 @@ YUI.add('juju-gui', function(Y) {
...
+      macaroon.slice(index);
+      macaroon = macaroon.replace('Macaroons-jem=', '');
+      macaroon = JSON.parse(atob(macaroon));
```
hatched: If this in incorrectly formatted the app will fall over here. This should probably be in a try/catch

jujugui/static/gui/src/app/app.js
```diff
@@ -1011,11 +1011,82 @@ YUI.add('juju-gui', function(Y) {
...
+      cookie.
+
+      @method _getAuth
+    */
+    _getAuth: function() {
+      var auth = this.get('auth');
+      if (auth === null) {
+        var username = this._getUsernameFromCookie();
+        if (username !== null) {
+          auth = { user: { name: username }};
```
hatched: since we're mocking it we could really put it in any format. Maybe it should have a comment here saying that it's mocked to mirror the api

jujugui/static/gui/src/app/app.js
```diff
@@ -1011,11 +1011,82 @@ YUI.add('juju-gui', function(Y) {
...
+      // We check all the caveats of each piece of the macaroon for the username
+      // data. If we find more than one username we abandon the macaroon as
+      // invalid.
+      macaroon.forEach(function(piece) {
+        var caveats = piece.caveats;
+        caveats.forEach(function(caveat) {
+          var declaration = 'declared username ';
+          if (caveat.cid.indexOf(declaration) !== -1) {
+            if (username === null) {
+              username = caveat.cid.replace(declaration, '');
```
rogpeppe: or:
    username = caveat.cid.slice(index + declaration.length)
?

jujugui/static/gui/src/app/app.js
rogpeppe: If it was to mirror the whoami API, it would be in the form { User: username }

jujugui/static/gui/src/app/app.js
```diff
@@ -1011,11 +1011,82 @@ YUI.add('juju-gui', function(Y) {
...
+      // XXX jcsackett 2016-01-08: We should be using the whoami endpoint on
+      // charmstore to get the username of a logged in jem user rather than
+      // parsing the cookie. Consider this all a bad hack for demo and we'll do
+      // something better in the very near future.
+
+      // Don't look for cookies if in sandbox.
+      if (this.get('sandbox')) {
+        return null;
+      }
+      var cookie = document.cookie;
```
hatched: instead of grabbing the document here...

create a method:
```Javascript
_getCookies: function() {
  document.cookie;
}
```

then this becomes

```Javascript
var cookie = this._getCookies();
```

And in the tests you can then stub out `_getCookies`

jujugui/static/gui/src/app/app.js
rogpeppe: That's presumably also true even if it's valid JSON. Say, for example it was a JSON string not an object,
then I'm guessing that the forEach loop below would fall over. It might be better to use the macaroons library to parse it, which should do appropriate verification. Or just put a try/catch around the whole function.

rogpeppe:
LGTM with a few minor suggestions and +1 to hatch's try/catch suggestion.

jujugui/static/gui/src/app/app.js
frankban: Let's keep in mind that this code will be removed later, probably soon.

jujugui/static/gui/src/app/app.js
frankban: I think this mirrors what set by horizon juju gui in auth. And AFAICT it looks good.

jujugui/static/gui/src/app/app.js
jaycee: Done.

jujugui/static/gui/src/app/app.js
jaycee: Updated.

jujugui/static/gui/src/app/app.js
jaycee: They're log info; I've updated both instances.

jujugui/static/gui/src/app/app.js
jaycee: -1 indicates nothing was found, so this says the prefix was found. We can't just look for an instance of prefix being found at 0, since there can be an unknown amount of leading space.

jujugui/static/gui/src/app/app.js
jaycee: data.find would require us to know the whole value of the item; we only know it will start with arbitrary whitespace and the prefix.

jujugui/static/gui/src/app/app.js
jaycee: Yes, this was just badly fixed up as I played around with how best to do it. macaroon.slice is the right approach.

jujugui/static/gui/src/app/app.js
jaycee: It's mocked to mirror the structure of the auth object in config.js; I've updated the code to say as much.

jujugui/static/gui/src/app/app.js
jaycee: I've added try/catch around the json and subsequent logic.

jujugui/static/gui/src/app/app.js
jaycee: Yeah, that could be better if we run into the same arbitrary whitespace issue as in the initial cookie parsing.

jujugui/static/gui/src/app/app.js
jaycee: Thanks, this made tests much easier.

jaycee:
QA:

Sandbox:

Run in development; everything should work normally. User breadcrumb should be anonymous.

JEM:

Deploy the gui with a package built from this branch and point it a guimaas's jem (assuming you have added the necessary state-servers and templates to guimaas to be able to use it). Confirm that the username is shown as your usso name in the breadcrumb. Confirm you can make new environments.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4097/
Test FAILed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4098/
Test FAILed.

frankban:
QA:
sandbox works.
In jes, the breadcrumbs user name is still "anonymous" and that's expected, but could you please hardcode "admin" instead of anonymous? it woild work better as a qucik fix while waiting for the proper one.
In jem, the username is "declared username frankban" (without quotes). Therefore creating new models does not work. This is just a parsing issue AFAIK, but should be captured in tests.
I'll QA again once those are fixed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4102/
Test PASSed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4103/
Test FAILed.

frankban:
QA ok.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4104/
Test PASSed.

jaycee:
:shipit:

Thanks for the reviews, all.

jujugui:

PR #1191, opened 2016-01-08
Set autoplace units to be the default on commit

### LINKED ISSUE #1169: First commit/deploy in admin model doesn't select 'automatically place'
In a real JES enabled model I deployed Wordpress, when to commit, and then deploy. I noticed however that it still had changes. It didn't auto-select "Automatically Place" as the default. This could be confusing to new users so we should make sure that it's the default.

hatched:
Having auto-place of units not be the default action was causing people to be confused as to why their services weren't starting. This sets auto-place as the default action, people in the know are still able to disable it if desired. Fixes #1169

### DIFF
--- a/jujugui/static/gui/src/app/components/deployment/deployment.js
+++ b/jujugui/static/gui/src/app/components/deployment/deployment.js
@@ -36,7 +36,7 @@ YUI.add('deployment-component', function() {
       // Setting a default state object.
       var state = {
         hasCommits: false,
-        autoPlace: !!localStorage.getItem('auto-place-default')
+        autoPlace: !localStorage.getItem('auto-place-default')
       };
       return this.generateState(this.props, state);
     },
--- a/jujugui/static/gui/src/app/components/deployment/deployment.js
+++ b/jujugui/static/gui/src/app/components/deployment/deployment.js
@@ -36,7 +36,7 @@ YUI.add('deployment-component', function() {
       // Setting a default state object.
       var state = {
         hasCommits: false,
-        autoPlace: !localStorage.getItem('auto-place-default')
+        autoPlace: !localStorage.getItem('disable-auto-place')
       };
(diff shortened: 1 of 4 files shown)


hatched:
#### To QA

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4106/
Test PASSed.

jujugui/static/gui/src/app/components/deployment/deployment.js
```diff
@@ -36,7 +36,7 @@ YUI.add('deployment-component', function() {
       // Setting a default state object.
       var state = {
         hasCommits: false,
-        autoPlace: !!localStorage.getItem('auto-place-default')
+        autoPlace: !localStorage.getItem('auto-place-default')
```
jaycee: Shouldn't this be "disable-auto-place" instead of "auto-place-default"?

kadams54:
The QA steps as outlined work fine, however, setting the prefs has no effect, per @jaycee's comments in the code.

*QA Notes:* I worked through the steps outlined above, but then added some extra steps, which are necessary to confirm that setting prefs works properly:

1. Hit Shift + ? to bring up the help/prefs dialog.
2. Check "Default to not automatically place units on commit."
3. Click on the Save button.
4. Add a service to the canvas and check the deployment summary.
5. The Autodeploy option should not be selected this time.

bac:
:+1:, no QA

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4112/
Test PASSed.

jaycee:
:+1: with that change, thanks @hatched.

kadams54:
QA is :ok:, :+1: on the code.

hatched:
Thanks all! :shipit:

jujugui:

PR #1210, opened 2016-01-14
Mark units as resolved on remove so that they can be removed if in error

### LINKED ISSUE #1204: Remove button not working when unit is in error state.
We discussed today and the semantic of destroy, if an unit is in an error state, is more like "please allow me to get rid of these units in an error state and never look back".
For this reason code in https://github.com/juju/juju-gui/blob/develop/jujugui/static/gui/src/app/components/unit-list/unit-list.js#L101 should also mark units as resolved before calling destroyUnits (only in the case units are in error).

hatched:
If a unit is in error then Juju won't remove it so we will mark it as resolved prior to removing. Fixes #1204

### DIFF
--- a/jujugui/static/gui/src/app/components/unit-list/test-unit-list.js
+++ b/jujugui/static/gui/src/app/components/unit-list/test-unit-list.js
@@ -314,6 +314,7 @@ describe('UnitList', () => {
   it('can remove the selected units', function() {
     var destroyUnits = sinon.stub();
     var changeState = sinon.stub();
+    var envResolved = sinon.stub();
     var units = [{
       displayName: 'mysql/0',
       id: 'mysql/0'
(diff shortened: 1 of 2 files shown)


hatched:
#### To QA

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4143/
Test PASSed.

huwshimi:
Code :+1:. Will try and QA a bit later if I have time before my EOD.

kadams54:
Code is :+1: should be finished with the QA shortly.

kadams54:
QA is :ok: with a caveat: everything worked as described; however, there was no indicator that anything was happening on the unit after the remove was confirmed until it actually disappeared. It would be nice to see that UX tightened up a bit in a future card.

hatched:
@kadams54 Yep I've actually already sent a request to UX about that - good to see others think the same :)

Thanks for the reviews and qa! :shipit:

jujugui:

jujugui:
        build url: http://ci.jujugui.org:8080/job/juju-gui-merge/1432

jujugui:

PR #1258, opened 2016-01-19
Add optional gzip compression (on by default).

jaycee:


jaycee:
QA:

Run `make run`; load the app while watching the network tab of your browser's debug tools. Note the compression in effect by the amount being transferred being less than the file size.

Edit development.ini to set `jujugui.gzip` to "false". Re run `make run` and again check the network tab--you should see transfer size and file size now be the same.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4202/
Test PASSed.

fabricematrat:
:+1:  Code :ok: Just wonder why we do it in the wsgi app instead of Apache with mod deflate.

jaycee:
@fabricematrat We don't actually provide the gui with apache in the charm--the guiserver directly serves the gui, so we need this option there.

In HJC, where we do have apache (and I believe openstack is already configured to use mod_deflate or could be) we'll turn this setting off.

hatched:
:+1: :ok:

Thanks this is awesome! With minification only 505KB(total) are now being sent over the wire :)

jaycee:
:shipit:

jujugui:

PR #1284, opened 2016-01-26
Replace juju_config.GA_KEY with juju_config.GTM_enabled

anthonydillon:


jujugui/options.py
```diff
@@ -20,7 +20,7 @@ def update(settings):
     _update(settings, 'jujugui.base_url', default=None)
     _update(settings, 'jujugui.charmstore_url', default=DEFAULT_CHARMSTORE_URL)
     _update(settings, 'jujugui.api_path', default=DEFAULT_API_PATH)
-    _update(settings, 'jujugui.ga_key', default='')
+    _update(settings, 'jujugui.GTM_enabled', default='')
```
jaycee: Default here should be 'False', as we want this to be a boolean setting.

hatched: Should this not have a boolean default value? `False`

jujugui/static/gui/src/app/config-debug.js
```diff
@@ -53,10 +53,10 @@ var juju_config = { // eslint-disable-line no-unused-vars
...
   simulateEvents: false,
   readOnly: false,
-  // Set the GA_key to enable Google Analytics usage and calls. Also implies
-  // using cookies. For the debug configuration, the GA_key should be blank to
-  // prevent muddying the gathered statistics.
-  GA_key: '',
+  // Set the GTM_enabled to enable Google Tag Manager usage and calls. Also
+  // implies using cookies. For the debug configuration, the GTM_enabled should
+  // be false to prevent muddying the gathered statistics.
+  GTM_enabled: false,
```
jaycee: I think we can actually remove config-debug.js and config-prod.js, can't we? @hatched?

jujugui/tests/test_options.py
```diff
@@ -15,7 +15,7 @@ class TestUpdate(unittest.TestCase):
         'jujugui.base_url': None,
         'jujugui.charmstore_url': options.DEFAULT_CHARMSTORE_URL,
         'jujugui.combine': True,
-        'jujugui.ga_key': '',
+        'jujugui.GTM_enabled': '',
```
jaycee: Again, as this is defaults we should do False, not ''.

jujugui/tests/test_views.py
```diff
@@ -98,7 +98,7 @@ def test_default_options(self):
         self.assertEqual('wss', config['socket_protocol'])
         self.assertEqual(
             options.DEFAULT_CHARMSTORE_URL, config['charmstoreURL'])
-        self.assertEqual('', config['GA_key'])
+        self.assertEqual('', config['GTM_enabled'])
```
jaycee: And then self.assertFalse(config['GTM_enabled'])

jujugui/static/gui/src/app/config-debug.js
hatched: That's correct - although out of scope for this branch.

jaycee:
@anthonydillon thanks, just a few fixes so we're using boolean values instead of empty strings.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4272/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4273/
Test PASSed.

jujugui/static/gui/src/app/app.js
```diff
@@ -2067,8 +2067,8 @@ YUI.add('juju-gui', function(Y) {
      *
      */
     authorizeCookieUse: function(req, res, next) {
-      var ga_key = this.get('GA_key');
-      if (ga_key) {
+      var GTM_enabled = this.get('GTM_enabled');
```
jaycee: I'm not sure this is going to work--GTM_enabled isn't an attr on app, so this.get shouldn't work.

That said GA_key wasn't an attr either--I'm not sure how this worked previously. @hatched, you have any ideas?

hatched: Anything that's set in `juju_config` will get applied to app.js so this should work.
https://github.com/juju/juju-gui/blob/develop/jujugui/templates/index.html.mako#L398

hatched:
:+1: :ok:

When setting `jujugui.GTM_enabled` in `development.ini` the cookie notification was shown the first page load and the GTM script was executed and additional scripts were downloaded.

jaycee:
:+1: thanks @anthonydillon

anthonydillon:
Thank you :shipit:

jujugui:

PR #1310, opened 2016-02-01
Bundle name collisions

### LINKED ISSUE #1303: Services with the same name clash when deploying a bundle
If you have wordpress on your canvas, then you add a bundle which includes wordpress then the two services clash and it puts the model into a weird state where they are both shown on the canvas but only one exists in the added services list.

huwshimi:
Handle adding a bundle with a service name that matches an existing service. Fixes #1303.

### DIFF
--- a/jujugui/static/gui/src/app/views/utils.js
+++ b/jujugui/static/gui/src/app/views/utils.js
@@ -1810,6 +1810,57 @@ YUI.add('juju-view-utils', function(Y) {
     return multiple ? utils.numToLetter(multiple) + char : char;
   };
 
+  /**
+    Generate a name for a service with an incremented number if needed.
+
+    @method generateServiceName
+    @param {String} charmName The charm name.
+    @param {String} charmId the full charm id.
+    @param {Object} services the list of services.
+    @param {Boolean} ignoreSelf ignore this service in the service count.
+    @returns {String} The name for the service.
+  */
+  utils.generateServiceName = function(
+    charmName, charmId, services, ignoreSelf) {
+    var charmCount = 0;
+    // Loop through each service and grab its charmid to see if it matches.
+    services.each((service) => {
+      // Remove the version number from the charm id as we don't care about the
+      // version when comparing charms.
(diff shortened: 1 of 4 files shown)


huwshimi:
QA:
- add wordpress
- add the wordpress bundle
- you should now have two wordpress services: "wordpress" and "wordpress-a" and there should be no errors in your browser console.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4314/
Test PASSed.

anthonydillon:
Code looks :ok: QA issue :+1: but spotted this issue (https://github.com/juju/juju-gui/issues/1311) but outside the scope of this branch.

jujugui/static/gui/src/app/views/utils.js
```diff
@@ -1810,6 +1810,57 @@ YUI.add('juju-view-utils', function(Y) {
...
+      var serviceCharmId = utils._extractCharmName(service.get('charm'));
+      var providedCharmId = utils._extractCharmName(charmId);
+      // If we have a matching charm then increase the count.
+      if (serviceCharmId === providedCharmId) {
+        charmCount += 1;
+      }
+    });
+    // In some cases the service has already been added to the db so we need to
+    // ignore it in the final count.
+    if (ignoreSelf) {
```
hatched: I don't really understand this - it's supposed to be generating the name to be added to the db...so how could it be in the db already?

hatched:
:+1: :ok: With one question that might be a 'fix' for something that shouldn't happen.

jujugui/static/gui/src/app/views/utils.js
huwshimi: It creates a reference in the db when it does `this.db.services.ghostService(charm);` above. See this line: https://github.com/juju/juju-gui/pull/1310/files#diff-1dcaa888078db4ae76c5fb071e2ead75R432

hatched:
:shipit:

jujugui:

PR #1317, opened 2016-02-03
Correctly increment service names without using the service count

### LINKED ISSUE #1309: Service name collision does not increment correctly
If you remove services you can end up in a situation where it tries to add multiple services with the same increment.

- add a service
- add the same service again
- you should have wordpress and wordpress-a
- remove wordpress
- add another wordpress
- you will get an error in the console and the service won't appear in the added services list as it has tried to add another wordpress-a

huwshimi:
The service names were using the service count to prevent collisions, however if services were removed and others there were duplicate names. This increments the highest number instead. Fixes #1309.

### DIFF
--- a/jujugui/static/gui/src/app/views/utils.js
+++ b/jujugui/static/gui/src/app/views/utils.js
@@ -1846,7 +1846,9 @@ YUI.add('juju-view-utils', function(Y) {
   */
   utils.generateServiceName = function(
     charmName, charmId, services, ignoreSelf) {
-    // Loop through each service and grab its charmid to see if it matches.
+    // Find the highest counter from the existing services. We can't just
+    // use the number of services as there could be duplicate names if services
+    // have been removed.
     var highestCounter = Math.max.apply(Math, services.map((service) => {
       // Remove the version number from the charm id as we don't care about the
       // version when comparing charms.
@@ -1854,7 +1856,9 @@ YUI.add('juju-view-utils', function(Y) {
       var providedCharmId = utils._extractCharmName(charmId);
       // If we have a matching charm then increase the count.
       if (serviceCharmId === providedCharmId) {
-        // Remove the service name and dash from the name to get the counter.
+        // Remove the service name and dash to get the counter. It is done this
+        // way so that we can get just the counter from the string and don't get
+        // false positives from names with dashes in them.
         var counter = service.get('name').replace(
           charmName, '').replace('-', '');
         if (counter === '') {
--- a/jujugui/static/gui/src/app/views/utils.js
+++ b/jujugui/static/gui/src/app/views/utils.js
@@ -1839,63 +1839,22 @@ YUI.add('juju-view-utils', function(Y) {
 
     @method generateServiceName
     @param {String} charmName The charm name.
-    @param {String} charmId the full charm id.
     @param {Object} services the list of services.
-    @param {Boolean} ignoreSelf ignore this service in the service count.
+    @param {Integer} counter The optional increment counter.
     @returns {String} The name for the service.
   */
-  utils.generateServiceName = function(
-    charmName, charmId, services, ignoreSelf) {
-    // Find the highest counter from the existing services. We can't just
-    // use the number of services as there could be duplicate names if services
-    // have been removed.
-    var highestCounter = Math.max.apply(Math, services.map((service) => {
-      // Remove the version number from the charm id as we don't care about the
-      // version when comparing charms.
-      var serviceCharmId = utils._extractCharmName(service.get('charm'));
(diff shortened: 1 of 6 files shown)


huwshimi:
QA:
- add wordpress
- add wordpress again
- you should have wordpress and wordpress-a
- remove wordpress
- add another wordpress
- you should now have wordpress-a and wordpress-b

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4328/
Test FAILed.

jujugui/static/gui/src/app/views/utils.js
```diff
@@ -1822,31 +1846,40 @@ YUI.add('juju-view-utils', function(Y) {
...
+    var highestCounter = Math.max.apply(Math, services.map((service) => {
       // Remove the version number from the charm id as we don't care about the
       // version when comparing charms.
       var serviceCharmId = utils._extractCharmName(service.get('charm'));
       var providedCharmId = utils._extractCharmName(charmId);
       // If we have a matching charm then increase the count.
       if (serviceCharmId === providedCharmId) {
-        charmCount += 1;
+        // Remove the service name and dash from the name to get the counter.
+        var counter = service.get('name').replace(
```
hatched: why not split on last - and grab the last block? I figure this would be less error prone.

jujugui/static/gui/src/test/test_ghost_deployer_extension.js
```diff
@@ -157,6 +157,18 @@ describe('Ghost Deployer Extension', function() {
     assert.equal(services.item(3).get('name'), 'mysql-a');
   });
 
+  it('increments the name with deleted services', function() {
```
hatched: This test should probably be expanded to make sure that it's not actually renaming existing services

jujugui/static/gui/src/app/views/utils.js
```diff
@@ -1822,31 +1846,40 @@ YUI.add('juju-view-utils', function(Y) {
   */
   utils.generateServiceName = function(
     charmName, charmId, services, ignoreSelf) {
-    var charmCount = 0;
     // Loop through each service and grab its charmid to see if it matches.
-    services.each((service) => {
+    var highestCounter = Math.max.apply(Math, services.map((service) => {
```
hatched: Can you expand the reasoning behind this approach in a comment. Something like... "We can't simply count the number of charms deployed because a user may remove a service in the middle of the charms and this causes a duplication of the 'next' id"

jujugui/static/gui/src/test/test_ghost_deployer_extension.js
```diff
@@ -157,6 +157,18 @@ describe('Ghost Deployer Extension', function() {
...
+    var charm = makeCharm();
+    ghostDeployer.db.services = new Y.juju.models.ServiceList();
+    var services = ghostDeployer.db.services;
+    services.ghostService(charm);
+    services.ghostService(charm);
+    services.item(0).destroy();
```
hatched: Could you also add a test where you remove a middle service and make sure it still increments the last one so they don't conflict.

hatched:
:+1: :ok: With some comments and tests
Thanks for this fix!

jujugui/static/gui/src/app/views/utils.js
huwshimi: This is because of two things. Firstly, if we just split on a hyphen we would get 'gui' from 'juju-gui' so by replacing the name we know we will only get the counter. Secondly, the first service name won't include the hypen. e.g. first service will be "wordpress" so if we were to remove "wordpress-" we'd be left with 'wordpress' which is not what we want. I'll extend the comment.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4337/
Test FAILed.

huwshimi:
This has been almost completely rewritten and needs brand new reviews. Thanks!

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4349/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4350/
Test PASSed.

hatched:
:+1: :ok: 
Thanks for finding the best way to tackle this problem.

anthonydillon:
Looks good :+1:

huwshimi:
Squirrel! :shipit:

jujugui:

jujugui:
        build url: http://ci.jujugui.org:8080/job/juju-gui-merge/1484

jujugui:

PR #1346, opened 2016-02-12
Open docs windows when exporting

anthonydillon:


anthonydillon:
## QA
- Click on export
- Check the download happens and a new tab to https://jujucharms.com/docs/stable/charms-bundles#local-deploy-via-command-line opens

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4425/
Test PASSed.

hatched:
This isn't going to be a very good user experience for anyone but absolute beginners. There should probably be a popup in the GUI instead which can be disabled/dismissed similar to the cookie warning.

anthonydillon:
Totally agree @Lucapaulina

Lucapaulina:
We'll be wrapping this into a new notification so it'll be disconnected.
This should only work for the demo not for standalone.

PR #1350, opened 2016-02-15
Use https for anonymous pulling of download-cache

mitechie:
I wanted to build the GUI Into a lxc container that was not setup with my ssh setup. I needed to update the download-cache to be https so that I could do an anonymous pull of the download-cache and a one off build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -271,7 +271,7 @@ watch:
 # Download cache
 ################
 $(CACHE):
-	git clone --depth=1 "git@github.com:juju/juju-gui-downloadcache.git" $(CACHE)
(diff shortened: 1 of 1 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4439/
Test PASSed.

hatched:
This can be considerably slower - Maybe this should be configurable instead, but still have git be the default?

mitechie:
Hmm, I know it's slower in theory but didn't seem slow on my end here. You
can leave it I guess as long as we document this or something.

hatched:
I suppose the slowness is caused by the number of requests which isn't really an issue in this case because of the small number of actual files being downloaded. It is definitely worth while, but customizable might be the best way :)

PR #1358, opened 2016-02-18
Implement Service.AddUnits API call.

frankban:
This is used to add units in a way compatible with Juju 2.0 and lower.
This needs to be QAed as usual on embedded 2.0 and charmed 1.25: add units, place them to specific machines, check everything works well and no console errors are printed.

### DIFF
--- a/jujugui/static/gui/src/app/store/env/go.js
+++ b/jujugui/static/gui/src/app/store/env/go.js
@@ -92,15 +92,15 @@ YUI.add('juju-env-go', function(Y) {
     structure, with Scope and Directive fields.
 
     @method parsePlacement
-    @param {String} machineTo The string placement.
+    @param {String} toMachine The string placement.
     @return {Object} A structure including Scope and Directive fields, or
       null if no placement is specified.
   */
-  var parsePlacement = function(machineTo) {
-    if (!machineTo) {
+  var parsePlacement = function(toMachine) {
+    if (!toMachine) {
       return null;
     }
-    var parts = machineTo.split(':');
+    var parts = toMachine.split(':');
     if (parts.length === 2) {
       return {Scope: parts[0], Directive: parts[1]};
     }
--- a/jujugui/static/gui/src/app/store/env/go.js
+++ b/jujugui/static/gui/src/app/store/env/go.js
@@ -84,6 +84,33 @@ YUI.add('juju-env-go', function(Y) {
     return newObj;
   };
 
+  // Define the special machine scope used to specify unit placements.
+  var MACHINE_SCOPE = '#';
+
+  /**
+    Attempt to parse the specified string and create a corresponding placement
+    structure, with Scope and Directive fields.
+
+    @method parsePlacement
+    @param {String} machineTo The string placement.
+    @return {Object} A structure including Scope and Directive fields, or
+      null if no placement is specified.
+  */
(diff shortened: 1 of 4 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4456/
Test PASSed.

jujugui/static/gui/src/app/store/env/go.js
```diff
@@ -84,6 +84,33 @@ YUI.add('juju-env-go', function(Y) {
...
 
+  // Define the special machine scope used to specify unit placements.
+  var MACHINE_SCOPE = '#';
+
+  /**
+    Attempt to parse the specified string and create a corresponding placement
+    structure, with Scope and Directive fields.
+
+    @method parsePlacement
+    @param {String} machineTo The string placement.
```
hatched: toMachine might make more sense here?

jujugui/static/gui/src/app/store/env/go.js
```diff
@@ -84,6 +84,33 @@ YUI.add('juju-env-go', function(Y) {
     return newObj;
   };
 
+  // Define the special machine scope used to specify unit placements.
+  var MACHINE_SCOPE = '#';
```
hatched: why is this out here if it never changes?

jujugui/static/gui/src/app/store/env/go.js
```diff
@@ -1488,50 +1515,51 @@ YUI.add('juju-env-go', function(Y) {
       @return {undefined} Sends a message to the server only.
     */
     _add_unit: function(service, numUnits, toMachine, callback) {
```
hatched: is the description for toMachine no longer valid based on the code in `parsePlacement`? Does the syntax of the placement need to change for the app code which is calling `add_unit` ?

jujugui/static/gui/src/app/store/env/go.js
frankban: Indeed, done.

jujugui/static/gui/src/app/store/env/go.js
frankban: It's a "constant".

jujugui/static/gui/src/app/store/env/go.js
frankban: My understanding is that passing 42 or 2/lxc/1 should just work: the parsed placement is like {Scope: '#', Directive: '42'} and that's it. I QAed with top level machines, we should try with containers during QA.

jujugui/static/gui/src/app/store/env/go.js
hatched: Will it ever get used anywhere else? 

I believe we can use:

`const MACHINE_SCOPE = '#'`

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4457/
Test PASSed.

hatched:
:+1: 

Even if I dont' totally understand the rational for the scope and directive changes :)

jujugui/static/gui/src/app/store/env/go.js
frankban: Absolutely, will do that in a follow up.

frankban:
:shipit:

jujugui:

PR #1359, opened 2016-02-18
Update SetCharm to support Juju 2.0

hatched:


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4459/
Test PASSed.

jujugui/static/gui/src/app/store/env/go.js
```diff
@@ -1396,30 +1396,47 @@ YUI.add('juju-env-go', function(Y) {
...
+        Type: 'Service',
+        Request: 'SetCharm',
+        Params: {
+          ServiceName: serviceName,
+          CharmUrl: charmUrl,
+          ForceUnits: forceUnits,
+          ForceSeries: forceSeries
+        }
+      };
+      if (this.findFacadeVersion('Service') === null) {
```
frankban: Are we sure this was enabled in all service versions?
I'd use the usual version check here (=== null || < 3).

jujugui/static/gui/src/app/store/env/go.js
```diff
@@ -1396,30 +1396,47 @@ YUI.add('juju-env-go', function(Y) {
...
+        // facade for setting the service charm.
+        rpc = {
+          Type: 'Client',
+          Request: 'ServiceSetCharm',
+          Params: {
+            ServiceName: serviceName,
+            CharmUrl: charmUrl,
+            // Because the call signature has changed a bit to properly set
+            // force on the old facade we will force if either of the forced
+            // values are truthy.
```
frankban: Sounds reasonable.

frankban:
:+1: with a single minor suggestion.
Also: it seems we don't have tests for SetCharm in test_env_go.js: could you please add at least a simple one for the 2.0 case?
Will QA asap, thank you!

frankban:
QA 2.0: the API calls look good but I encountered a weird behavior while downgrading or upgrading charms. Every time the charm is not found. When trying from the CLI I was able to downgrade, and at that point a was also able to re-upgrade to latest version from the GUI, and downgrade again to the version used in the CLI call, but not other versions. I suspect that we need to add the charm before setting it on a service. At least that's what the CLI does, and it actually make sense.

frankban:
QA not ok on 1.25: method Service(1).SetCharm is not implemented.
This is I think related to my previous comment about checking that the version is < 3, not just that a Service facade exists. At the 1.25 time, the service facade existed but did not implement that call.

frankban:
So please fix the check and add the test and then land this.
The AddCharm can be tackled in a follow up branch.
Thank you!

jujugui/static/gui/src/app/store/env/go.js
hatched: good catch, done

hatched:
Thanks for the thorough qa, changes have been made and updated. :shipit:

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4461/
Test FAILed.

jujugui:

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4462/
Test PASSed.

PR #1368, opened 2016-02-22
Profile rows

huwshimi:
Add styling and components for displaying models, bundles and charms on the profile page.


huwshimi:
This is using mocked data and will need to be updated when we can get the data into the GUI. I've also not added tests yet as I'm not sure how much will need to be modified.

We could hold off landing of until the data is available, but I wanted to get feedback on the approach at this point.

QA:
- visit `http://0.0.0.0:6543/profile/`
- check that the display looks somewhat like: https://github.com/CanonicalLtd/juju-design/issues/15#issuecomment-185731774

There will be followups to add the animation, flesh out the data displayed etc.

jujugui/static/gui/src/app/assets/css/_mixins.scss
```diff
@@ -141,28 +131,14 @@ button,
     }
 }
 
-.uncommitted-circle {
```
huwshimi: A few of these mixins are no longer required so I removed them as a drive-by.

jujugui/static/gui/src/app/components/user-profile/test-user-profile.js
```diff
@@ -27,7 +27,8 @@ describe('UserProfile', () => {
     YUI().use('user-profile', () => { done(); });
   });
 
-  it('renders the header and lists', () => {
+  // XXX Skip this test until the content of the profile has been locked down.
+  xit('renders the header and lists', () => {
```
huwshimi: This would be fixed up when I add the tests.

jujugui/static/gui/src/app/components/user-profile/user-profile.js
```diff
@@ -150,8 +151,267 @@ YUI.add('user-profile', function() {
...
+    /**
+      Generate the rows of bundle details.
+
+      @method _generateBundleRows
+      @returns {Array} The markup for the rows.
+    */
+    _generateBundleRows: function() {
+      var components = [];
+      // XXX mocked for now!
+      var bundles = [{
```
huwshimi: Dodgy mocks.

jujugui/static/gui/src/app/components/user-profile/user-profile.js
```diff
@@ -150,8 +151,267 @@ YUI.add('user-profile', function() {
...
+    /**
+      Generate the rows of charm details.
+
+      @method _generateCharmRows
+      @returns {Array} The markup for the rows.
+    */
+    _generateCharmRows: function() {
+      var components = [];
+      // XXX mocked for now!
+      var charms = [{
```
huwshimi: Dodgy mocks.

jujugui/static/gui/src/app/components/user-profile/user-profile.js
```diff
@@ -150,8 +151,267 @@ YUI.add('user-profile', function() {
...
+
+      @method _generateModelRows
+      @returns {Array} The markup for the rows.
+    */
+    _generateModelRows: function() {
+      var components = [];
+      this.state.envList.forEach((model) => {
+        var uuid = model.uuid;
+        components.push(
+          <juju.components.UserProfileEntity
```
huwshimi: Instead of using some kind of magic component with a whitelist of fields and some kind of definition for all the different labels and column widths etc. I ended up going with this wrapper component. It means a little more duplication, but it was a lot easier to build, and hopefully therefore to maintain.

kadams54:
Issues encountered in QA:

* When I scroll, I lose the close icon in the upper right. That should be fixed.
* Padding between a list and the header for the next list seems tight, both by itself and when compared to mock.
* I'm so confused... I thought these icons were supposed to be squircles? That the only place with round icons was going to be the canvas?

Everything else looks good, QA-wise.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4488/
Test PASSed.

Lucapaulina:
Hiya,

We've agreed to use circles everywhere from now on. No more squircles :)

anthonydillon:
QA issues:
- The background of the profile panel should be `#F7F7F7`
- The title "Charms" is not right aligned with the icons
- When you open a row the title "Description" should line up with "Credential" title

anthonydillon:
Apart from that code looks good.

kadams54:
This work has been merged over to https://github.com/juju/juju-gui/pull/1371 and will land there. The QA issues noted will be fixed in subsequent PRs (likely by either myself or @huwshimi).

PR #1371, opened 2016-02-22
Display my charms and bundles on user profile page.

kadams54:


jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4486/
Test FAILed.

kadams54:
## QA

I should note that this branch is still pretty rough; there's lots of comments in the code about work to be done later, we're missing data, the UI doesn't function and looks very different from the design, etc.

That said, the purpose of the branch is to get data to the page. We will refine that data and clean up design/UI issues in subsequent branches. What we don't want is for people to be blocked because the page has no real data.

So, with that disclaimer aside...

1. Load `/profile/` - you'll see calls in inspector being made to list both charms and bundles.
2. If, like me, you have no charms and bundles, you'll see some "No data found" messages. (Yeah, I know, terrible message, there's a comment about fixing that in the code, see disclaimer.)
3. If, like me, you're a slacker charm author, then update line 105 of `user-profile.js` so that the string 'lazypower' is being passed into `charmstore.list` instead of the `username` variable. Refresh, and behold! Charms!

Not currently supported: scrolling down to actually see the bundles (because I'm hoping @huwshimi's branch fixes this).

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4487/
Test PASSed.

jujugui/static/gui/src/app/app.js
```diff
@@ -728,8 +728,8 @@ YUI.add('juju-gui', function(Y) {
       @method _renderUserProfile
     */
     _renderUserProfile: function() {
-      var user = this.get('users')['charmstore'];
-      var username = user && user.user && user.user.name;
+      var auth = this.get('users')['charmstore'];
+      var username = auth && auth.user && auth.user.name;
```
hatched: aww c'mon spoiling my fun ;)

jujugui/static/gui/src/app/assets/javascripts/jujulib/juju.js
```diff
@@ -468,18 +468,22 @@ var module = module;
...
       @param author {String} The charm author's username.
-      @params callback {Function} A callback to handle errors or accept the data
+      @param callback {Function} A callback to handle errors or accept the data
           from the request. Must accept an error message or null as its first
           parameter and the response data as its second.
+      @param type {String} Type of entity to list, must be either 'bundle' or
+          'charm'. Defaults to charm.
     */
-    list: function(author, callback) {
+    list: function(author, callback, type) {
```
hatched: I feel like we should maintain a consistent public api call signature with the callback at the end. I see that `search` follows this pattern and should probably also be updated.

jujugui/static/gui/src/app/assets/javascripts/jujulib/juju.js
```diff
@@ -468,18 +468,22 @@ var module = module;
...
       author = encodeURIComponent(author);
+      type = encodeURIComponent(type || 'charm');
```
hatched: Instead of doing this here we can now rely on default function params

`list: function(author, type='charm', callback) {`

jujugui/static/gui/src/app/assets/javascripts/jujulib/test-jujulib-charmstore.js
```diff
@@ -317,19 +317,26 @@ describe('jujulib charmstore', function() {
...
-            'include=extra-info%2Fbzr-owner&' +
+            'include=extra-info&' +
             'include=supported-series&' +
             'include=stats']);
     });
 
+    it('can list bundles', function() {
+      charmstore.list('test-author', 'cb', {type: 'bundle'});
+      var qs = generatePath.lastCall.args[1];
+      assert.ok(qs.indexOf('type=bundle'), 'bundle not set in query string');
```
hatched: `ok` returns truthy for `-1` so you'll want to use equals and check that it's greater than -1

jujugui/static/gui/src/app/components/user-profile/user-profile.js
```diff
@@ -152,6 +245,9 @@ YUI.add('user-profile', function() {
 
     render: function() {
       var whitelist = ['path', 'name', 'user', 'uuid', 'host-ports'];
+      var charmCount = this.state.charmList && this.state.charmList.length;
```
hatched: To help with minifcation here... could you...
`var charmList = this.state.charmList`...

jujugui/static/gui/src/app/components/user-profile/user-profile.js
```diff
@@ -166,9 +262,9 @@ YUI.add('user-profile', function() {
             <div className="inner-wrapper">
               <juju.components.UserProfileHeader
                 avatar=""
-                bundleCount={0}
-                charmCount={0}
-                environmentCount={this.state.envList.length}
+                bundleCount={bundleCount || 0}
```
hatched: I'd prefer this logic to be in the definition - I don't feel too strongly though :)

jujugui/static/gui/src/app/app.js
kadams54: I hit a conflict here because we had pretty much *exactly* the same code. The only diff was that I went with `auth` specifically to avoid `user.user` ;-)

jujugui/static/gui/src/app/components/user-profile/user-profile.js
```diff
@@ -82,6 +92,89 @@ YUI.add('user-profile', function() {
...
+        return (
+          <span
+            key={tag}
+            className="user-profile__entityList-tags">
+            {tag}
+          </span>
+        );
+      });
+      var cs = this.props.charmstore;
+      var url = (cs && cs.url && cs.version) && cs.url + cs.version;
```
hatched: Is it even possible that there is no charmstore provided? This might be overly complicated for a case which isn't supported.

jujugui/static/gui/src/app/assets/javascripts/jujulib/juju.js
kadams54: Yeah, I tried that... and was very quickly reminded that `juju.js` is not Babelized, so it needs to be plain vanilla JS, rather than fancy ES5/6.

jujugui/static/gui/src/app/components/user-profile/user-profile.js
kadams54: Will do.

jujugui/static/gui/src/app/components/user-profile/user-profile.js
kadams54: Yeah, I agree. Originally the definition got pretty hairy on me so I moved the "or 0" bit down here, but if I make the "helps minification" change then I think the "or 0" can be moved back up.

jujugui/static/gui/src/app/assets/javascripts/jujulib/test-jujulib-charmstore.js
kadams54: Will do.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4489/
Test PASSed.

hatched:
Looks good! :+1: :ok:

Few qa notes for follow up:
- The profile page needs to be able to scroll
- Logging in requires you to close and re-open the profile page for it to show the charms
- When first loading the profile page it would be nice if there was a loading indicator instead of a message for no charms.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4493/
Test PASSed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4495/
Test FAILed.

kadams54:
## Updated QA

Alright, the UI is in place, so the real data looks prettier. There are still some rough spots, comments in the code about work to be done later, etc., but now we have a foundation for future user profile work.

As before...

1. Load `/profile/` - you'll see XHR calls in inspector being made to list both charms and bundles.
2. If, like me, you have no charms and bundles, you'll see nothing in each section. (We need better "nothing found" handling.)
3. If, like me, you're a slacker charm author, then update line 106 of `user-profile.js` so that the string 'lazypower' is being passed into `charmstore.list` instead of the `username` variable. Refresh, and behold! Charms and bundles!

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4496/
Test PASSed.

kadams54:
After talking over with @hatched we're going to handle fixes in subsequent PRs. :shipit:

jujugui:

PR #1389, opened 2016-02-24
Add supported series and unit counts to user profile.

kadams54:


kadams54:
## QA

If you don't have charms/bundles of your own, you may need to edit `user-profile.js` to pass a hardcoded username (like 'lazypower') into `charmstore.list()` instead of the `username` variable.

1. Pull up the user profile.
2. Verify that the unit count (for bundles) and series (for charms) are now displayed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4502/
Test FAILed.

hatched:
:+1: :ok:
Thanks for this update!

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4507/
Test PASSed.

jujugui/static/gui/src/app/assets/javascripts/jujulib/juju.js
```diff
@@ -347,6 +350,9 @@ var module = module;
         }, this);
       }
       if (processed.entityType === 'bundle') {
+        if (meta['bundle-unit-count']) {
```
hatched: I don't think that this is the correct way to protect against this, what will the component display in the event this isn't defined?

kadams54: Nothing is displayed. That causes problems with our layout, but I think that's more a function of CSS than JS (i.e., our table cells need to remain the same width even when they don't have content).

kadams54:
Thanks @hatched! :shipit:

jujugui:

PR #1392, opened 2016-02-25
Animate the entities on the profile page

huwshimi:
Animate the models, bundles and charms on the profile page when they are opened or closed.

### DIFF
--- a/jujugui/static/gui/src/app/components/user-profile/entity/entity.js
+++ b/jujugui/static/gui/src/app/components/user-profile/entity/entity.js
@@ -37,7 +37,11 @@ YUI.add('user-profile-entity', function() {
     */
     getInitialState: function() {
       return {
-        expanded: false
+        expanded: false,
+        styles: {
+          height: '0px',
+          opacity: 0
+        }
       };
     },
 
@@ -94,7 +98,16 @@ YUI.add('user-profile-entity', function() {
       @method _toggle
     */
     _toggle: function() {
-      this.setState({expanded: !this.state.expanded});
+      var expanded = this.state.expanded;
+      this.setState({expanded: !expanded}, () => {
+        var node = ReactDOM.findDOMNode(this);
+        var newHeight = expanded ? '0px' : node.querySelector(
+          '.user-profile__entity-details-inner').offsetHeight + 'px';
+        this.setState({styles: {
+          height: newHeight,
+          opacity: expanded ? 0 : 100
+        }});
+      });
     },
 
     /**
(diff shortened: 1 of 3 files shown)


huwshimi:
QA:
- view /profile/
- click on a model, bundle or charm
- notice that it animates open and closed

jujugui/static/gui/src/app/components/user-profile/entity/test-entity.js
```diff
@@ -321,7 +333,7 @@ describe('UserProfileEntity', () => {
     });
   });
 
-  it('can toggle to the expanded view', () => {
+  xit('can toggle to the expanded view', () => {
```
huwshimi: This test blows up when it hits `ReactDOM.findDOMNode(this)` in https://github.com/juju/juju-gui/pull/1392/files#diff-da75d897bfd779b4984b2f0681ca0cd4R103 with `TypeError: undefined is not an object (evaluating 'deepestAncestor.firstChild')`.

Any suggestions on why that might not work would be greatly appreciated.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4520/
Test PASSed.

jujugui/static/gui/src/app/components/user-profile/entity/entity.js
```diff
@@ -94,7 +98,16 @@ YUI.add('user-profile-entity', function() {
...
     _toggle: function() {
-      this.setState({expanded: !this.state.expanded});
+      var expanded = this.state.expanded;
+      this.setState({expanded: !expanded}, () => {
+        var node = ReactDOM.findDOMNode(this);
+        var newHeight = expanded ? '0px' : node.querySelector(
+          '.user-profile__entity-details-inner').offsetHeight + 'px';
+        this.setState({styles: {
+          height: newHeight,
+          opacity: expanded ? 0 : 100
```
anthonydillon: Opacities range is 0 - 1.

anthonydillon:
Code and QA :+1: with a small comment in code.

jujugui/static/gui/src/app/components/user-profile/entity/entity.js
huwshimi: Hah! Oops :)

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4525/
Test PASSed.

jujugui/static/gui/src/app/components/user-profile/entity/test-entity.js
huwshimi: It looks like we're having the same issue with the deployment bar notification tests: https://github.com/juju/juju-gui/blob/develop/jujugui/static/gui/src/app/components/deployment-bar-notification/test-deployment-bar-notification.js

hatched:
I find the animation quite jumpy. It would be nice if it animated like https://jsbin.com/sivajogasa/1/edit?html,css,js,output

jujugui/static/gui/src/app/components/user-profile/entity/test-entity.js
hatched: When this test is enabled we get the following error at the start of the suite: `ERROR: 'Warning: React can't find the root component node for data-reactid value `.6`. If you're seeing this message, it probably means that you've loaded two copies 
of React on the page. At this time, only a single copy of React can be loaded at a time.'`

hatched: Looking at the react source (and I suppose we should have assumed this by the method name) you can't find a DOM node if there is no DOM node created :) 

Are you able to use a ref in the setState instead of calling `findDOMNode` ?

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4550/
Test PASSed.

huwshimi:
@hatched I've updated the animation to not just hide the collapsed row.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4553/
Test PASSed.

hatched:
Thanks the animation is better now but when I open and close the whole profile shifts right to left.

huwshimi:
@hatched as discussed that jump is due to the scrollbar appearing as the extra content shifts the page down. :shipit:

jujugui:

PR #1393, opened 2016-02-25
Replace Install Juju button with sandbox/demo notification

anthonydillon:
A prominent notification at the bottom of the screen to inform your in sandbox mode.


anthonydillon:
## QA
- Open the GUI locally and you should see a dark bar along the bottom stating its a demo
- Check that the Install Juju button has been removed
- Check that the deployment bar is moved up when in sandbox to make room for the notification
- Click the Get started button and it should open the get-started page in a new tab
- Set the GUI to none sandbox or run it in a real environment and check the bar is not there
- Check the deployment bar is back to its original position

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4521/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4522/
Test PASSed.

jujugui/static/gui/src/app/components/demo-notification/demo-notification.js
```diff
@@ -0,0 +1,65 @@
...
+      sandboxMode: React.PropTypes.bool.isRequired
+    },
+
+    /**
+      Get the current state of the demo notification.
+
+      @method getInitialState
+      @returns {Object} The current state.
+    */
+    getInitialState: function() {
```
hatched: This function is not necessary you don't have any component state

jujugui/static/gui/src/app/components/demo-notification/demo-notification.js
```diff
@@ -0,0 +1,65 @@
...
+
+You should have received a copy of the GNU Affero General Public License along
+with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+'use strict';
+
+YUI.add('demo-notification', function() {
+
+  juju.components.DemoNotification = React.createClass({
```
hatched: This component looks like a good case for https://facebook.github.io/react/docs/reusable-components.html#stateless-functions

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4523/
Test PASSed.

jujugui/static/gui/src/app/components/demo-notification/test-demo-notification.js
```diff
@@ -0,0 +1,62 @@
...
+    // By loading this file it adds the component to the juju components.
+    YUI().use('demo-notification', function() { done(); });
+  });
+
+  beforeEach(function() {
+    var DemoNotification = juju.components.DemoNotification;
+    DemoNotification.prototype.sandboxMode = true;
+  });
+
+  xit('can render and in sandbox mode', function() {
```
huwshimi: This test is currently skipped...

huwshimi:
Thanks looks good @anthonydillon. The message appeared in sandbox mode and was hidden in a real env.

One QA issue: notifications appear under the message, I guess they'll need to be moved up too.

:+1: with the QA and test issue fixed.

anthonydillon:
Ah good spot with the notifications. I should have mentioned but not sure shallow renderer works with stateless functions so skipped it cos I didn't want the test file to be empty. Also there is no functionality or state so don't think the notification needs tests. What do you think? Can the test file be removed?

huwshimi:
@anthonydillon From memory you don't have to use the shallow renderer, you can just call the stateless function directly to get the output, see: https://github.com/juju/juju-gui/blob/develop/jujugui/static/gui/src/app/components/deployment-summary/test-deployment-summary-placement.js

As for whether or not it needs a test, I'm not sure. It might be nice to have a render test in case we add functionality in the future and then we don't have to do the test setup and we know if we've broken something. @hatched do you have any opinions on this?

hatched:
@huwshimi That sounds correct I believe @anthonydillon and I talked about that at EOD on Friday.

anthonydillon:
This is being closed as discussed in the issue to go no further with this.

PR #1417, opened 2016-03-02
Add and active Vanilla framework

anthonydillon:
Add Vanilla framework as the base CSS dependency. Included into the base style sheet with some updates to components to make them compatible.


anthonydillon:
## QA
- I'm afraid this is a difficult QA
- This branch should look the same as before (although there is some small movement in the header)
- You may want to load the gui from develop and keep that open in a new tab
- Load this branch in the next tab and flick between them
- Please compare all views and make sure there is nothing out of place

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4582/
Test PASSed.

hatched:
#### QA Notes:
- Size increase from 28498 to 31228, not a huge deal, but we should probably find out if we can remove that extra 2kb, as 30 is already a lot :)
- There appears to be too much padding on the bottom of the machines and the more menu
screenshot from 2016-03-02 13 00 51
- The top padding of the added services list looks like it's too big, or not even.
screenshot from 2016-03-02 13 03 09
- The service/machine view links do not have a pointer cursor
- The tables in the profile view appear to be slightly off in some columns
screenshot from 2016-03-02 13 08 09
- The tables don't have proper header spacing

screenshot from 2016-03-02 13 08 57

jujugui/static/gui/src/app/components/header-search/_header-search.scss
```diff
@@ -68,8 +68,12 @@
     &__input {
         width: 136px;
         background-color: transparent;
-        border: none;
+        border: none !important;
```
hatched: Heh why?

This feels like a bug in vanilla?

hatched:
You might also be able to fix this bug at the same time https://github.com/juju/juju-gui/issues/1419

jujugui/static/gui/src/app/components/header-search/_header-search.scss
anthonydillon: Yes this is a bug but also you caught me being lazy. I'll work out the important's.

anthonydillon:
Thanks Jeff, the file size will be reduced once the Vanilla theme had been created which will involve refactoring all the scss. Whaa, totally forgot to check MV. Will fix up.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4586/
Test PASSed.

hatched:
There appears to be a large top padding on the first element in this list:

screenshot from 2016-03-03 09 25 45

anthonydillon:
That gap is in develop so thought it was desired. @Lucapaulina @faooful

hatched:
Other than that :+1: :ok: Thanks for putting this together!

Lucapaulina:
That is the desired amount of padding.

huwshimi:
Nice work @anthonydillon :+1:! Might be worth an email at some point about how we should start using this stuff.

hatched:
:shipit:

jujugui:

PR #1427, opened 2016-03-03
Display the correct username in the breadcrumbs.

kadams54:
And also handle the plethora of users to properly request the right user profile data behind the scenes.


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4588/
Test PASSed.

kadams54:
## QA

In an incognito window:

1. Head to the user profile page.
2. In the breadcrumbs, you should see the username from the controller, 'user-admin'.
3. The only model listed should be sandbox, with no charms or bundles.
4. You should also see a "Login to charmstore" button. Even though the user is set to the controller, we still have no charmstore identity.
5. Click on the button and login to charmstore.
6. When the login process is finished, you should still see the controller username being displayed in the breadcrumbs and at the top of the user profile; however, you should see your bundles and charms listed. (I've raised the disconnect here between displayed username and seeing charms/bundles associated with anther username with @urosj and we'll address it appropriately with design on down the road.)
7. Now we need to make sure the breadcrumb behaves properly when there's no user. I've found the best way to do this is to either hack `_getAuth()` in `app.js` to always return `null` or set a breakpoint there and set it to null in the console every time the breakpoint is hit. Hacking is the easiest.
8. Once hacked so that the user is null, you should see the username drop altogether from the breadcrumb.

Makefile
```diff
@@ -71,7 +71,7 @@ help:
 	@echo "test - run python tests with the default Python"
 	@echo "test-deps - install the test dependencies"
 	@echo "test-js-karma - run newer js tests in terminal; primarily for CI build"
-	@echo "test-mocha-karma - run older js tests that have transitioned to karma in the terminal"
+	@echo "test-js-old - run older js tests that have transitioned to karma in the terminal"
```
hatched: We didn't choose this because the code is still modified and updated regularly so it's not 'old'. If you'd like the suites to be more consistent then `test-js-karma` should be renamed to `test-jasmine-karma`

jujugui/static/gui/src/app/app.js
```diff
@@ -2116,15 +2114,30 @@ YUI.add('juju-gui', function(Y) {
...
       var externalAuth = this.get('auth');
       if (externalAuth) {
         return externalAuth;
       }
       var users = this.get('users');
-      return users && users[service];
+      var user;
+      if (users) {
+        var controllerUser;
+        if (this.env) {
```
hatched: why would this.env ever not exist at this point? Can you add a code comment to that effect.

jujugui/static/gui/src/app/components/header-breadcrumb/header-breadcrumb.js
```diff
@@ -65,15 +65,22 @@ YUI.add('header-breadcrumb', function() {
...
+        var username = auth.loading ? '...' : auth.user;
+        userItem = (
           <li className="header-breadcrumb__list-item">
             <a className="header-breadcrumb--link" href="/profile/">
               {username}
             </a>
           </li>
+        );
+      } else {
+        userItem = undefined;
```
hatched: `userItem` already is undefined (line 68)

jujugui/static/gui/src/app/components/user-profile/user-profile.js
```diff
@@ -146,21 +143,21 @@ YUI.add('user-profile', function() {
       @param {Object} data The data from the request.
     */
     _fetchEntitiesCallback: function(type, error, data) {
+      // Turn off the loader, regardless of success or error.
+      if (type === 'charm') {
```
hatched: Why did you split this out of the conditional below? I don't really see any advantage to it being up here with the exception of if there is an error it'll still turn off the loading indicators. If so could you update the code comment plz

jujugui/static/gui/src/app/app.js
kadams54: Because people are lazy with mocking things up in tests. Not sure that really merits a code comment.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4592/
Test PASSed.

jujugui/static/gui/src/app/app.js
```diff
@@ -2110,15 +2108,30 @@ YUI.add('juju-gui', function(Y) {
...
-    _getAuth: function(service) {
+    _getAuth: function() {
       var externalAuth = this.get('auth');
       if (externalAuth) {
         return externalAuth;
       }
       var users = this.get('users');
-      return users && users[service];
+      var user;
+      if (users) {
```
huwshimi: Can you have a users and not a users.jem or users.charmstore? Just wonder when you'd end up with a controllerUser?

jujugui/static/gui/src/app/components/user-profile/user-profile.js
```diff
@@ -58,16 +52,18 @@ YUI.add('user-profile', function() {
...
     },
 
     componentDidUpdate: function(prevProps, prevState) {
-      // If the user has just been authenticated then update the data.
-      if (!prevProps.authenticated && this.props.authenticated) {
+      // If the user has changed then update the data.
+      if (prevProps.user !== this.props.user) {
         this._fetchEnvironments();
+      }
+      if (prevProps.users.charmstore !== this.props.users.charmstore) {
```
huwshimi: Do we need to check for `prevProps.users` and `this.props.users` here? If they were not previously authenticated might they not have a .charmstore property?

jujugui/static/gui/src/app/app.js
huwshimi: I think it would be nice to try and remove this condition and fix the tests, it's probably the only way the tests are going to get any better...

jujugui/static/gui/src/app/components/user-profile/user-profile.js
kadams54: Yes, that is the reason why. If there is an error, we still want the load indicator to go away. Which is why I put a comment to that effect on line 146 ;-)

Makefile
kadams54: I'd like it if both targets started with "test-js-" - the reason I changed this is that I can never remember which testing framework is used for which set of tests, so I kept having to run `make help` every time. I think "old" is still fair because we want to move tests out of there, even if we have to maintain them short term.

jujugui/static/gui/src/app/components/header-breadcrumb/header-breadcrumb.js
```diff
@@ -65,15 +65,22 @@ YUI.add('header-breadcrumb', function() {
 
     render: function() {
       var auth = this.props.authDetails;
-      var loading = auth && auth.loading;
-      var username = loading ? '...' : auth && auth.user || 'anonymous';
-      return (
-        <ul className="header-breadcrumb">
+      var userItem;
+      if (auth && (auth.user || auth.loading)) {
```
huwshimi: I might be missing something here but it seems strange that you sometimes might have a different breadcrumb structure here. It might be more helpful to have the anonymous user there so you can go back up to your profile and log in to the charmstore and see your bundles etc.

Not sure if there's a conversation I missed out on that makes sense of this...

jujugui/static/gui/src/app/components/user-profile/user-profile.js
kadams54: The users prop is always set, if only to an empty object.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4593/
Test PASSed.

jujugui/static/gui/src/app/app.js
kadams54: I think this section of code may also be triggered before an env connection is established. I'll verify if that's the case. If it's just tests, then I'll see how painful it will be to fix, but these are the old tests so... mocking. Ugh.

jujugui/static/gui/src/app/components/user-profile/user-profile.js
```diff
@@ -35,13 +34,8 @@ YUI.add('user-profile', function() {
...
       switchModel: React.PropTypes.func.isRequired,
-      username: React.PropTypes.string
-    },
-
-    getDefaultProps: function() {
-      return {
-        username: ''
-      };
+      users: React.PropTypes.object.isRequired,
+      user: React.PropTypes.string
```
huwshimi: should be an object?

jujugui/static/gui/src/app/components/header-breadcrumb/header-breadcrumb.js
kadams54: Yeah, this is the official solution for the time being - drop the username from the breadcrumbs if we're not logged in.

jujugui/static/gui/src/app/components/user-profile/user-profile.js
kadams54: Good catch.

jujugui/static/gui/src/app/app.js
kadams54: Any time you're connected to the GUI but haven't yet logged into the charmstore. Which is most of the time for local dev.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4598/
Test PASSed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4602/
Test FAILed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4603/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4604/
Test PASSed.

hatched:
QA Issue: In a Juju 2.0 env in the profile view there is no log into charmstore button

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4605/
Test PASSed.

kadams54:
After talking with @hatched, we've decided to land this branch and address any lingering issues in follow-up PRs. :shipit:

jujugui:

PR #1471, opened 2016-03-11
Don't override the model name unless GUI is embedded.

kadams54:
We only need to override the model name with the UUID string when the GUI is running in embedded mode.

### DIFF
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -1086,6 +1086,10 @@ YUI.add('juju-gui', function(Y) {
         // as it's visible but not functional.
         showEnvSwitcher = false;
       }
+      // If we're in sandbox we don't want to display the switcher.
+      if (this.get('sandbox')) {
+        showEnvSwitcher = false;
+      }
       var uncommittedChanges = false;
       if (env) {
         var currentChangeSet = env.get('ecs').getCurrentChangeSet();
--- a/jujugui/static/gui/src/app/models/models.js
+++ b/jujugui/static/gui/src/app/models/models.js
@@ -925,62 +925,62 @@ YUI.add('juju-models', function(Y) {
   // human maintenance, so verify your assumptions before proceeding from
   // reading this code.
   var ServiceUnit = Y.Base.create('serviceUnit', Y.Model, [], {},
-      {
-        ATTRS: {
-          displayName: {},
-          charmUrl: {},
-          machine: {},
-          agent_state: {},
-          agent_state_info: {},
-          agent_state_data: {},
-          // This is empty if there are no relation errors, and otherwise
-          // shows only the relations with errors.  The data structure in that
-          // case is a hash mapping a local relation name to a list of services
-          // on the other end, like {'cache': ['memcached']}.
-          relation_errors: {},
-          /**
-            If the unit has been marked for deletion via the ECS.
+    {
+      ATTRS: {
+        displayName: {},
+        charmUrl: {},
+        machine: {},
+        agent_state: {},
+        agent_state_info: {},
+        agent_state_data: {},
+        // This is empty if there are no relation errors, and otherwise
+        // shows only the relations with errors.  The data structure in that
+        // case is a hash mapping a local relation name to a list of services
+        // on the other end, like {'cache': ['memcached']}.
+        relation_errors: {},
+        /**
+          If the unit has been marked for deletion via the ECS.
 
-            @attribute deleted
-            @type {Boolean}
-          */
-          deleted: {},
-          /**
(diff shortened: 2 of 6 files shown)


jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4666/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4667/
Test PASSed.

development.ini
```diff
@@ -15,6 +15,7 @@ jujugui.combine = true
 jujugui.gzip = true
 jujugui.cachebuster = dev
 jujugui.socketTemplate = /environment/$uuid/api
+jujugui.embedded = true
```
jaycee: Let's not call this embedded, b/c it's not necessarily true for all embedded contexts.

jujugui/static/gui/src/app/app.js
```diff
@@ -1905,10 +1905,10 @@ YUI.add('juju-gui', function(Y) {
       if (localStorage.getItem('environmentName')) {
         environmentName = localStorage.getItem('environmentName');
       }
-      // The config.js jujuEnvUUID trumps 'sandbox'; in any other
+      // The config.js jujuEnvUUID trumps when the GUI is embedded; in any other
       // situation, the provided environmentName should update the app's
       // jujuEnvUUID.
-      if (environmentName === 'sandbox') {
+      if (this.get('embedded')) {
```
jaycee: This should be `this.get('embedded`) && environmentName === 'sandbox'`; we get environmentName up above and if another name has been assigned we don't want to trump that.

jujugui/views.py
```diff
@@ -109,6 +109,8 @@ def config(request):
         # Enable/disable sandbox (demonstration) mode.
         'sandbox': sandbox_enabled,
         'sandboxSocketURL': 'wss://demo.jujucharms.com/ws',
+        # Enable/disable embedded gui
+        'embedded': settings.get('jujugui.embedded', False),
```
jaycee: Rather than doing that here, can you update options.py to set the correct defaults? That handles things like interpreting the ini providing the string "true" and makes sure we're only doing defaults in one place.

hatched:
So when the GUI is in sandbox mode it shouldn't be showing the model switcher at all. I think you should look into why this is being shown at all - this may even be a miss-configured mojo script.

hatched:
Here is a good place to check if we are in sandbox mode and shouldn't show the model switcher: https://github.com/juju/juju-gui/blob/develop/jujugui/static/gui/src/app/components/header-breadcrumb/header-breadcrumb.js#L49

However it's possible that in sandbox mode we may not want to show the username either, at which point this is a better place to parse that logic: https://github.com/juju/juju-gui/blob/develop/jujugui/static/gui/src/app/app.js#L1075

It looks like the 'if sandbox' logic was missed when creating the breadcrumb component.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4669/
Test FAILed.

jujugui/static/gui/src/app/app.js
```diff
@@ -1086,6 +1086,10 @@ YUI.add('juju-gui', function(Y) {
         // as it's visible but not functional.
         showEnvSwitcher = false;
       }
+      // If we're in sandbox we don't want to display the switcher.
+      if (this.get('sandbox')) {
+        showEnvSwitcher = false;
+      }
```
jaycee: Not that I think it makes a huge difference, but is there any reason this isn't merged with the if condition above? This is just one more condition for which the env switcher should not be shown.

jaycee:
@kadams54 Can we add a test showing that the setting is correctly toggled false when the gui is in sandbox? Otherwise this SGTM.

kadams54:
## QA

Run the GUI in local dev mode; load the front page. There should be no env switcher. Click around to make sure the env switcher doesn't load up on other pages as well.

jujugui/static/gui/src/app/app.js
kadams54: Readability and cleanliness. They're two different situations, so I wanted to keep them separate, with separate comments, which would make it easier should we ever change our minds on one or the other.

jaycee:
Running in sandbox I see `juju / admin` in the breadcrumbs, with no env switcher. That seems correct to me.

QA :OK:
:+1: 

Thanks @kadams54.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4672/
Test PASSed.

kadams54:
Thanks all! :shipit:

jujugui:

PR #1472, opened 2016-03-15
Add panel component for deployment flow

huwshimi:
In this branch:
- moved the deployment bar out of the deployment component so that it can be rendered independently of the deployment flow
- added a new panel for displaying the deployment steps
- added the summary as a deployment step
- updated the state for handling the new deployment flow
- updated the existing deployment summary to work with the changes to the state/rendering of the deployment components. The old summary will be retained while we are building the new deployment flow. Some of the sub-components are duplicated for the summary so that they can be modified without affecting the other flow, hence the size of this branch.

The deployment flow will work by adding another component following the structure of summary.js that will be rendered by deployment.js using another case here: https://github.com/juju/juju-gui/pull/1472/files#diff-760cf384f3bb1b0f33ff2b63af7534f9R43

The case is parsed from the deploy url param so for `.../deploy/setup/` "setup" would be the case and whatever is returned will be displayed in the deployment panel.

Those panels can then be navigated to from any control in the flow with a standard `changeState` call such as:

```
    changeState({
        sectionC: {
          component: 'deploy',
          metadata: {
            activeComponent: 'setup'
          }
        }
      });
```

The step can be displayed in the head by adding another step here: https://github.com/juju/juju-gui/pull/1472/files#diff-760cf384f3bb1b0f33ff2b63af7534f9R59


huwshimi:
QA:
- add a service or machine
- click deploy
- you should get the new full-screen deployment panel

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4674/
Test PASSed.

jujugui/static/gui/src/app/components/deployment/deployment.js
```diff
@@ -36,68 +36,38 @@ YUI.add('deployment-component', function() {
...
-              importBundleFile={this.props.importBundleFile}
-              hasEntities={this.props.hasEntities}
-              hideDragOverNotification={this.props.hideDragOverNotification}
-              generateChangeDescription={generateChangeDescription}
-              currentChangeSet={currentChangeSet}
-              showInstall={this.props.showInstall} />
-          };
-          break;
+    _generateActivePanel: function() {
+      switch (this.state.activeComponent) {
```
hatched: `getInitialState` doesn't have `activeComponent` defined

jujugui/static/gui/src/app/components/deployment/deployment.js
```diff
@@ -217,15 +185,36 @@ YUI.add('deployment-component', function() {
...
+      var activeComponent = this.state.activeComponent;
+      var activeChild = this._generateActivePanel();
+      var steps = [{
+        title: 'Deploy',
+        active: activeComponent === 'deployment-summary'
+      }];
       return (
         <div className="deployment-view">
-          {this.state.activeChild.component}
+          <juju.components.DeploymentBar
```
hatched: I'm not sure it makes sense for this to be here anymore. Is there a reason why it can't be moved into its own separate component, especially now that it's being completely covered? 

The new designs also have the import/export going into the header so the logic around it is going to get simpler.

jujugui/static/gui/src/app/components/deployment/panel/panel.js
```diff
@@ -0,0 +1,118 @@
...
+              <juju.components.ButtonRow
+                buttons={buttons} />
+            </div>
+          </div>
+        </div>);
+    },
+
+    render: function() {
+      return (
+        <juju.components.Panel
```
hatched: Now that this is gone full screen wrapping it in a panel doesn't really appear to get us much beyond another layer of abstraction. Is this really necessary?

hatched:
Lets chat about the implementation here - I think we can probably break away from the old implementation entirely, possibly even fork it with a feature flag.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4691/
Test FAILed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4692/
Test FAILed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4703/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4704/
Test PASSed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4705/
Test FAILed.

jujugui/static/gui/src/app/components/deployment/summary/change-item/_change-item.scss
```diff
@@ -0,0 +1,28 @@
+.deployment-summary-change-item {
```
huwshimi: Note for reviewer: this component is a duplicate of the existing change-item scss.

jujugui/static/gui/src/app/components/deployment/summary/change-item/change-item.js
```diff
@@ -0,0 +1,64 @@
+/*
```
huwshimi: Note for reviewer: this component is a duplicate of the existing change-item component.

jujugui/static/gui/src/app/components/deployment/summary/change-item/test-change-item.js
```diff
@@ -0,0 +1,77 @@
+/*
```
huwshimi: Note for reviewer: this component is a duplicate of the existing change-item tests.

jujugui/static/gui/src/app/components/deployment/summary/placement/placement.js
```diff
@@ -0,0 +1,68 @@
+/*
```
huwshimi: Note for reviewer: this component is a duplicate of the existing placement component.

jujugui/static/gui/src/app/components/deployment/summary/placement/test-placement.js
```diff
@@ -0,0 +1,91 @@
+/*
```
huwshimi: Note for reviewer: this component is a duplicate of the existing placement tests.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4706/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4707/
Test PASSed.

huwshimi:
This is now ready for review.

jujugui/static/gui/src/app/components/deployment-bar/deployment-bar.js
```diff
@@ -39,15 +45,48 @@ YUI.add('deployment-bar', function() {
...
     /**
+      Check if we have any commits.
+
+      @param {Function} callback A function to call once the state has updated.
+      @method _updateHasCommits
+    */
+    _updateHasCommits: function() {
+      var hasCommits = false;
+      if (!this.state.hasCommits) {
+        this.props.services.forEach(service => {
```
hatched: Because this is called from `componentWillReceiveProps` shouldn't it also take a props argument and use that one instead? Because here you're using `currentProps` instead of `nextProps` in the `componentWillReceiveProps` case.

jujugui/static/gui/src/app/components/deployment-bar/deployment-bar.js
```diff
@@ -22,10 +22,16 @@ YUI.add('deployment-bar', function() {
 
   juju.components.DeploymentBar = React.createClass({
     propTypes: {
+      changeState: React.PropTypes.func.isRequired,
+      currentChangeSet: React.PropTypes.object.isRequired,
```
hatched: I'm wondering if you shouldn't instead pass a 'getCurrentChangeSet' method and then everywhere you'd use this, you'd fetch the existing changeset. This would have the side effect of also being useful in the `_updateHasCommits` which would simply become a length check on the return value of this method.

jujugui/static/gui/src/app/components/deployment-summary/deployment-summary.js
```diff
@@ -18,12 +18,90 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
...
 'use strict';
 
-YUI.add('deployment-summary', function() {
+YUI.add('deployment-summary-classic', function() {
 
-  juju.components.DeploymentSummary = React.createClass({
+  juju.components.DeploymentSummaryClassic = React.createClass({
 
-    propTypes: {
-      summaryClearAction: React.PropTypes.func.isRequired
```
hatched: I'm assuming that these were left out because we'll be removing this component in the long term anyways so it wasn't worth the effort?

jujugui/static/gui/src/app/components/deployment/panel/panel.js
hatched: As discussed it adds a few things which are shared elsewhere so there is enough that it doesn't make sense to duplicate the logic now.

jujugui/static/gui/src/app/components/deployment/summary/summary.js
```diff
@@ -0,0 +1,193 @@
...
+      }];
+      return (
+        <div className="deployment-panel__child">
+          <div className="deployment-panel__content">
+            <div className="twelve-col">
+              <div className="inner-wrapper">
+                <h2 className="deployment-panel__title">
+                  Deployment summary
+                </h2>
+                <juju.components.DeploymentSummaryPlacement
```
hatched: This will need to be flagged - if we're in sandbox mode then we autoplace by default in the new deploy flow, but in non-sandbox we give the option.

jujugui/static/gui/src/app/components/deployment/summary/summary.js
```diff
@@ -0,0 +1,193 @@
...
+                    <span className="deployment-summary-change-item__time">
+                      Time
+                    </span>
+                  </li>
+                  {this._generateChangeItems()}
+                </ul>
+              </div>
+            </div>
+          </div>
+          <div className="deployment-panel__footer">
```
hatched: So this is the pattern we'll need to follow on each one of these steps? Does it make sense to split this out into its own component for `DeploymentPanelFooter`?

jujugui/static/gui/src/app/components/header-search/header-search.js
```diff
@@ -47,7 +47,7 @@ YUI.add('header-search', function() {
     _activeForComponent: function() {
       var component = this.props.getAppState(
         'current', 'sectionC', 'component');
-      return !!component && component !== 'profile';
+      return !!component && component !== 'profile' && component !== 'deploy';
```
hatched: Maybe this should be converted to be 'truthy' as in "we only want to show it for these components...."

jujugui/static/gui/src/test/test_ui_state.js
```diff
@@ -846,8 +864,26 @@ describe('UI State object', function() {
...
+        sectionC: {
+          component: 'deploy',
+          metadata: {
+            activeComponent: 'summary'
+          }
+        },
+      },
       // Multi section urls.
-      '/inspector/apache2/machine/3/lxc-0': {
+      '/inspector/apache2/machine/3/lxc-0/deploy/foo': {
```
hatched: We should have a test for the logic around a url with charmbrowser and deploy because they occupy the same space in the gui but could be supplied in a manually written url.

hatched:
:+1: :ok:

Thanks for sticking through with this massive change - it's a much needed improvement, good work! I have a number of comments but we are in need of this branch to land so that others can continue to work on parts that it adds, so if this lands before you have a chance to do updates they can be done in follow-ups.

jujugui/static/gui/src/app/components/deployment/summary/summary.js
kadams54: Do we want to set it up so that in sandbox we can still force it to give the option, for testing purposes?

hatched: Sorry I mean, sandbox will have the autoplace but `sax` feature flag won't have autoplace.

kadams54:
:+1: from me as well.

jujugui/static/gui/src/app/components/deployment-summary/deployment-summary.js
huwshimi: Yeah, we're not using `summaryClearAction` anymore and as you say this component will go away so I didn't add the missing props.

jujugui/static/gui/src/app/components/deployment/summary/summary.js
huwshimi: Yeah, we could do that, we also have to follow the pattern for the content above the footer:

```
        <div className="deployment-panel__child">
          <div className="deployment-panel__content">
            <div className="twelve-col">
              <div className="inner-wrapper">
```

Maybe we should do a stateless component for that as well that just wraps the children.

jujugui/static/gui/src/app/components/deployment-bar/deployment-bar.js
huwshimi: Good catch.

jujugui/static/gui/src/app/components/deployment-bar/deployment-bar.js
huwshimi: As discussed, the value of that method is passed in and is updated with the calls here: https://github.com/juju/juju-gui/pull/1472/files#diff-21dd467cb411bdd922556d1af7461864R429

I've renamed the hasCommits method and variable to be more clear about its purpose.

jujugui/static/gui/src/test/test_ui_state.js
huwshimi: Added.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4713/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4714/
Test PASSed.

huwshimi:
Thanks @hatched and @kadams54 :shipit:

jujugui:

PR #1488, opened 2016-03-23
Add listTemplates API call to juju.js.

kadams54:


jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4731/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4732/
Test PASSed.

jujugui/static/gui/src/app/assets/javascripts/jujulib/juju.js
```diff
@@ -205,6 +205,19 @@ var module = module;
         // if the macaroon is valid.
         false
       );
+    },
+
+    /**
+      Lists all the templates for cloud credentials for a specific user.
```
frankban: What user? We don;t pass a user here, so I guess it is "for the currently authenticated user"?

bac:
:+1:

jaycee:
Code is :+1:

QA to follow.

jujugui/static/gui/src/app/assets/javascripts/jujulib/juju.js
```diff
@@ -205,6 +205,19 @@ var module = module;
...
+
+      @method listTemplates
+      @param callback {Function} A callback to handle errors or accept the data
+          from the request. Must accept an error message or null as its first
+          parameter and the response data as its second.
+    */
+    listTemplates: function(callback) {
+      var url = this.jemURL + '/template';
+      return _makeRequest(this.bakery, url, 'GET', null, callback, true);
```
frankban: It would be nice to not just propagate the data sent over the wire.
Transforming that data to some internal data structure ensures the client results are JS friendly and sane, we can easily switch API versions without potentially change internal code and also documents what's passed to callbacks, which is also something we need to add here, as reading this function does not inform me about what's data is actually passed to callbacks.

jujugui/static/gui/src/app/assets/javascripts/jujulib/test-jujulib-environment.js
```diff
@@ -206,4 +206,27 @@ describe('jujulib environment manager', function() {
       }
     );
   });
+
+  it('lists the current user\'s templates', function(done) {
```
frankban: Nice test!
Can we also exercise the error case?

jujugui/static/gui/src/app/assets/javascripts/jujulib/test-jujulib-environment.js
```diff
@@ -206,4 +206,27 @@ describe('jujulib environment manager', function() {
...
+        assert.equal(path, 'http://example.com/v1/template');
+        var xhr = _makeXHRRequest(templates);
+        success(xhr);
+      },
+    };
+
+    env = new window.jujulib.environment('http://example.com/', 'v1', bakery);
+    env.listTemplates(
+      function(error, data) {
+        if (error) {
```
frankban: Or just `assert.strictEqual(error, null)`?

frankban:
This branch looks good, with some suggestions.
Thank you!

jujugui/static/gui/src/app/assets/javascripts/jujulib/juju.js
jaycee: I don't think we want to massage too much, if only b/c then we have different expectations for what each method does; however, like the listServers method this should probably return the data inside the templates key, rather than the whole response. This will also involve a test update.

frankban: To be clear, I am suggesting the translation to be done by endpoints, not at the _makeRequest level, so different expectations is not an issue. On the other hand, implicit expectations can surely become a problem, sooner or later.

jujugui/static/gui/src/app/assets/javascripts/jujulib/juju.js
kadams54: Yup, good point. I'll clarify.

jujugui/static/gui/src/app/assets/javascripts/jujulib/test-jujulib-environment.js
kadams54: Yeah, I'll add an error case test.

jujugui/static/gui/src/app/assets/javascripts/jujulib/juju.js
kadams54: Yeah, most other API functions have some sort of translation that happens. This was a case where I put in the translation layer and then kept simplifying until it was eventually non-existent. That said, I think I went too far and missed the data.templates bit that @jaycee pointed out. I'll add a minimal translation layer back in.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4735/
Test PASSed.

jujugui/static/gui/src/app/assets/javascripts/jujulib/juju.js
```diff
@@ -205,6 +205,26 @@ var module = module;
...
+      @method listTemplates
+      @param callback {Function} A callback to handle errors or accept the data
+          from the request. Must accept an error message or null as its first
+          parameter and the response data as its second.
+    */
+    listTemplates: function(callback) {
+      var url = this.jemURL + '/template';
+      var _listTemplates = function(error, data) {
+        if (error === null) {
+          data = data.templates;
```
hatched: Is the rest of the data provided not valuable? I understand this is a listTemplates method, I'm just wondering what information we're not showing the user, maybe the full data object could be passed as a third argument to the cb?

kadams54: The only data we're missing out on is the error, and only if it's null.

hatched: ahh ok so it just returns the nested structure. Gotcha!

hatched:
:+1:

jaycee:
QA notes:

I began with no templates on the staging JEM. Calling listTemplates from the web console returned null for the data, which seems to be correct looking at the jem code.

I then added a template; listTemplates then returned an array with one element, the template I added.

QA OK.

jaycee:
:+1:

kadams54:
Thanks all! :shipit:

jujugui:

PR #1493, opened 2016-03-24
Skip connecting to a model on load

hatched:
Depending on how the GUI is being run we may not want to connect to a model on load. This branch prevents that until you provide a socket url to the gui's env and then manually connect it using `app.env.connect()`

### DIFF
--- a/development.ini
+++ b/development.ini
@@ -15,7 +15,7 @@ jujugui.combine = true
 jujugui.gzip = true
 jujugui.cachebuster = dev
 jujugui.socketTemplate = /environment/$uuid/api
-
+jujugui.gisf = false
 
 # API config
 
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -561,7 +561,11 @@ YUI.add('juju-gui', function(Y) {
       // arrive with successful authentication), and redispatch.
       this.env.after('connectedChange', function(ev) {
         if (ev.newVal === true) {
-          this.db.reset();
+          // If we're in gisf we do not want to empty the db when we connect
+          // because the user may have made changes to the temporary model.
+          if (!this.get('gisf')) {
+            this.db.reset();
+          }
           this.env.userIsAuthenticated = false;
           // Do not attempt environment login without credentials.
           var credentials = this.env.getCredentials();
@@ -641,8 +645,10 @@ YUI.add('juju-gui', function(Y) {
 
       // We are now ready to connect the environment and bootstrap the app.
       this.once('ready', function(e) {
-        if (this.get('socket_url') || this.get('sandbox')) {
-          // Connect to the environment.
+        // We only want to connect to the model on application load if we are
+        // in a sandbox or real model and not in gisf.
+        if (!this.get('gisf') &&
+            (this.get('socket_url') || this.get('sandbox'))) {
           this.env.connect();
         }
         this.dispatch();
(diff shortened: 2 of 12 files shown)


jujugui/static/gui/src/app/app.js
```diff
@@ -1614,6 +1620,15 @@ YUI.add('juju-gui', function(Y) {
      *
      */
     checkUserCredentials: function(req, res, next) {
```
makyo: Called as middleware when YUI dispatches routes

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4738/
Test FAILed.

jujugui/static/gui/src/app/app.js
hatched: This means that every time the route changes (which is every time state changes) this method is called. The methods in the `gisf` block below are not idempotent but the values which they are setting should not change between routes. There is some micro performance improvements to be had here if one wanted to set up some sort of flag.

jujugui/static/gui/src/app/app.js
```diff
@@ -641,8 +645,10 @@ YUI.add('juju-gui', function(Y) {
 
       // We are now ready to connect the environment and bootstrap the app.
       this.once('ready', function(e) {
-        if (this.get('socket_url') || this.get('sandbox')) {
-          // Connect to the environment.
+        // We only want to connect to the model on application load if we are
+        // in a sandbox or real model and not in gisf.
+        if (!this.get('gisf') &&
```
hatched: After setting the socket url from the new model in `app.env` you simply need to call `app.env.connect` for it to connect and it'll start to process the incoming deltas as normal.

jujugui/static/gui/src/app/app.js
```diff
@@ -561,7 +561,11 @@ YUI.add('juju-gui', function(Y) {
       // arrive with successful authentication), and redispatch.
       this.env.after('connectedChange', function(ev) {
         if (ev.newVal === true) {
-          this.db.reset();
+          // If we're in gisf we do not want to empty the db when we connect
+          // because the user may have made changes to the temporary model.
+          if (!this.get('gisf')) {
```
hatched: Typically, when switching models we want to reset the db so that we are starting from scratch when all of the new deltas come in. In the case where we haven't yet connected to any model yet we do not want to empty out the db because the ecs and db rely on each other for any changes that still need to be applied to the model. 

This means that we will need a flag to indicate that, not only are we in gisf, but that we're switching models and not just haven't yet connected or we'll introduce a bug switching models here.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4761/
Test FAILed.

development.ini
```diff
@@ -15,7 +15,7 @@ jujugui.combine = true
 jujugui.gzip = true
 jujugui.cachebuster = dev
 jujugui.socketTemplate = /environment/$uuid/api
-
+jujugui.gisf = false
```
huwshimi: Might be worth adding a comment where this variable is used so we know what it means or maybe even rename it to something more self explanatory :).

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4767/
Test FAILed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4772/
Test FAILed.

makyo:
:+1:

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4774/
Test PASSed.

hatched:
Thanks @makyo :shipit:

jujugui:

PR #1509, opened 2016-04-04
Create a new model from the profile page.

kadams54:
Fixes https://github.com/juju/juju-gui/issues/1528.

### DIFF
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -1738,6 +1738,13 @@ YUI.add('juju-gui', function(Y) {
       // route dispatch proceed. On env connection change, the app will
       // re-dispatch and this route callback will be executed again.
       if (!this.env || !this.env.get('connected')) {
+        // If connected to a jem, we may be in a disconnected state while
+        // waiting for the new model to be defined as part of the deployment
+        // flow. In that case, carry on as normal.
+        if (this.jem) {
+          this.maskVisibility(false);
+          next();
+        }
         return;
       }
       var credentials = this.env.getCredentials();
@@ -1919,15 +1926,17 @@ YUI.add('juju-gui', function(Y) {
     /**
       Switch the application to another environment.
       Disconnect the current WebSocket connection and establish a new one
-      pointed to the environment referenced by the given unique identifier.
+      pointed to the environment referenced by the given URL.
 
       @method switchEnv
-      @param {String} uuid The environment UUID where to switch to.
+      @param {String} socketUrl The URL for the environment's websocket.
       @param {String} username The username for the new environment.
       @param {String} password The password for the new environment.
+      @param {Boolean} reconnect Whether to reconnect to a new environment; by
+                                 default, if the socketUrl is set, we assume we
+                                 want to reconnect to the provided URL.
     */
-    switchEnv: function(socketUrl, username, password) {
-      console.log('switching to new socket URL:', socketUrl);
+    switchEnv: function(socketUrl, username, password, reconnect=!!socketUrl) {
       if (this.get('sandbox')) {
         console.log('switching models is not supported in sandbox');
       }
@@ -1947,7 +1956,9 @@ YUI.add('juju-gui', function(Y) {
       // Disconnect and reconnect the environment.
       this.env.ws.onclose = function(event) {
         this.env.on_close();
-        this.env.connect();
+        if (reconnect) {
+          this.env.connect();
+        }
       }.bind(this);
       this.env.close();
       this.db.reset();
--- a/jujugui/static/gui/src/app/components/generic-button/generic-button.js
+++ b/jujugui/static/gui/src/app/components/generic-button/generic-button.js
@@ -42,7 +42,7 @@ YUI.add('generic-button', function() {
     },
 
     /**
-      Call the action if not disabled.  
+      Call the action if not disabled.
 
       @method _handleClick
       @param {Object} e The click event.
--- a/jujugui/static/gui/src/app/components/user-profile/_user-profile.scss
+++ b/jujugui/static/gui/src/app/components/user-profile/_user-profile.scss
@@ -1,4 +1,9 @@
 .user-profile {
+    &__create-new {
+        float: right;
+        font-size: 1rem;
+    }
+
     &__close {
         position: fixed;
         z-index: 1;
@@ -11,7 +16,9 @@
         margin-top: 10px;
         padding-bottom: 20px;
         padding-left: 15px;
+        padding-right: 15px;
         font-size: 1.375em;
+        line-height: 40px;
     }
 
     &__empty {
--- a/jujugui/static/gui/src/app/components/user-profile/test-user-profile.js
+++ b/jujugui/static/gui/src/app/components/user-profile/test-user-profile.js
@@ -273,6 +273,12 @@ describe('UserProfile', () => {
               <span className="user-profile__size">
                 ({1})
               </span>
+              <div className='user-profile__create-new'>
+                <juju.components.GenericButton
+                  action={instance.switchModel}
+                  type='inline-neutral'
+                  title='Create new' />
+              </div>
             </div>
             <ul className="user-profile__list twelve-col">
               <li className="user-profile__list-header twelve-col">
@@ -297,7 +303,7 @@ describe('UserProfile', () => {
                 entity={models[0]}
                 expanded={false}
                 key="env1"
-                switchModel={instance.switchEnv}
+                switchModel={instance.switchModel}
                 type="model">
                 <span className="user-profile__list-col three-col">
                   sandbox
@@ -323,6 +329,7 @@ describe('UserProfile', () => {
               <span className="user-profile__size">
                 ({1})
               </span>
+              {undefined}
             </div>
             <ul className="user-profile__list twelve-col">
               <li className="user-profile__list-header twelve-col">
@@ -383,6 +390,7 @@ describe('UserProfile', () => {
               <span className="user-profile__size">
                 ({1})
               </span>
+              {undefined}
             </div>
             <ul className="user-profile__list twelve-col">
               <li className="user-profile__list-header twelve-col">
@@ -647,12 +655,12 @@ describe('UserProfile', () => {
     listEnvs.args[0][1](envs);
     // Call the method that's passed down. We test that this method is correctly
     // passed down in the initial 'happy path' full rendering test.
-    instance.switchEnv('abc123', 'modelname');
+    instance.switchModel('abc123', 'modelname');
     // Make sure we show the canvas loading mask when switching models.
     assert.equal(showMask.callCount, 1);
     // We need to call to generate the proper socket URL.
     // Check that switchModel is called with the proper values.
-    assert.equal(switchModel.callCount, 1, 'switchEnv not called');
+    assert.equal(switchModel.callCount, 1, 'switchModel not called');
     assert.deepEqual(switchModel.args[0], ['abc123', [{
       uuid: 'abc123', user: 'foo', password: 'bar'}]]);
     // The database needs to be updated with the new model name.
@@ -749,5 +757,48 @@ describe('UserProfile', () => {
     renderer.unmount();
     assert.equal(charmstoreAbort.callCount, 2);
     assert.equal(listEnvsAbort.callCount, 1);
+  });
+
+  it('can reset the model connection', () => {
+    var pluralize = sinon.stub();
+    pluralize.withArgs('model', sinon.match.any).returns('models');
+    pluralize.withArgs('bundle', sinon.match.any).returns('bundles');
+    pluralize.withArgs('charm', sinon.match.any).returns('charms');
+    var utilsSwitchModel = sinon.stub();
+    var dbEnvironmentSet = sinon.stub();
+    var changeState = sinon.stub();
+    var renderer = jsTestUtils.shallowRender(
+      <juju.components.UserProfile
+        switchModel={utilsSwitchModel}
+        users={users}
+        listEnvs={sinon.stub()}
+        showConnectingMask={sinon.stub()}
+        changeState={changeState}
+        charmstore={{}}
+        dbEnvironmentSet={dbEnvironmentSet}
+        env={env}
+        getDiagramURL={sinon.stub()}
+        gisf={false}
+        interactiveLogin={true}
+        storeUser={sinon.stub()}
+        pluralize={pluralize}
+        user={users.charmstore} />, true);
+    var instance = renderer.getMountedInstance();
+    instance.switchModel();
+    assert.equal(utilsSwitchModel.callCount, 1,
+                 'Model disconnect not called');
+    var switchArgs = utilsSwitchModel.args[0];
+    assert.equal(switchArgs[0], undefined,
+                 'UUID should not be defined');
+    assert.equal(dbEnvironmentSet.callCount, 1,
+                 'Model name not set in db');
+    var dbEnvSetArgs = dbEnvironmentSet.args[0];
+    assert.equal(dbEnvSetArgs[1], 'untitled_model');
+    assert.deepEqual(changeState.getCall(0).args[0], {
+      sectionC: {
+        component: null,
+        metadata: null
+      }
+    }, 'App state not reset');
   });
 });
--- a/jujugui/static/gui/src/app/components/user-profile/user-profile.js
+++ b/jujugui/static/gui/src/app/components/user-profile/user-profile.js
@@ -216,21 +216,22 @@ YUI.add('user-profile', function() {
         this.setState({bundleList: data});
       }
     },
-
     /**
-      Take the supplied UUID, fetch the username and password then call the
-      passed in switchModel method.
 
-      @method switchEnv
-      @param {String} uuid The env UUID.
-      @param {String} name The env name.
+      Take the supplied UUID, fetch the username and password then call the
+      passed in switchModel method. If the UUID and name are null, simply
+      disconnect without reconnecting to a new model.
+
+      @method switchModel
+      @param {String} uuid The model UUID.
+      @param {String} name The model name.
     */
-    switchEnv: function(uuid, name) {
+    switchModel: function(uuid, name) {
       var props = this.props;
       props.showConnectingMask();
-      props.switchModel(uuid, this.state.envList);
-      props.dbEnvironmentSet('name', name);
       this.close();
+      props.switchModel(uuid, this.state.envList);
+      props.dbEnvironmentSet('name', name || 'untitled_model');
     },
 
     close: function() {
@@ -281,7 +282,7 @@ YUI.add('user-profile', function() {
           entity={model}
           expanded={model.name === this.props.currentModel}
           key={uuid}
-          switchModel={this.switchEnv}
+          switchModel={this.switchModel}
           type="model">
           <span className="user-profile__list-col three-col">
             {model.name || '--'}
@@ -551,10 +552,19 @@ YUI.add('user-profile', function() {
       var header;
       var rows = [];
       var title;
+      var actions;
       if (type === 'models') {
         generateRow = this._generateModelRow;
         header = this._generateModelHeader();
         title = 'Models';
+        actions = (
+          <div className="user-profile__create-new">
+            <juju.components.GenericButton
+              action={this.switchModel}
+              type='inline-neutral'
+              title='Create new' />
+          </div>
+        );
       } else if (type === 'bundles') {
         generateRow = this._generateBundleRow;
         header = this._generateBundleHeader();
@@ -574,6 +584,7 @@ YUI.add('user-profile', function() {
             <span className="user-profile__size">
               ({list.length})
             </span>
+            {actions}
           </div>
           <ul className="user-profile__list twelve-col">
             {header}
--- a/jujugui/static/gui/src/app/views/utils.js
+++ b/jujugui/static/gui/src/app/views/utils.js
@@ -1913,23 +1913,28 @@ YUI.add('juju-view-utils', function(Y) {
   */
   utils.switchModel = function(createSocketURL, switchEnv, uuid, modelList) {
     var username, password, address, port;
-    var found = modelList.some((model) => {
-      if (model.uuid === uuid) {
-        username = model.user;
-        password = model.password;
-        if (model['host-ports']) {
-          var hostport = model['host-ports'][0].split(':');
-          address = hostport[0];
-          port = hostport[1];
+    if (uuid && modelList) {
+      var found = modelList.some((model) => {
+        if (model.uuid === uuid) {
+          username = model.user;
+          password = model.password;
+          if (model['host-ports']) {
+            var hostport = model['host-ports'][0].split(':');
+            address = hostport[0];
+            port = hostport[1];
+          }
+          return true;
         }
-        return true;
+      });
+      if (!found) {
+        console.log('No user credentials for model: ', uuid);
       }
-    });
-    if (!found) {
-      console.log('No user credentials for model: ', uuid);
+      var socketUrl = createSocketURL(address, port, uuid);
+      switchEnv(socketUrl, username, password);
+    } else {
+      // Just reset without reconnecting to an env.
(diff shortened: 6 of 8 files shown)


jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4765/
Test FAILed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4786/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4787/
Test PASSed.

jujugui/static/gui/src/app/components/user-profile/user-profile.js
```diff
@@ -224,6 +224,18 @@ YUI.add('user-profile', function() {
...
 
+    /**
+      Set to a new, blank canvas with no model.
+
+      @method resetModel
+    */
+    resetModel: function() {
+      this.props.switchModel();
+      // XXX kadams54: What to do here?
+      this.props.dbEnvironmentSet('name', null);
```
hatched: In the upcoming designs the breadcrumb will have an input which in which the user can specify the name before going through the flow. I'd imaging at this point we'd probably want to call it something like 'model1' (assuming there isn't a model1 already)

jujugui/static/gui/src/app/app.js
```diff
@@ -1894,7 +1894,12 @@ YUI.add('juju-gui', function(Y) {
       @param {String} password The password for the new environment.
     */
     switchEnv: function(socketUrl, username, password) {
```
hatched: You can actually do something cool here for the reconnect while also allowing it to be a param
```JavaScript
  switchEnv: function(socketUrl, username, password, reconnect=!!socketUrl) {
```

jujugui/static/gui/src/app/app.js
```diff
@@ -1907,14 +1912,18 @@ YUI.add('juju-gui', function(Y) {
           password: password
         });
       };
-      // Tell the environment to use the new socket URL when reconnecting.
-      this.env.set('socket_url', socketUrl);
+      if (reconnect) {
```
hatched: Is there really any reason to not allow this to set the socket_url? Sure it'll be invalid but we aren't connecting so that's ok?

jujugui/static/gui/src/app/components/user-profile/user-profile.js
```diff
@@ -224,6 +224,18 @@ YUI.add('user-profile', function() {
       this.close();
     },
 
+    /**
+      Set to a new, blank canvas with no model.
+
+      @method resetModel
+    */
+    resetModel: function() {
+      this.props.switchModel();
```
hatched: This is kind of confusing, I feel like it should take an argument to signify that it's being set to empty....like maybe
```JavaScript
  this.props.switchModel(null)
```

jujugui/static/gui/src/app/views/utils.js
```diff
@@ -1910,23 +1910,28 @@ YUI.add('juju-view-utils', function(Y) {
...
+        console.log('No user credentials for model: ', uuid);
       }
-    });
-    if (!found) {
-      console.log('No user credentials for model: ', uuid);
+      var socketUrl = createSocketURL(address, port, uuid);
+      switchEnv(socketUrl, username, password);
+    } else {
+      // Just reset without reconnecting to an env.
+      switchEnv();
```
hatched: `switchEnv(null)` similar to `resetModel` above

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4788/
Test PASSed.

kadams54:
## QA

1. Load a blank canvas.
2. Click on the "admin" link (assuming you're running in sandbox, locally) to pull up the profile.
3. Click the "Create new" button to nav back to the root URL.

Right now the app doesn't seem to be handling state properly to unmount the profile. Still investigating whether that is a bug in this code, or a pre-existing bug in our app state logic.

jujugui/static/gui/src/app/views/utils.js
kadams54: I started to change to `switchEnv(null)`, but that felt wrong since the first param is `socketUrl`, not the "env". That is, normal invocations of this function are `switchEnv('http://...', ...)`, not `switchEnv(myEnvObj)`. In the end I left it as is, hoping the comment will help make up for the unclear code.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4791/
Test PASSed.

jujugui/static/gui/src/app/app.js
```diff
@@ -1892,9 +1892,14 @@ YUI.add('juju-gui', function(Y) {
       @param {String} uuid The environment UUID where to switch to.
       @param {String} username The username for the new environment.
       @param {String} password The password for the new environment.
+      @param {Boolean} reconnect Whether to reconnect to a new environment.
     */
-    switchEnv: function(socketUrl, username, password) {
-      console.log('switching to new socket URL:', socketUrl);
+    switchEnv: function(socketUrl, username, password, reconnect=!!socketUrl) {
+      if (reconnect) {
+        console.log('switching to new socket URL:', socketUrl);
```
huwshimi: Do we need these logs? Are they more of a debugging thing? If it's a user facing thing can we put it in the UI as a notification?

jujugui/static/gui/src/app/app.js
```diff
@@ -1892,9 +1892,14 @@ YUI.add('juju-gui', function(Y) {
...
+      } else {
+        console.log('resetting to no env');
```
huwshimi: Same as above.

jujugui/static/gui/src/app/app.js
```diff
@@ -1892,9 +1892,14 @@ YUI.add('juju-gui', function(Y) {
       @param {String} uuid The environment UUID where to switch to.
       @param {String} username The username for the new environment.
       @param {String} password The password for the new environment.
+      @param {Boolean} reconnect Whether to reconnect to a new environment.
     */
-    switchEnv: function(socketUrl, username, password) {
-      console.log('switching to new socket URL:', socketUrl);
+    switchEnv: function(socketUrl, username, password, reconnect=!!socketUrl) {
```
huwshimi: It doesn't look like you ever pass in a reconnect param so maybe that should be moved to a scoped var rather than assigning the var using the param defaults.

huwshimi: That way you could also add a comment as to why we're reconnecting when there is no socketUrl.

jujugui/static/gui/src/app/components/generic-button/generic-button.js
```diff
@@ -35,6 +36,7 @@ YUI.add('generic-button', function() {
     _generateClasses: function() {
       return classNames(
         this.props.type ? 'button--' + this.props.type : 'button--neutral',
+        this.props.classNames,
```
huwshimi: I'm not a huge fan of adding more class names to the buttons in this way. You could use the type prop or add a wrapping element.

jujugui/static/gui/src/app/components/user-profile/_user-profile.scss
```diff
@@ -9,6 +9,11 @@
     border-top: 1px solid $mid-grey;
     box-shadow: none;
 
+    &__create-new {
+      float: right;
```
huwshimi: Four spaces please.

jujugui/static/gui/src/app/components/user-profile/user-profile.js
```diff
@@ -224,6 +224,52 @@ YUI.add('user-profile', function() {
...
+      @method resetModel
+    */
+    resetModel: function() {
+      var props = this.props;
+      // Reset the connection without reconnecting, so no need to pass any
+      // params.
+      props.switchModel();
+      // XXX 2016-04-07 kadams54: the random name generator is intended to be a
+      // temporary solution until we fully implement deployment flow.
+      props.dbEnvironmentSet('name', this.generateRandomName());
```
huwshimi: Once the model name input is working on the deploy flow will you need to set a name here at all?

huwshimi:
Thanks for this tricky bit of work @kadams54. A few comments above.

urosj:
QA failed: tested it with juju2 beta4, with juju upgrade-gui from dist file, running with juju gui. So, this was on a real controller and clicking from profile page to "create new" led me to a canvas, but environment switcher, midpoint and all other panes were inactive, URL changes, but nothing happens. I am also unable to go back to the profile pane. 

@huwshimi will try it out on a real model in a standalone mode deployed through a charm.

urosj:
Next try was "make run" in sandbox mode. There profile page loads and you can click on manage button and you get back to canvas. Once you click "create new", nothing happens. If you then click on "manage", you get a spinner and never go back to canvas. Only refresh of the page in the browser gets you out of spinner situation.

huwshimi:
@urosj I'm seeing the same thing in standalone. After creating a new model nothing responds to clicks properly and clicking on the model switcher gives a console error `Uncaught INVALID_STATE_ERR : Pausing to reconnect websocket`.

jujugui/static/gui/src/app/app.js
kadams54: https://github.com/juju/juju-gui/pull/1509#discussion-diff-58904948

I'll let you and @hatched battle this one out :-)

Just to be clear: we reconnect only if there *is* a socketUrl.

jujugui/static/gui/src/app/components/user-profile/user-profile.js
kadams54: I think whatever value is in that input will need to eventually find its way here, to be set in the DB.

kadams54:
Yes, that is the expected behavior. I probably should have made it more clear in my QA instructions.

Right now the URL changes but the profile is not replaced by a blank canvas. After discussing with @hatched yesterday we felt like this might be due to a state bug. That is, the UserProfile component is doing the right thing (`this.props.changeState({sectionC:{component: null, metadata: null}})`) to go back to the state we want the app in, but something is going awry in the app state and as a result the UserProfile isn't cleared.

I initially tried to debug and fix as a drive-by in this PR, but it ended up being more complex than originally thought, so I split it off into it's own card to get this one unstuck.

jujugui/static/gui/src/app/app.js
kadams54: The logs pre-dated my work on this function, so I just preserved them. They seem to be debugging rather than user-related. I have no objections to dropping them.

jujugui/static/gui/src/app/components/generic-button/generic-button.js
kadams54: Will do.

jujugui/static/gui/src/app/components/user-profile/_user-profile.scss
kadams54: :+1:

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4792/
Test PASSed.

kadams54:
Note that this "navigation out of the profile" issue may also be connected to https://github.com/juju/juju-gui/issues/1527 and https://github.com/juju/juju-gui/issues/1528.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4793/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4794/
Test PASSed.

kadams54:
OK, bugs have been squashed. When you click on "Create new" you leave the profile. That said, because there's no process/UI yet for actually creating and connecting to a new model, nothing works. This feels like it gets us halfway there in terms of solving the "OK, how do I navigate back out of my profile?" question.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4795/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4796/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4852/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4853/
Test PASSed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4860/
Test FAILed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4861/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4865/
Test PASSed.

kadams54:
## QA, Round 2

1. Load a blank canvas.
2. Click on the "admin" link (assuming you're running in sandbox, locally) to pull up the profile.
3. Click the "Create new" button to nav back to the root URL.

Ideally, the functionality should be tested in a variety of situations: sandbox, GI Joe, GISF, etc.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/4866/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4867/
Test PASSed.

jujugui/static/gui/src/app/components/user-profile/user-profile.js
huwshimi: Now that we can set the name in the deployment flow can this be updated to not use a random name?

jujugui/static/gui/src/app/app.js
huwshimi: Either way, I still think this needs a comment as to why we want to reconnect if there is a socketURL.

jujugui/static/gui/src/app/components/user-profile/user-profile.js
```diff
@@ -228,9 +228,56 @@ YUI.add('user-profile', function() {
...
       props.switchModel(uuid, this.state.envList);
       props.dbEnvironmentSet('name', name);
+    },
+
+    /**
+      Set to a new, blank canvas with no model.
+
+      @method resetModel
+    */
+    resetModel: function() {
```
huwshimi: I wonder if this is something that should be done outside of the user profile, might save some code duplication later (e.g. shouldn't the new button in the model switcher do the same thing?).

jujugui/static/gui/src/app/app.js
kadams54: I'll clarify in the docstring for that function.

jujugui/static/gui/src/app/components/user-profile/user-profile.js
kadams54: Where is the name set/stored?

jujugui/static/gui/src/app/app.js
```diff
@@ -1727,7 +1727,13 @@ YUI.add('juju-gui', function(Y) {
       // route dispatch proceed. On env connection change, the app will
       // re-dispatch and this route callback will be executed again.
       if (!this.env || !this.env.get('connected')) {
-        return;
+        if (this.jem) {
```
hatched: Can you add a comment here as to why this is necessary

jujugui/static/gui/src/app/components/user-profile/user-profile.js
hatched: This is a good point

jujugui/static/gui/src/app/components/user-profile/user-profile.js
hatched: The name gets set in the deployment flow component - the only reason this is necessary is because the breadcrumb needs to have /something/ in it. After talking with @Lucapaulina it's decided that the environment name in the breadcrumb can be set to 'untitled_model' which means you can drop all this random name stuff

jujugui/static/gui/src/app/components/user-profile/user-profile.js
hatched: The logic here and the logic in `switchEnv` above are nearly identical and should probably be abstracted out and moved into the apps `switchEnv` passed through the `switchModel`

jujugui/static/gui/src/app/components/user-profile/user-profile.js
kadams54: Will implement.

jujugui/static/gui/src/app/app.js
kadams54: Done.

jujugui/static/gui/src/app/components/user-profile/user-profile.js
kadams54: Will make the change.

kadams54: So… having actually done the coding… I'm not sure it's really worth it. First off, this method really isn't reusable outside of `user-profile.js` because it's a wrapper around the more generic `utils.switchModel`. `utils.switchModel` is what other similar functionality should call. This function does a few other extra `UserProfile`-specific things like popping up the closing the profile.

The only function call that isn't `UserProfile`-specific is setting the model name in the DB. The problem there is that `this.db.evironment` is only in `app.js`, which the model name only goes down to the `switchModel` in `utils.js`. I think the right solution there is to pass `this.db.environment` into `utils.js`.

When all is said and done, there are two changes that seem worthwhile:

1. Combine `switchEnv` and `resetModel` into one function in `user-profile.js`.
2. Move the `dbEnvironmentSet` invocation into the utils' `switchModel`.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4875/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4876/
Test PASSed.

jujugui/static/gui/src/app/app.js
```diff
@@ -1727,7 +1727,16 @@ YUI.add('juju-gui', function(Y) {
...
       if (!this.env || !this.env.get('connected')) {
-        return;
+        // If connected to a jem, we may be in a disconnected state while
+        // waiting for the new model to be defined as part of the deployment
+        // flow. In that case, carry on as normal.
+        if (this.jem) {
+          this.set('loggedIn', false);
+          this.maskVisibility(false);
+          next();
+        } else {
```
hatched: This else block can be removed and the surrounding conditional (!this.env) should simply end in `return;` That way it maintains it's original functionality while special casing the (this.jem) case.

jujugui/static/gui/src/app/app.js
```diff
@@ -1908,15 +1917,17 @@ YUI.add('juju-gui', function(Y) {
...
+      @param {String} socketUrl The URL for the environment's websocket.
       @param {String} username The username for the new environment.
       @param {String} password The password for the new environment.
+      @param {Boolean} reconnect Whether to reconnect to a new environment; by
+                                 default, if the socketUrl is set, we assume we
+                                 want to reconnect to the provided URL.
     */
-    switchEnv: function(socketUrl, username, password) {
-      console.log('switching to new socket URL:', socketUrl);
+    switchEnv: function(socketUrl, username, password, reconnect=!!socketUrl) {
```
hatched: I don't see any tests for the modifications done to this method.

jujugui/static/gui/src/app/components/user-profile/test-user-profile.js
```diff
@@ -750,4 +758,41 @@ describe('UserProfile', () => {
     assert.equal(charmstoreAbort.callCount, 2);
     assert.equal(listEnvsAbort.callCount, 1);
   });
+
+  it('can reset the model connection', () => {
+    var pluralize = sinon.stub();
+    pluralize.withArgs('model', sinon.match.any).returns('models');
+    pluralize.withArgs('bundle', sinon.match.any).returns('bundles');
+    pluralize.withArgs('charm', sinon.match.any).returns('charms');
+    var switchModel = sinon.stub();
```
hatched: to avoid confusion could you call this something like `utilsSwitchModel`

jujugui/static/gui/src/app/components/user-profile/test-user-profile.js
```diff
@@ -750,4 +758,41 @@ describe('UserProfile', () => {
...
+        dbEnvironmentSet={dbEnvironmentSet}
+        env={env}
+        getDiagramURL={sinon.stub()}
+        interactiveLogin={true}
+        storeUser={sinon.stub()}
+        pluralize={pluralize}
+        user={users.charmstore} />, true);
+    var instance = renderer.getMountedInstance();
+    instance.switchModel();
+    assert.equal(switchModel.callCount, 1,
```
hatched: because the params that this `utilsSwitchModel` is called with are very important they should be checked here. Same goes for the `dbEnvironmentSet` call below.

jujugui/static/gui/src/app/components/user-profile/user-profile.js
```diff
@@ -216,21 +216,27 @@ YUI.add('user-profile', function() {
...
       var props = this.props;
       props.showConnectingMask();
-      props.switchModel(uuid, this.state.envList);
-      props.dbEnvironmentSet('name', name);
       this.close();
+      if (!uuid && !name) {
+        // No parameters = disconnect with no reconnect.
+        props.switchModel();
+      } else {
+        props.switchModel(uuid, this.state.envList);
```
hatched: because the utils `switchModel` method requires the existence of this uuid then I feel this entire conditional can be truncated down to simply `props.switchModel(uuid, this.state.envList);` and rely on the utils `switchModel` to do the right thing.

jujugui/static/gui/src/app/views/utils.js
```diff
@@ -1913,23 +1913,28 @@ YUI.add('juju-view-utils', function(Y) {
   */
   utils.switchModel = function(createSocketURL, switchEnv, uuid, modelList) {
```
hatched: I don't see any test additions/changes for these modifications

kadams54: Tests added.

jujugui/static/gui/src/app/app.js
kadams54: Tests added.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4877/
Test PASSed.

huwshimi:
In the sandbox if I Create new it works but then if I open the profile and click Create new again I get an error in the console "Uncaught INVALID_STATE_ERR : Connection is closed."

kadams54:
@huwshimi Good point. We should probably hide the "Create new" button if we're already in a disconnected state.

jaycee:
QA notes:

* Deployed on trusty with juju 1.25.5-wily. Set gisf-enabled to true and set jem-location to staging.
* As I was logged in to jem I got my username for the profile link; clicking it took me to the profile page.
* Clicked the "Create new" button; I was taken back to the canvas with the initial unconnected state.

Console showed `TypeError: this.env.ws is undefined`; this is expected as there's currently a branch in progress to enable switching to environments from unconnected state.

I'll see what happens if I apply the fix for that bug.

jaycee:
After applying the patch, the ws undefined message is gone. On a whim, I also did the following:

* Switched to one of my existing, deployed environments.
* Went to the profile and selected create new.

I was dropped into a new blank canvas in the unconnected state, as expected. The breadcrumbs still showed me on my existing env (creatively named foo), but I believe we have a bug for breadcrumbs being wrong in this sort of situation.

QA OK for gisf enabled on trusty.

kadams54:
@huwshimi After talking with @hatched we're going to defer "hide the button when disconnected" until https://github.com/juju/juju-gui/issues/1569 is fixed. Once there's a deployment flag it'll be easy to hide/show the button based on how that flag is set.

jaycee:
:+1: from me, thanks @kadams54.

hatched:
:+1: Thanks for stickin with this one!

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4890/
Test PASSed.

hatched:
:shipit:

jujugui:

PR #1533, opened 2016-04-11
Mobile spike

anthonydillon:
Added a mobile styling to style the GUI under the small screen breakpoint


anthonydillon:
## QA
- Load the GUI on a mobile
- Check the views look OK
- Check there are no errors
- See that all visible functionality is usable

_I created a separate style sheet to for the overrides to make it easier to maintain in the early stages. These styles should move into the components once these styles settle._

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4800/
Test PASSed.

jujugui/static/gui/src/app/assets/css/base.scss
```diff
@@ -81,10 +81,10 @@ $z-indexed-elements: search-results-list-block__tags,
     search-results-cell--hover,
     boolean-config--handle,
     boolean-config--label,
+    inspector-view,
```
kadams54: Intentional change?

jujugui/static/gui/src/app/views/utils.js
```diff
@@ -1330,7 +1330,8 @@ YUI.add('juju-view-utils', function(Y) {
         navbar = Y.one('.header-banner'),
         viewport = Y.one('#viewport'),
         result = {height: minheight || 0, width: minwidth || 0};
-    if (containerHeight && navbar && viewport) {
+    if (containerHeight && navbar && viewport &&
+      viewport.getComputedStyle('width') !== 'auto') {
```
kadams54: Not sure what exactly is happening here?

jujugui/static/gui/src/app/assets/css/base.scss
anthonydillon: Yea, I had to change the layer logic here so when you load machine view it appears inferring of the inspector. Saves having to unload and reload the inspector in switch. 

Smoke and mirrors on small screens

jujugui/static/gui/src/app/views/utils.js
anthonydillon: So it was getting an error because it sets the height and width based on the viewport. But when an element is set to `display: none` it sets the width to auto. This was resulting in a NaN further down in this statement when the width value is computed.

anthonydillon: I set the viewport element to `display: none ` in small screens.

kadams54:
## QA Notes

### Search Results

* Series dropdown extends off edge of screen.
* Layout within each recommended result seems wonky; less than optimal with the tiny icon on its own line. That layout works better for bundles than for individual charms.
Deployment Summary
* “Leave unplaced” radio buttons don’t flow very well; probably should be `display: block` for each.

### Inspector
* Click targets are a bit wonky for Change version. I was trying to change the version and kept ending up at the charm details page for the current version.
* No real indicators that “Units”, “Configure”, “Expose”, and “Change version” are tappable.
* Scale up UI has some issues with the placeholder text in the input.

### Machine View
* Manually placing new units doesn’t seem to work; I’m guessing that’s true of most drag-n-drop interactions in MV.

### User Profile
* The “X model Y bundles Z charms” bit at the top wraps poorly in mobile.

Doesn’t seem to be any way to create relations?

All in all I think this is a really good start at a mobile version of the UI. I was surprised by how usable it was. I'm also not sure how many of these need to be fixed in this PR versus followup polish work. I'll leave that to @anthonydillon's discretion.

Lucapaulina:
We're going to work on adding the ability to build relations this week or
next week.

anthonydillon:
Thanks very much Kyle, I'll do another pass on this branch and see how much
I can resolve. I'll add issues for the ones I leave out.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4842/
Test PASSed.

anthonydillon:
# Updates

### Search Results

* Series dropdown extends off edge of screen.
** The filter dropdowns are now `width: 100%;` so they show on there on line.

* Layout within each recommended result seems wonky; less than optimal with the tiny icon on its own line. That layout works better for bundles than for individual charms.
** This is the current mobile design im afraid

### Deployment Summary
* “Leave unplaced” radio buttons don’t flow very well; probably should be `display: block` for each.
** The radio buttons fit side by side with a little tweaking to the spacing around them

### Inspector
* Click targets are a bit wonky for Change version. I was trying to change the version and kept ending up at the charm details page for the current version.
** I have removed the link to the current version making clicking on the Change version tab easier. The current version link is available on the change version window. I have made the Downgrade and Upgrade buttons in the change version window display at all times as we dont have hover on mobile.

* No real indicators that “Units”, “Configure”, “Expose”, and “Change version” are tappable.
** I think this is one for UX/Design to think about. (_For a later branch_)

* Scale up UI has some issues with the placeholder text in the input.
** I dont see any overlapping on both the iPhone 6 and Galaxy S5

### Machine View
* Manually placing new units doesn’t seem to work; I’m guessing that’s true of most drag-n-drop interactions in MV.
** I seem to be able to use the drop down and select Deploy to. The flow seems to work on both mobiles for me. Is there an edge case I'm missing?

### User Profile
* The “X model Y bundles Z charms” bit at the top wraps poorly in mobile.
** Reworked the profile header to look nicer and accommodate long usernames and models names

jujugui/static/gui/src/app/views/utils.js
hatched: Could you add a comment to explain why you're doing this.

hatched:
Is this how we want to deal with media queries on components going forward? I would have thought that we'd want to keep the media query definitions in the components themselves. Now we have to know to modify some styling in the components css and some in some other files which aren't clear.

anthonydillon:
We do want the media queries in the components I believe but put them in a mobile file for now to make development and reviewing easier. Also reduces the chance of style leakage while the GUI is becoming stable. I mentioned that once the mobile experience has settled down we should port the styles into the components. They are commented by component so should be easy enough to move over.

hatched:
Good to hear, can you make a tracking issue for this so that we don't forget to do it in the future?

kadams54:
All the addressed issues look good. I did run into two new items:

image

The onboarding in MV's machine column talks about drag-and-drop, which I suspect is why I ran into problems initially. Also note: 1) the spacing below the onboarding text and 2) the border around the current machine, which really doesn't work without the container column next to it.

image

Model listing on the user profile page needs some TLC, which is not surprising since this page is currently under development.

I'd be fine with landing this branch and tackling the items above in follow-up PRs.

QA is 🆗  and code is 👍  - nice work on a really big undertaking!

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4869/
Test PASSed.

anthonydillon:
Thanks @kadams54 I fixed up the machine view and profile. Looks much better now thanks.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4870/
Test PASSed.

kadams54:
Looks great, thanks!

makyo:
:+1:!

hatched:
:shipit:

jujugui:

PR #1535, opened 2016-04-11
WIP - Change the asset-path used by vanilla to point locally

anthonydillon:


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4802/
Test PASSed.

jujugui/static/gui/src/app/assets/css/base.scss
```diff
@@ -37,6 +37,8 @@ for indentation.
 $font-family: 'Ubuntu', 'Helvetica Neue', Helvetica, Arial, sans-serif;
 $font-family-mono: 'Ubuntu Mono', 'Ubuntu-Mono', 'Bitstream Vera Sans Mono', 'Courier New', Courier, monospace;
 
+$asset-path: 'juju-ui/assets/';
```
hatched: This is interesting because the asset path in Juju 2.0 is actually different than the root path that the gui is loaded from so we actually have to get that value from juju on load. So this won't work for Juju 2.0 story.

anthonydillon: There will not be a standard place for assets within the app? If that is
the case we will not be able to load any assets from CSS in the GUI. I will
work out a way of removing the font load on Vanilla if you want and then we
can inject the font into the the head of the index.

Or I'll play with base64'ing the font into Vanilla and see what size that
adds to it.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4812/
Test PASSed.

anthonydillon:
This branch now works locally.

anthonydillon:
Superseded by https://github.com/juju/juju-gui/pull/1593

PR #1607, opened 2016-04-30
WIP: Support creating models in gisf.

kadams54:


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4942/
Test PASSed.

jujugui/static/gui/src/app/app.js
```diff
@@ -1933,6 +1933,16 @@ YUI.add('juju-gui', function(Y) {
       if (this.get('sandbox')) {
         console.log('switching models is not supported in sandbox');
       }
+      // Optionally update to the new base URL.
+      if (this.get('gisf')) {
+        var state = this.state;
+        var baseUrl = state.get('baseUrl');
+        var unsavedModelUrl = 'sandbox';
+        // Make sure we're not already in an unsaved model.
+        if (baseUrl.indexOf(unsavedModelUrl) < 0) {
```
anthonydillon: This seems like an unsafe way of telling if the model is unsaved. Do we not have a specific flag for this? What if a user called there real model sandbox?

hatched: You can tell if you're in a real model by seeing if you're connected to one, have all of the credentials, etc.

jujugui/static/gui/src/test/test_app.js
```diff
@@ -1241,6 +1241,29 @@ describe('App', function() {
...
+          this._state[key] = value;
+        }
+      };
+      app.set('gisf', true);
+      app.switchEnv();
+      assert.equal(app.state.get('baseUrl'), '/sandbox',
+                   'Base url not updated');
+      app.switchEnv();
+      // Make sure base URL isn't updated unecessarily.
+      assert.equal(app.state.get('baseUrl'), '/sandbox',
```
hatched: You're checking that it's being set to sandbox when you're switching to sandbox twice in a row?

hatched:
I'm not really sure what this branch is to accomplish, just by updating the baseUrl it won't update the url until the next time it navigates. You'll need to navigate after setting the baseUrl to not rely on side effects.

kadams54:
Navigation happens as the URL goes from `/profile` to `/`, but unfortunately I haven't figured out yet how to get it to pick up the new `baseUrl` before that nav.

jujugui/static/gui/src/test/test_app.js
kadams54: Right. What we don't want is to end up at `/sandbox/sandbox`. It should be an idempotent operation.

jujugui/static/gui/src/app/app.js
kadams54: Yeah, sorry, the comment is unclear. Two things:

1. "sandbox" is the special URL used in gisf to denote a disconnected GUI.
2. The point here isn't really to see if the model is unsaved. It's more to make sure that the baseUrl update is idempotent; that is, that we're not adding multiple instances of "sandbox" to the URL.

hatched: Ok so lets change `unsavedModelUrl` to be something like `gisfDisconnectedModelName` and add code comments for both points so that they are very clear in the future.

jujugui/static/gui/src/test/test_app.js
hatched: Alright can you clarify the code comment to that effect plz.

hatched:
Having this logic in this `switchEnv` method doesn't feel right, I feel like it should probably be moved out into a separate method. We only want to update the path once the switch has been made and not before as there are a number of places which could prevent the switch. I believe this is the best place to then call that method from https://github.com/juju/juju-gui/blob/develop/jujugui/static/gui/src/app/app.js#L564. At which point you can then call navigate.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4946/
Test PASSed.

kadams54:
I gave it a whirl in `connectedChange`, but that event doesn't seem to fire when you just go into a disconnected state. Consequently the logic is back in `switchEnv` right now, but it's a pretty contained chunk of code and easy to move around, should we find a better home for it.

kadams54:
## QA

OK, so the real QA is beyond me as it requires deploying in a gisf env. That said, here's what I've been doing locally to sort of fake things out:

1) Update `development.ini` to set `jujugui.gisf` to true and point to staging for the `jem_url`.
2) `make run`
3) Hit the user profile page, then click "Create new".
4) You should end up at `/u/{username}/sandbox` with a blank canvas.
5) All subsequent nav and actions should happen under that new URL.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4947/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4948/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4966/
Test PASSed.

PR #1610, opened 2016-05-04
Disconnected flag

### LINKED ISSUE #1569: Create a deployment flow flag
There are multiple situations where the user will enter the deployment flow which requires the env to be disconnected from a model but we do not want the application to log us out. At the moment we rely on flags like `gisf` and the existence of a `jem`. These may not always be there but the deployment flow use is still valid. We need a flag which is to be set when the user enters the deployment flow which starts when the user enters the disconnected empty canvas state and is unset whenever they leave this flow. They can leave the flow either by completing the deployment flow or by switching to another model at any point during the flow progression.

### LINKED ISSUE #1577: Hide the "Create new" button for new models when disconnected/in deployment flow
Depends on https://github.com/juju/juju-gui/issues/1569 being fixed. Once a flag is available, hide the create new button when the flag is set. Otherwise trying to create a new model just results on an "already disconnected" error in the console.

huwshimi:
Set a flag for when the GUI is connected to a model. Fixes #1569. Fixes #1577.


huwshimi:
QA:
- deploy with a real juju and jem
- open the profile while in the disconnected state
- you should see a "Create new" button
- click it
- open the profile
- you should not see a "Create new" button

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4963/
Test PASSed.

hatched:
Thanks, this is an elegant solution. However I'm not entirely sure it's necessary. In theory we should be able to check `app.env.get('connected')` to know if we're connected or not. In some work that @kadams54 is doing right now he's saying that's not the case, but once that is fixed we should be able to rely on that attribute. So lets hold off here until the aforementioned branch has landed.

kadams54:
FWIW, what I've found is that app.env.get('connected') is set properly, it's just the connectedChanged event isn't fired when that attribute is set to false. I have no idea why this happens and it's so contrary to what *should* happen that I don't really trust myself. It would be great if someone else could confirm what I'm seeing so we all know it's not just me.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4973/
Test PASSed.

huwshimi:
This has been updated to use `env.get('connected')`.

hatched:
:+1: Looks good thanks!

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/4974/
Test PASSed.

jujugui/static/gui/src/app/components/user-profile/user-profile.js
```diff
@@ -579,14 +580,15 @@ YUI.add('user-profile', function() {
...
-              type='inline-neutral'
-              title='Create new' />
-          </div>
-        );
+        if (this.props.canCreateNew) {
+          actions = (
+            <div className="user-profile__create-new">
+              <juju.components.GenericButton
+                action={this.switchModel}
+                type='inline-neutral'
```
anthonydillon: I create new is a positive action. Should this be `inline-positive`?

anthonydillon:
Code is good. :+1: comment is code for design. The design has neutral button with orange text but I think that's wrong.

hatched:
Thanks! :shipit: We can address that class in a follow-up.

jujugui:

PR #1659, opened 2016-05-16
Load when no models

### LINKED ISSUE #1657: Unable to load the GUI with JEM if you have no models
`Uncaught TypeError: Cannot read property 'hostPorts' of undefined` This is because we try to fetch the model details to pre-populate the u/p on load. This is only necessary for gui in Juju and not for gisf. The temporary fix is to simply call the callback with empty u/p and then return. The fix in the further future is to remove that call entirely and rely on macaroons.

huwshimi:
Don't try and connect to the first model if JEM doesn't return any. Fixes #1657.

### DIFF
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -1389,8 +1389,14 @@ YUI.add('juju-gui', function(Y) {
           this.storeUser('jem', null, true);
         }
 
-        var modelData = this._pickModel(modelList);
         this.set('environmentList', modelList);
+        if (modelList.length === 0) {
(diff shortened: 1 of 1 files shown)


huwshimi:
QA:
- load with jem but no models (I had to hack in an empty array: https://github.com/juju/juju-gui/blob/develop/jujugui/static/gui/src/app/app.js#L1376)
- the GUI should load without any `hostPorts` errors

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5063/
Test PASSed.

hatched:
Thanks! :shipit:

jujugui:

PR #1675, opened 2016-05-17
Add the ability to select region in the final step of the deployment flow

### LINKED ISSUE #1629: Deployment summary region does nothing
Setting the region to not use the default on the deployment summary does nothing. From the code it looks like it is not hooked up.

hatched:
While this adds the ability to modify the region you're deploying to the functionality is currently disabled because of missing JEM functionality. Fixes #1629

### DIFF
--- a/jujugui/static/gui/src/app/components/deployment/panel/content/content.js
+++ b/jujugui/static/gui/src/app/components/deployment/panel/content/content.js
@@ -27,15 +27,16 @@ YUI.add('deployment-panel-content', function() {
     },
 
     render: function() {
+      var props = this.props;
       return (
-        <div className="deployment-panel__content">
+        <div className={'deployment-panel__content ' + props.className || ''}>
           <div className="twelve-col">
             <div className="inner-wrapper">
               <div className="deployment-panel__content-inner">
                 <h2 className="deployment-panel__title">
-                  {this.props.title}
+                  {props.title}
                 </h2>
-                {this.props.children}
+                {props.children}
               </div>
             </div>
           </div>
--- a/jujugui/static/gui/src/app/components/deployment/panel/content/content.js
+++ b/jujugui/static/gui/src/app/components/deployment/panel/content/content.js
@@ -27,17 +27,15 @@ YUI.add('deployment-panel-content', function() {
     },
 
     render: function() {
-      var props = this.props;
-      var className = props.className || '';
       return (
-        <div className={'deployment-panel__content ' + className}>
+        <div className="deployment-panel__content">
           <div className="twelve-col">
             <div className="inner-wrapper">
               <div className="deployment-panel__content-inner">
                 <h2 className="deployment-panel__title">
-                  {props.title}
+                  {this.props.title}
                 </h2>
-                {props.children}
+                {this.props.children}
               </div>
             </div>
           </div>
--- a/jujugui/static/gui/src/app/components/deployment/panel/content/content.js
+++ b/jujugui/static/gui/src/app/components/deployment/panel/content/content.js
@@ -28,8 +28,9 @@ YUI.add('deployment-panel-content', function() {
 
     render: function() {
       var props = this.props;
+      var className = props.className || '';
       return (
-        <div className={'deployment-panel__content ' + props.className || ''}>
+        <div className={'deployment-panel__content ' + className}>
           <div className="twelve-col">
             <div className="inner-wrapper">
               <div className="deployment-panel__content-inner">
--- a/jujugui/static/gui/src/app/components/deployment/summary/summary.js
+++ b/jujugui/static/gui/src/app/components/deployment/summary/summary.js
@@ -43,6 +43,55 @@ YUI.add('deployment-summary', function() {
       validateForm: React.PropTypes.func.isRequired
     },
 
+    getInitialState: function() {
+      return {
+        regions: [],
+        activeRegion: this.props.deploymentStorage.region
+      };
+    },
+
+    componentWillMount: function() {
+      // If the model has already been deployed then we already
+      // have all the information we need.
+      var props = this.props;
+      if (props.modelCommitted) { return; }
+      // Because the user may have selected a template from the choose-cloud
+      // component we won't have any of the location data for the template
+      // to be able to properly select the correct region or cloud from the
+      // drop down list.
+      var deploymentStorage = props.deploymentStorage;
+      var templateName = deploymentStorage.templateName;
+      props.jem.listTemplates((error, credentials) => {
+        if (error) {
+          console.error('Unable to list templates', error);
+          return;
+        }
+        // The credentials are returned in an array so we need to loop through
+        // the list to find the one with the proper path.
+        credentials.forEach(credential => {
+          if (credential.path === templateName) {
+            deploymentStorage.cloud = credential.location.cloud;
+            deploymentStorage.region = credential.location.region;
+          }
+        });
+        if (deploymentStorage.cloud && deploymentStorage.region) {
+          props.jem.listRegions(deploymentStorage.cloud, (error, regions) => {
+            if (error) {
+              console.error('Unable to list regions', error);
+              return;
+            }
+            this.setState({
+              regions: regions,
+              activeRegion: deploymentStorage.region
+            });
+          });
+        } else {
+          console.log(
+            'No matching credential found', templateName, credentials);
+        }
+      });
+    },
+
     /**
       Holds the onLogin event handler which is attached when
       creating a new model.
@@ -122,10 +171,7 @@ YUI.add('deployment-summary', function() {
         return;
       }
       // Validate both fields if we are creating a new model
-      if (!this.props.validateForm([
-        'modelName',
-        'templateRegion'
-      ], this.refs)) { return; }
+      if (!this.props.validateForm(['modelName'], this.refs)) { return; }
 
       var deploymentStorage = this.props.deploymentStorage;
       this.props.jem.newModel(
@@ -134,7 +180,7 @@ YUI.add('deployment-summary', function() {
         deploymentStorage.templateName,
         {
           cloud: deploymentStorage.cloud,
-          region: deploymentStorage.region
+          region: this.state.activeRegion
         },
         null, // Controller, using the location argument instead.
         (error, data) => {
@@ -185,6 +231,20 @@ YUI.add('deployment-summary', function() {
       });
     },
 
+    /*
+      Stores the region in state on select change.
+      @method _storeRegion
+      @param {Object} e The change event.
+    */
+    _storeRegion: function(e) {
+      var selectRegion = this.refs.selectRegion;
+      var region = selectRegion.options[selectRegion.selectedIndex].value;
+      if (region) {
+        this.props.deploymentStorage.region = region;
+        this.setState({activeRegion: region});
+      }
+    },
+
     /**
       Generate the list of change items.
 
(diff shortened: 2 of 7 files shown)


hatched:
#### To QA

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5101/
Test PASSed.

kadams54:
QA is :ok:.

jujugui/static/gui/src/app/components/deployment/panel/content/content.js
```diff
@@ -27,15 +27,17 @@ YUI.add('deployment-panel-content', function() {
     },
 
     render: function() {
+      var props = this.props;
+      var className = props.className || '';
```
huwshimi: You could make the classNames util to make this all a bit cleaner.

jujugui/static/gui/src/app/components/deployment/summary/summary.js
```diff
@@ -43,6 +43,55 @@ YUI.add('deployment-summary', function() {
...
+      // have all the information we need.
+      var props = this.props;
+      if (props.modelCommitted) { return; }
+      // Because the user may have selected a template from the choose-cloud
+      // component we won't have any of the location data for the template
+      // to be able to properly select the correct region or cloud from the
+      // drop down list.
+      var deploymentStorage = props.deploymentStorage;
+      var templateName = deploymentStorage.templateName;
+      props.jem.listTemplates((error, credentials) => {
```
huwshimi: Can you please store the XHR object returned from this and cancel the request if the component is unmounted before it completes?

jujugui/static/gui/src/app/components/deployment/summary/summary.js
```diff
@@ -43,6 +43,55 @@ YUI.add('deployment-summary', function() {
...
+        // The credentials are returned in an array so we need to loop through
+        // the list to find the one with the proper path.
+        credentials.forEach(credential => {
+          if (credential.path === templateName) {
+            deploymentStorage.cloud = credential.location.cloud;
+            deploymentStorage.region = credential.location.region;
+          }
+        });
+        if (deploymentStorage.cloud && deploymentStorage.region) {
+          props.jem.listRegions(deploymentStorage.cloud, (error, regions) => {
```
huwshimi: As above, can you store and abort this request when the component is unmounted.

jujugui/static/gui/src/app/components/deployment/summary/summary.js
```diff
@@ -43,6 +43,55 @@ YUI.add('deployment-summary', function() {
       validateForm: React.PropTypes.func.isRequired
     },
 
+    getInitialState: function() {
+      return {
+        regions: [],
+        activeRegion: this.props.deploymentStorage.region
```
huwshimi: I think this will break when you try and commit to an existing model. In that case we don't need the region info.

jujugui/static/gui/src/app/components/deployment/summary/summary.js
```diff
@@ -43,6 +43,55 @@ YUI.add('deployment-summary', function() {
...
+      };
+    },
+
+    componentWillMount: function() {
+      // If the model has already been deployed then we already
+      // have all the information we need.
+      var props = this.props;
+      if (props.modelCommitted) { return; }
```
huwshimi: This might need a comment to remind us why we don't need to get the region info in this case.

jujugui/static/gui/src/app/components/deployment/summary/summary.js
```diff
@@ -43,6 +43,55 @@ YUI.add('deployment-summary', function() {
...
+      var deploymentStorage = props.deploymentStorage;
+      var templateName = deploymentStorage.templateName;
+      props.jem.listTemplates((error, credentials) => {
+        if (error) {
+          console.error('Unable to list templates', error);
+          return;
+        }
+        // The credentials are returned in an array so we need to loop through
+        // the list to find the one with the proper path.
+        credentials.forEach(credential => {
```
huwshimi: You could use `credentials.some(...` here to break the loop when it finds the matching template name.

jujugui/static/gui/src/app/components/deployment/summary/summary.js
```diff
@@ -43,6 +43,55 @@ YUI.add('deployment-summary', function() {
...
+              console.error('Unable to list regions', error);
+              return;
+            }
+            this.setState({
+              regions: regions,
+              activeRegion: deploymentStorage.region
+            });
+          });
+        } else {
+          console.log(
```
huwshimi: Maybe `console.error`. We should probably even remove the ability to set the region if we don't have this info.

jujugui/static/gui/src/app/components/deployment/summary/summary.js
```diff
@@ -122,10 +171,7 @@ YUI.add('deployment-summary', function() {
         return;
       }
       // Validate both fields if we are creating a new model
-      if (!this.props.validateForm([
-        'modelName',
-        'templateRegion'
-      ], this.refs)) { return; }
+      if (!this.props.validateForm(['modelName'], this.refs)) { return; }
```
huwshimi: Do we not need to validate that one of the options was selected (I couldn't QA this to see what exactly happens here).

jujugui/static/gui/src/app/components/deployment/summary/summary.js
```diff
@@ -348,6 +437,7 @@ YUI.add('deployment-summary', function() {
       return (
         <div className="deployment-panel__child">
           <juju.components.DeploymentPanelContent
+            className="deployment-summary"
```
huwshimi: You could just add this class to the above wrapping div rather than adding this extra className functionality to the DeploymentPanelContent component. Might make that component slightly less convoluted in its purpose.

jujugui/static/gui/src/app/components/deployment/panel/content/content.js
huwshimi: Or in fact maybe not need this it all if you just apply a wrapping class to the component this is used with (see my other component where this is used).

huwshimi:
Thanks for this @hatched a few little comments to take a look at.

bac:
@hatched Your QA instructions are beautifully minimalistic.

jujugui/static/gui/src/app/components/deployment/summary/test-summary.js
```diff
@@ -71,11 +71,26 @@ describe('DeploymentSummary', function() {
...
+            cb(null, [{
+              path: 'spinach/my-creds',
+              location: { cloud: 'aws', region: 'us-west-1' }
+            }]);
+          },
+          listRegions: (cloud, cb) => {
+            assert.equal(cloud, 'aws');
+            cb(null, ['us-west-1', 'us-east-1']);
+          }
+        }}
```
kadams54: We've got JS nested in JSX nested in JS here. If it's much more than a simple expression, I'd rather see it setup outside the JSX, assigned to a variable, and that keeps the JSX more presentation-focused and HTML-like.

kadams54:
@bac I feel it's a very Dada-esque commentary on QA notes.

jujugui/static/gui/src/app/components/deployment/summary/summary.js
hatched: good call, moved!

jujugui/static/gui/src/app/components/deployment/summary/test-summary.js
hatched: done

jujugui/static/gui/src/app/components/deployment/summary/summary.js
hatched: good find, done

jujugui/static/gui/src/app/components/deployment/panel/content/content.js
hatched: good call, removing this functionality.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5118/
Test PASSed.

jujugui/static/gui/src/app/components/deployment/summary/summary.js
hatched: good idea, added

jujugui/static/gui/src/app/components/deployment/summary/summary.js
hatched: Well if we do not have a matching region to the selected cred then we would need the user to choose one before continuing. So I'll have to expand the logic around the form submission validation as you mentioned elsewhere.

jujugui/static/gui/src/app/components/deployment/summary/summary.js
hatched: Line 54/55 ;)

jujugui/static/gui/src/app/components/deployment/summary/summary.js
hatched: See above comment - I'll have to expand this validation.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5121/
Test PASSed.

jujugui/static/gui/src/app/components/deployment/summary/summary.js
hatched: I don't think that's the case but I'll check in qa just to be sure. If anything 'activeRegion' will simply be undefined

kadams54:
QA is still :ok:. Code looks good, :+1: - thanks!

jujugui/static/gui/src/app/components/deployment/summary/summary.js
```diff
@@ -43,6 +43,59 @@ YUI.add('deployment-summary', function() {
...
+          if (credential.path === templateName && credential.location) {
+            deploymentStorage.cloud = credential.location.cloud;
+            deploymentStorage.region = credential.location.region;
+            return true;
+          }
+        });
+        if (deploymentStorage.cloud && deploymentStorage.region) {
+          this._listRegionsXHR = props.jem.listRegions(
+            deploymentStorage.cloud, (error, regions) => {
+              this._abortXHR('_listRegionsXHR');
```
huwshimi: Do you really need to abort an XHR request once it is complete?

huwshimi: If you don't need to abort these requests individually then you could just store all the XHRs in an array instead of separately named variables.

huwshimi:
:+1: with one question.

jujugui/static/gui/src/app/components/deployment/summary/summary.js
hatched: Technically we don't need to abort it, but this just cleans up after the request by also nulling out the reference to it.
I wanted to keep the requests separate for now so that we can track them individually in the future if needed. I don't actually like that we're making these requests her at all. We should be requesting this from the DB and have the DB fetch them for us. But that is in the pipeline for sometime in the future.

huwshimi: Yep, I completely agree, for example, we currently request things like credentials/templates in several different components, when really we could just be requesting them once and passing the data down into the components.

Not a big deal, but I probably wouldn't bother with that cleanup, but if it makes you happy then leave it :).

hatched:
Because of the changes coming with the DF design I'm landing this and we can do the UI updates in upcoming branches. :shipit: 
Thanks for the reviews and qa's

jujugui:

PR #1681, opened 2016-05-19
Disconnect from the model when loading the profile.

huwshimi:


huwshimi:
QA:
- load the gui
- click on the username in the breadcrumb or click "More" in the model switcher
- the gui should disconnect from the current model and then load the profile

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5113/
Test PASSed.

jujugui/static/gui/src/app/components/header-breadcrumb/header-breadcrumb.js
```diff
@@ -60,17 +60,22 @@ YUI.add('header-breadcrumb', function() {
...
-      this.props.changeState({
-        sectionB: {
-          component: 'profile',
-          metadata: null
-        }
+    _displayProfile: function(e) {
+      if (e) {
+        e.preventDefault();
+      }
+      this.props.switchModel(null, null, null, () => {
```
kadams54: This `null, null, null` makes me sad. Que sera, sera.

kadams54:
In QA: when I click on the profile link, the "Connecting to the Juju environment" spinner comes up and then nothing happens. No errors in the console, nothing. I'm QAing in GIJoe on AWS, in Chrome/Ubuntu.

huwshimi:
Hi @kadams54 I had this same problem with GIJ the other day, would you mind QAing this with the charm to see if this branch is working OK separate from that existing issue?

hatched:
There is a qa issue but not the one mentioned by @kadams54 ( sorry :) ) 

The websocket gets properly terminated when viewing the profile but afterwards it's impossible to re-connect to any model. This is because it tries to close an already closed websocket here https://github.com/juju/juju-gui/blob/develop/jujugui/static/gui/src/app/app.js#L2020 and fails so it can't ever get to the reconnect stage.

huwshimi:
@hatched I've tried cleaning up the websocked when the env disconnects, but I'm not completely sure I've fixed it correctly as there is now a failing test that I'm not sure can be resolved. Maybe you have some idea about how we should fix this:

https://github.com/juju/juju-gui/blob/develop/jujugui/static/gui/src/test/test_app.js#L920

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/5143/
Test FAILed.

hatched:
@huwshimi I'm not sure we want to null it out, but instead check to see if it's connected before trying to send messages down it. So that conditional could become something like:
(psudo code)
`if (this.env.ws && this.env.ws.connected) {`

huwshimi:
@hatched hmm... I can't seem to get it to work when I check the connected status, like it worked when I nulled the ws.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5159/
Test PASSed.

huwshimi:
@hatched this branch has now been rebased on #1725 so it might be ready for another QA.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5162/
Test PASSed.

hatched:
Turns out we can't actually do this because when using a controller we make the listModels requests over the active websocket. Lets discuss a new approach for this.

huwshimi:
Closing as this won't work in the controller case.

PR #1687, opened 2016-05-23
Set the focus on panels when they open

### LINKED ISSUE #1524: Focus on the pages with large vertical content
Pages with a lot of vertical content like the charmbrowser, inspector config, etc are not able to be navigated with the keyboard because they aren't focused when rendered. It would be nice if we auto focused on render so that the user could use the arrow/tab/page keys to move around.

huwshimi:
When panels are displayed then set the keyboard focus on the content so they can be navigated with the keyboard. Fixes #1524.

### DIFF
--- a/jujugui/static/gui/src/app/components/entity-details/entity-details.js
+++ b/jujugui/static/gui/src/app/components/entity-details/entity-details.js
@@ -149,6 +149,9 @@ YUI.add('entity-details', function() {
     },
 
     componentDidMount: function() {
+      // Set the keyboard focus on the component so it can be scrolled with the
+      // keyboard. Requires tabIndex to be set on the element.
+      this.refs.content.focus();
       this.detailsXhr = this.props.getEntity(
           this.props.id, this.fetchCallback);
     },
@@ -202,10 +205,12 @@ YUI.add('entity-details', function() {
 
     render: function() {
       return (
-        <div className={this._generateClasses()}>
+        <div className={this._generateClasses()}
+          ref="content"
+          tabIndex="0">
           {this._generateContent()}
         </div>
-      );;
+      );
     }
   });
 
--- a/jujugui/static/gui/src/app/components/panel/panel.js
+++ b/jujugui/static/gui/src/app/components/panel/panel.js
@@ -22,6 +22,12 @@ YUI.add('panel-component', function() {
 
   juju.components.Panel = React.createClass({
 
+    componentDidMount: function() {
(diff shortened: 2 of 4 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5125/
Test PASSed.

huwshimi:
QA:
- open the profile or the account page
- use the keyboard to scroll up and down (you might have to adjust your browser size to make the content scrollable)
- open a charm or bundle details
- again you should be able to use the keyboard to scroll up and down

anthonydillon:
Code :ok: and QA :+1:

jujugui/static/gui/src/app/components/entity-details/entity-details.js
```diff
@@ -202,10 +205,12 @@ YUI.add('entity-details', function() {
 
     render: function() {
       return (
-        <div className={this._generateClasses()}>
+        <div className={this._generateClasses()}
+          ref="content"
+          tabIndex="0">
```
hatched: Does this mess with any of the accessibility work we've done for the breadcrumb?

jujugui/static/gui/src/app/components/panel/panel.js
```diff
@@ -68,7 +74,9 @@ YUI.add('panel-component', function() {
     render: function() {
       return (
         <div className={this._genClasses()}
-          onClick={this._handleClick}>
+          onClick={this._handleClick}
+          ref="content"
+          tabIndex="0">
```
hatched: same as above

hatched:
:+1: With one q

jujugui/static/gui/src/app/components/entity-details/entity-details.js
huwshimi: I don't believe so, all it's doing is setting the focus when the panel or entity details open. I don't think this can impact on the breadcrumbs at all, but I may have missed something. What do you think the issue might be?

hatched: I was just thinking that it may pick up the tab instead of the breadcrumb first.

huwshimi: @hatched Ah, no setting the tabIndex just allows it to be tabbed to. You can set the order, but we don't.

huwshimi:
Thanks @anthonydillon, @hatched. :shipit:

jujugui:

PR #1708, opened 2016-05-27
Updating node deps and removing d3 build

hatched:
This branch updates all of the node dependencies to their latest counterparts including React to 15.1. This also uses a pre-built version of d3 and saves it as an asset so that we do not need to build it every time. This means that to update d3 we will have to build it again using this project https://github.com/juju/juju-gui-d3 and then update the asset manually. This is necessary to allow us to update to the latest node/npm in a follow-up branch.


hatched:
#### To QA

make clean-all deps fast-babel test-js test-js-old lint-js run

That will run through all of the various tests and then start the GUI server. Play around with the GUI and make sure it works as expected.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5141/
Test PASSed.

huwshimi:
Looks great, thanks for this Jeff. :+1: QA OK. I think you can probably remove this now too: https://github.com/juju/juju-gui/blob/develop/jujugui/static/gui/src/app/utils/component-test-utils.js#L44

jaycee:
:+1: looks good; I'm happy to see all that Makefile cruft get cleared out.

hatched:
Thanks! :shipit:

jujugui:

PR #1728, opened 2016-05-31
Close ws profile

### LINKED ISSUE #1527: Console error when switching model from profile page
WebSocket is already in CLOSING or CLOSED state.
The switching works, but that error is printed.

hatched:
Any rpc requests which are made during a closing or closed websocket will now be dropped. Fixes #1527

### DIFF
--- a/jujugui/static/gui/src/app/assets/javascripts/reconnecting-websocket.js
+++ b/jujugui/static/gui/src/app/assets/javascripts/reconnecting-websocket.js
@@ -148,6 +148,9 @@ function ReconnectingWebSocket(url, protocols) {
             self.reconnect = false;
             if (close) {
                 ws.close();
+                // When we call to close the websocket propagate the CLOSING
+                // state up to the reconnecting websocket instance.
+                self.readyState = WebSocket.CLOSING;
             }
 
         }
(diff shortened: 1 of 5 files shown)


hatched:
#### To QA

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5166/
Test PASSed.

huwshimi:
@hatched this seems to work fine, no errors in sandbox, deployed with charm, GUI in Juju on AWS and GUI in Juju on LXD, except that with GIJ I only ever got a "default" model so I couldn't test the model switching.

:+1:

huwshimi:
Additionally, when I open the profile (in GIJ) and click 'Create new' I get `Websocket is not open, dropping request. readyState: 2 undefined` in the console.

frankban:
Code :+1: and QA ok, with the consideration below.
I am ok with avoiding sending requests to a closing ws connection. But here a proper fix would be not sending them on the first place. The profile page should wait for the WebSocket to be properly reconnected before re-dispatching the application. Could you please add a TODO if not there?

hatched:
@frankban after investigating a fix to make the switch only when the websocket has been connected. It actually has to wait until the user has logged into the websocket, not just having it be connected. By simply navigating once the websocket has connected still introduces a race condition where it tries to send the incorrect facade because it hasn't yet recieved the correct responses.

A proper fix to this entire stack is likely to implement these calls behind a database which could offer up what it believes to be a recent version and update it in the background. This will need to be done in follow-up branches over time. 

jujugui:

PR #1732, opened 2016-06-01
Implement the "make uitest" target.

frankban:
This is used to run uitest powered Juju GUI functional tests.

QA: make uitest.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,8 @@ STATIC_CSS_FILES = \
 LSB_RELEASE = $(shell lsb_release -cs)
 
 # libfontconfig1 is required by phantom.
-SYSDEPS = g++ git inotify-tools libfontconfig1 nodejs python-virtualenv
+SYSDEPS = coreutils g++ git inotify-tools libfontconfig1 nodejs \
+	python-virtualenv
 
 .PHONY: help
 help:
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,8 @@ help:
 	@echo "test-js - run newer js tests in terminal; primarily for CI build"
 	@echo "test-js-old - run older js tests that have transitioned to karma in the terminal"
 	@echo "update-downloadcache - update the download cache"
-	@echo "uitest - run functional tests"
+	@echo "uitest - run functional tests;"
+	@echo "  use the ARGS environment variable to append/override uitest arguments"
 
 
 #########
@@ -374,4 +375,4 @@ uitest: sudo dist
 .PHONY: sudo
 sudo:
 	@# Pre-cache sudo permissions, so that the process is not stopped later.
-	@sudo true
+	@sudo -v
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ LSB_RELEASE = $(shell lsb_release -cs)
 
 # libfontconfig1 is required by phantom.
 SYSDEPS = coreutils g++ git inotify-tools libfontconfig1 nodejs \
-	python-virtualenv
+	python-virtualenv realpath
 
 .PHONY: help
 help:
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ STATIC_CSS_FILES = \
 LSB_RELEASE = $(shell lsb_release -cs)
 
 # libfontconfig1 is required by phantom.
-SYSDEPS = nodejs python-virtualenv g++ inotify-tools libfontconfig1
+SYSDEPS = g++ git inotify-tools libfontconfig1 nodejs python-virtualenv
 
 .PHONY: help
 help:
@@ -74,6 +74,7 @@ help:
 	@echo "test-js - run newer js tests in terminal; primarily for CI build"
 	@echo "test-js-old - run older js tests that have transitioned to karma in the terminal"
 	@echo "update-downloadcache - update the download cache"
+	@echo "uitest - run functional tests"
 
 
 #########
@@ -359,5 +360,18 @@ clean-gui:
 	- rm -rf jujugui/static/gui/build
 
 .PHONY: clean-all
-clean-all: clean-venv clean-pyc clean-gui clean-dist
+clean-all: clean-venv clean-pyc clean-gui clean-dist clean-uitest
 	- rm -rf *.egg-info
+
+.PHONY: clean-uitest
+clean-uitest:
+	- rm -rf uitest
+
+.PHONY: uitest
+uitest: sudo dist
+	@scripts/uitest.sh $(ARGS)
+
+.PHONY: sudo
+sudo:
+	@# Pre-cache sudo permissions, so that the process is not stopped later.
+	@sudo true
--- a/docs/browser-testing.rst
+++ b/docs/browser-testing.rst
@@ -4,10 +4,14 @@
 Browser Testing
 ===============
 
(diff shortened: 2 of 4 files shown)


scripts/uitest.sh
```diff
@@ -0,0 +1,54 @@
...
+if [ ! -d "$uitest_dir" ]; then
+  git clone git@github.com:CanonicalLtd/juju-uitest.git $uitest_dir
+fi
+
+# Pull to get latest uitests.
+pushd $uitest_dir
+git checkout master
+git pull origin master
+popd
+
```
bac: Why all the pushing and popping since it is the same dir?

Makefile
```diff
@@ -359,5 +360,18 @@ clean-gui:
...
+	- rm -rf uitest
+
+.PHONY: uitest
+uitest: sudo dist
+	@scripts/uitest.sh $(ARGS)
+
+.PHONY: sudo
+sudo:
+	@# Pre-cache sudo permissions, so that the process is not stopped later.
+	@sudo true
```
bac: Nice

Makefile
```diff
@@ -359,5 +360,18 @@ clean-gui:
...
+clean-all: clean-venv clean-pyc clean-gui clean-dist clean-uitest
 	- rm -rf *.egg-info
+
+.PHONY: clean-uitest
+clean-uitest:
+	- rm -rf uitest
+
+.PHONY: uitest
+uitest: sudo dist
+	@scripts/uitest.sh $(ARGS)
```
hatched: If we're going to comment out the actual command I'd prefer a user friendly echo.

docs/browser-testing.rst
```diff
@@ -18,144 +22,48 @@ Building
...
 
+Running the browser tests
+=========================
 
-Manually running the browser tests
-===================================
+To start the Juju GUI related subset of the uitest suite, run the following::
 
-Note that there are four valid browser targets currently.
+  make uitest
```
bac: It would be nice to document the use of ARGS.

docs/browser-testing.rst
```diff
@@ -18,144 +22,48 @@ Building
...
-address of your home network.
+Running Juju GUI functional tests requires:
 
-You can also setup things through external server with handy tactics such as
-ssh tunnels, but direct exposure is noted below as the simplest use case.
+- some time to complete its execution: do not expect it to be blazing fast;
+- SSH keys already generated for the user running the tests;
+- a working Internet connection.
 
+Providing customized uitest options
```
hatched: Could you also add here how to run a single test or suite?

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5168/
Test PASSed.

scripts/uitest.sh
```diff
@@ -0,0 +1,54 @@
...
+    shift
+    ;;
+    --gui-archive)
+    archive="$2"
+    shift
+    ;;
+    *)
+    args="$args $key"
+    ;;
+esac
```
bac: Add -h|--help

scripts/uitest.sh
```diff
@@ -0,0 +1,54 @@
...
+
+# Define paths to uitest directories and command.
+script_dir=`dirname $0`
+root_dir=`realpath $script_dir/../`
+uitest_dir=$root_dir/uitest
+uitest=$uitest_dir/devenv/bin/uitest
+
+# Define default parameters.
+cloud="lxd"
+archive=`ls $root_dir/dist/jujugui-*.tar.bz2`
```
hatched: Should we have a check here for multiple returned values? I've frequently had multiple versions in the dist folder and it'll likely become more likely as we speed up the release process.

bac:
:+1: with a few suggestions.  Thanks!

scripts/uitest.sh
bac: No because 'dist' does a 'clean-dist'.  (I just checked because I had the same worry.)

scripts/uitest.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+
+set -e
+
+# Define paths to uitest directories and command.
+script_dir=`dirname $0`
+root_dir=`realpath $script_dir/../`
+uitest_dir=$root_dir/uitest
```
hatched: I'll likely have a version of uitest already checked out which may or may not have modifications I'd like to use. Should this be configurable? Or maybe instead be `$root_dir/../`

Also I'd prefer constants to be uppercase :) but up to you.

scripts/uitest.sh
hatched: Ahh I typically use `fast-dist` so that's probably why.

Makefile
jrwren: `sudo -v` could work here to pre-cache permissions and update their timestamp while also not running a command.

scripts/uitest.sh
frankban: Done.

Makefile
frankban: Nice, sounds good!

docs/browser-testing.rst
frankban: It is done below, isn't it?

frankban: Anyway, I added a description for ARGS also in "make help".

Makefile
frankban: As discussed, the script already prints out a similar message.

scripts/uitest.sh
frankban: --help override all other options allrady, so I consider it ok to just propagate that without any special handling.

docs/browser-testing.rst
frankban: Done.

scripts/uitest.sh
frankban: I think this is out of scope for the feature in the GUI. If you already have uitests elsewhere, why not using that directly? I'd like to keep things simple enough here so that when someone reports that "make uitest" failed, we are sure nothing too much exotic is going on (hence the "make dist", the pulling of latest uitest master etc).
I am not sure the result of dirname can be considered a constant.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5169/
Test PASSed.

hatched:
QA Issue:

After the gui dist was created: 
`scripts/uitest.sh: line 7: realpath: command not found`

frankban:
Ah! Really? So coreutils are not installed by default. Cool, fixed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5175/
Test PASSed.

hatched:
@frankban I see you added the `coreutils` package, which I already had installed and still don't have `realpath` Is it possible that coreutils on Trusty doesn't have realpath and instead needs to be installed as 'realpath' which appears to be available `apt-cache show realpath`

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5186/
Test PASSed.

hatched:
I'm running into issues once it starts the ui test where it just hangs. It doesn't appear to be trying to build Juju

frankban:
Merging this, will fix possible issues later.
Thanks for the reviews and QA attempts!
:shipit:

jujugui:

PR #1742, opened 2016-06-07
Profile uncommitted notification

### LINKED ISSUE #1729: Uncommitted changes are lost if you navigate to the profile
If you navigate to the profile when you have uncommitted changes you will lose the changes as there is no way to navigate back to your current model.

We can't just disconnect when you navigate back to the model as the controller case needs an active connection to list models etc.

We could potentially show the uncommitted changes popup when the user clicks on the profile (we'd just need to implement that for every profile nav link).

huwshimi:
Display a notification if you switch to the profile when there are uncommitted changes. Fixes #1729.

### DIFF
--- a/jujugui/static/gui/src/app/components/env-list/env-list.js
+++ b/jujugui/static/gui/src/app/components/env-list/env-list.js
@@ -24,7 +24,8 @@ YUI.add('env-list', function() {
 
     propTypes: {
       envs: React.PropTypes.array,
-      handleEnvClick: React.PropTypes.func
+      handleEnvClick: React.PropTypes.func,
+      showProfile: React.PropTypes.func.isRequired
     },
 
     getInitialState: function() {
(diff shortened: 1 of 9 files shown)


huwshimi:
QA:
- load the gui with JEM
- open the profile
- the profile should load
- click Create new
- add a machine or service
- click on your username in the breadcrumb
- you should get a confirmation
- click Continue
- you should get your profile
- click 'Create new' (you may have to reload the GUI if using it locally)
- add a machine or service
- open the env switcher
- click Profile
- you should get a confirmation

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5185/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5193/
Test PASSed.

PR #1769, opened 2016-06-16
mega-watcher handlers: always store the workload status message.

frankban:


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5252/
Test PASSed.

frankban:
Here is a command that can be useful while QAin or developing based on this new exposed info:
`juju run --unit wordpress/0 'status-set active "foo bar"'`

anthonydillon:
Code and QA :+1: using the juju run command in the QA tip. Could not test a true pending message due to the failure between beta8 and the latest GUI.

One small question, why is the new field camelcase when all other attrs are lowercase with underscores? Would it be better for consistency to be `workload_status_message`?

QA steps taken:
- Deployed this GUI to a new LXD controller
- Checked the juju-gui application using `app.db.units.getById('juju-gui/0')` in the console
- workloadStatusMessage was empty string
- Ran `juju run --unit juju-gui/0 'status-set active "foo bar"'` and ran the above again in the console
- workloadStatusMessage was 'foo bar'

hatched:
@anthonydillon the underscores are legacy from way back that haven't yet been updated.

frankban:
Thanks for the reviews!
:shipit:

jujugui:

PR #1791, opened 2016-06-23
Convert Juju2 API to lower-case dash-separated fields.

frankban:
This branch is huge but mostly mechanical.
It includes:
- converting all API fields to lower-case-dash-separated, excluding the internal response of CharmInfo, still not implemented in core;
- removing support to legacy API and facades from api.js, resulting in lots of code simplifications;
- removing support for token login in the new non-legacy API: it's only used by quickstart, which is deprecated in Juju 2;
- removing support to legacy bundle changes (making use of the GUI server) in the new non-legacy API: in Juju 2 controllers are always capable to return bundle changes natively;
- removing supoort for GUI server Deployer facade, only used in Juju1 as Juju2 has native bundle deployment supoort and the deployment from the GUI is managed by ECS.

Sandbox mode now temporarily use the legacy API file so that its tests still pass and we can defer updating it. This will be changed in follow up branches, this one is already huge.

QA is impossible.

### DIFF
--- a/jujugui/static/gui/src/app/store/env/api.js
+++ b/jujugui/static/gui/src/app/store/env/api.js
@@ -30,9 +30,8 @@ YUI.add('juju-env-api', function(Y) {
   // Define the pinger interval in seconds.
   var PING_INTERVAL = 10;
 
-  // Define the Admin API facade versions for Juju 1 and 2.
-  var ADMIN_FACADE_VERSION_JUJU1 = 0;
-  var ADMIN_FACADE_VERSION_JUJU2 = 3;
+  // Define the Admin API facade version.
+  var ADMIN_FACADE_VERSION = 3;
 
   var environments = Y.namespace('juju.environments');
   var utils = Y.namespace('juju.views.utils');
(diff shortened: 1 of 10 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5288/
Test PASSed.

jaycee:
> QA is impossible.

This may be the best line in a PR ever.

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -297,20 +277,18 @@ YUI.add('juju-env-api', function(Y) {
...
      */
     _watchAll: function() {
-      this._send_rpc(
-        {
-          Type: 'Client',
-          Request: 'WatchAll'
-        },
-        function(data) {
-          if (data.Error) {
-            console.log('aiiiiie!'); // retry and eventually alert user XXX
```
jaycee: We'll miss you, entirely inappropriate log message.

jaycee:
This looks good; I saw several TODOs about urls not working in GIJoe. What's the plan there?

:+1:

frankban:
Heh, we need to discuss the plan there, and restore support for local charms in gijoe.
Thanks for the review!
:shipit:

jujugui:

PR #1804, opened 2016-06-28
Refactor mega-watcher handler utilities.

frankban:
Specialise utilities for Juju 2 API and put the old logic in the legacy handlers, so that we don;t have to maintain the logic for both new and legacy API.

### DIFF
--- a/jujugui/static/gui/src/app/models/legacy-handlers.js
+++ b/jujugui/static/gui/src/app/models/legacy-handlers.js
@@ -65,6 +65,16 @@ YUI.add('juju-legacy-delta-handlers', function(Y) {
       @return {undefined} Nothing.
      */
     unitInfo: function(db, action, change) {
+      // Return a list of ports represented as "NUM/PROTOCOL", e.g. "80/tcp".
+      var convertOpenPorts = function(ports) {
+        if (!ports) {
+          return [];
+        }
+        return ports.map(port => {
+          return port.Number + '/' + port.Protocol;
+        });
+      };
+
       var unitData = {
         id: change.Name,
         charmUrl: change.CharmURL,
@@ -72,7 +82,7 @@ YUI.add('juju-legacy-delta-handlers', function(Y) {
         machine: change.MachineId,
         public_address: change.PublicAddress,
         private_address: change.PrivateAddress,
-        open_ports: utils.convertOpenPorts(change.Ports),
+        open_ports: convertOpenPorts(change.Ports),
         // Since less recent versions of juju-core (<= 1.20.7) do not include
         // the Subordinate field in the mega-watcher for units, the following
         // attribute could be undefined.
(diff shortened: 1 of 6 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5324/
Test PASSed.

bac:
:+1:

jujugui/static/gui/src/app/models/legacy-handlers.js
```diff
@@ -65,14 +65,24 @@ YUI.add('juju-legacy-delta-handlers', function(Y) {
...
+
       var unitData = {
         id: change.Name,
         charmUrl: change.CharmURL,
         service: change.Application || change.Service,
         machine: change.MachineId,
         public_address: change.PublicAddress,
         private_address: change.PrivateAddress,
-        open_ports: utils.convertOpenPorts(change.Ports),
+        open_ports: convertOpenPorts(change.Ports),
```
jaycee: Is this b/c the utils are now defined in the legacy files? Given that they seem to be used here as well shouldn't we just move them into a shared utils file rather than inlining them? Or am I misunderstanding this change?

jujugui/static/gui/src/app/models/legacy-handlers.js
```diff
@@ -299,9 +318,22 @@ YUI.add('juju-legacy-delta-handlers', function(Y) {
       @return {undefined} Nothing.
      */
     annotationInfo: function(db, action, change) {
+      var cleanUpEntityTags = function(tag) {
+        var result = tag.replace(/^(service|unit|machine|environment)-/, '');
```
jaycee: This looks like the second time we've done the cleanup code; this seems like another candidate for being moved to an external utils file to be shared between these two instances, unless there's a difference I'm missing?

jaycee:
Looks good, but I have 2 minor concerns about what looks like some code duplication.

:+1: with those fixed or explained to me.

Thanks!

jujugui/static/gui/src/app/models/legacy-handlers.js
frankban: So this was shared code before this change. This branch basically inlines what's legacy API specific, so that the functions in utils can only focus on new API.

jujugui/static/gui/src/app/models/legacy-handlers.js
frankban: Same as above. The two functions are slightly different, and this one is inlined because it's only intended to be use in the legacy API ghetto.

frankban:
:shipit:

jujugui:

jujugui/static/gui/src/app/models/legacy-handlers.js
jaycee: Ah, I had this flipped. Ok, thanks.

PR #1846, opened 2016-07-14
Fixes local charm drag-n-drop for sandbox mode

jaycee:


jaycee:
QA:

* Spin up the gui in sandbox.
* Drop a local charm on it.
* It should largely deploy alright--sandbox emulation of local charms is a bit error prone, so icons and the like may be wrong.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5421/
Test PASSed.

jujugui/static/gui/src/app/store/env/web-sandbox.js
```diff
@@ -29,12 +29,15 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
 YUI.add('juju-env-web-sandbox', function(Y) {
 
   var module = Y.namespace('juju.environments.web');
-  var localCharmExpression = /\/juju-core\/charms\?series=(\w+)/;
+  var localCharmExpression = new RegExp(
+      /\/juju-core\/model\/[A-Za-z0-9-]+\/charms\?/.source +
```
hatched: personally, I prefer regex to be in one long line and use the disable flag to make eslint to not complain.

hatched:
:+1: with one request

hatched:
Thanks @jaycee QA :ok:

jujugui/static/gui/src/app/store/env/web-sandbox.js
jaycee: Hm. I don't love littering our code with the disable comment, and this was actually a bit easier to read to me since it's broken up into path and querystring components.

jujugui/static/gui/src/test/test_web_sandbox.js
```diff
@@ -50,7 +50,7 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
     describe('sendPostRequest', function() {
 
       it('uses the given state to handle local charm uploads', function() {
-        var path = '/juju-core/charms?series=trusty';
+        var path = '/juju-core/model/some-uuid/charms?series=trusty';
```
kadams54: We've run into a problem before where test data wasn't accurate enough to real world data. Consequently there was a bug that a test would have exposed, had it had more realistic data. I think anytime we're testing code that parses a string into parts, that string needs to be as true to real life as possible; hence, I'd like to see a real UUID in here.

kadams54:
:+1: with a small comment on realistic test data. Thanks!

jaycee:
:shipit:

jujugui:

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5431/
Test PASSed.

PR #1849, opened 2016-07-15
Choose cloud component

huwshimi:
Split out the choose cloud section of the new deployment flow into a component.

### DIFF
--- a/jujugui/static/gui/src/app/components/deployment-flow/cloud/cloud.js
+++ b/jujugui/static/gui/src/app/components/deployment-flow/cloud/cloud.js
@@ -1,0 +1,154 @@
+/*
+This file is part of the Juju GUI, which lets users view and manage Juju
+environments within a graphical interface (https://launchpad.net/juju-gui).
+Copyright (C) 2016 Canonical Ltd.
+
+This program is free software: you can redistribute it and/or modify it under
+the terms of the GNU Affero General Public License version 3, as published by
+the Free Software Foundation.
+
+This program is distributed in the hope that it will be useful, but WITHOUT
+ANY WARRANTY; without even the implied warranties of MERCHANTABILITY,
+SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Affero
+General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License along
+with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+'use strict';
+
+YUI.add('deployment-cloud', function() {
+
+  juju.components.DeploymentCloud = React.createClass({
+    propTypes: {
+      acl: React.PropTypes.object.isRequired,
+      cloud: React.PropTypes.string,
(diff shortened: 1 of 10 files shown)


huwshimi:
QA:
- use the "df" flag
- click on a cloud
- the choose cloud section should collapse and the next sections should become enabled

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5429/
Test PASSed.

jujugui/static/gui/src/app/components/deployment-flow/cloud/cloud.js
```diff
@@ -0,0 +1,154 @@
...
+      @method _generateLogo
+      @param {String} id A cloud id.
+      @returns {Array} The logo.
+    */
+    _generateLogo: function(id) {
+      if (!id) {
+        return;
+      }
+      var cloud = this.CLOUDS[id];
+      return cloud.svgHeight ? (
```
hatched: Maybe we should be passing in the path for the icon instead of relying on a side effect of resizing an icon that hopefully is there.

jujugui/static/gui/src/app/components/deployment-flow/cloud/cloud.js
```diff
@@ -0,0 +1,154 @@
...
+        disabled: this.props.acl.isReadOnly(),
+        title: 'Change cloud',
+        type: 'neutral'
+      }];
+    },
+
+    render: function() {
+      var cloud = this.props.cloud;
+      return (
+        <juju.components.DeploymentSection
```
hatched: Because a `DeploymentSection` is a requirement of the deployment flow and not a requirement of this component maybe the `DeploymentSection` components should be wrapping the other components, in this case the `DeploymentCloud`, in the `DeploymentFlow` component.

hatched:
Thanks @huwshimi  looks great :+1: :ok: 

If you agree that the `DeploymentSection` component should wrap the children from within the `DeploymentFlow` then there will need to be some changes to how the `DeploymentSection` renders. I'll leave the final decision up to you there.

jujugui/static/gui/src/app/components/deployment-flow/cloud/cloud.js
huwshimi: That was my initial thinking too, but I think the problem then is you would be moving a lot of logic to the parent component (e.g. generation of buttons, info blocks, titles etc.)

I think it might be better to contain all of the section in a component and then any parts that you want to share outside of the deployment flow would then get included in this child component.

huwshimi: Maybe we can try it and if it's wrong we can move everything.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5436/
Test PASSed.

huwshimi:
Thanks @hatched :shipit:

jujugui:

PR #1868, opened 2016-07-21
Fix charmstore login

kadams54:


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5470/
Test PASSed.

kadams54:
## QA

1. Point your `development.ini` to jujugui.org and set `jujugui.sandbox` to false.
2. Fire up the GUI.
3. In an incognito window (to avoid problems due to stale macaroons), go to http://0.0.0.0:6543/profile and go through the login process (this is different from the charmstore login). Note that you may have to refresh the GUI, as there's currently a problem where the first login attempt fails due to an Access-Control-Allow-Origin issue. Best to have the console open at this point so you can be watching for errors. You may see 401s on budget and agreement endpoints; those are not detrimental to QA, though I'm curious why they're happening.
4. Once logged in and on the profile, click on the "Login to charmstore" button. You should go back through the SSO process. You should see no new errors show up in console and this time the login button should disappear at the end of the whole process.

jujugui/static/gui/src/app/components/user-profile/test-user-profile.js
```diff
@@ -260,6 +260,98 @@ describe('UserProfile', () => {
...
+                <p className="user-profile__empty-text">
+                  Your models, bundles and charms will appear here when the problem is resolved.
+                </p>
+              </div>
+            </div>
+          </div>
+        </div>
+      </juju.components.Panel>
+    );
+    /* eslint-disable max-len */
```
frankban: Why disable max len here?

jujugui/static/gui/src/app/components/user-profile/user-profile.js
```diff
@@ -198,7 +198,7 @@ YUI.add('user-profile', function() {
           model.owner = model.ownerTag;
           return model;
         });
-      } else {
+      } else if (data.map) {
```
frankban: Is this intended as a type check? Could `else if (data && data.length) be sufficient here?

jujugui/static/gui/src/app/components/user-profile/user-profile.js
```diff
@@ -790,7 +790,7 @@ YUI.add('user-profile', function() {
             <juju.components.Spinner />
           </div>);
       }
-      if (list.length === 0) {
+      if (!list || list.length === 0) {
```
frankban: `if (!list || !list.length)` would be more readable I guess? [shrug]

frankban:
:+1: with some changes, no QA.

jujugui/static/gui/src/app/components/user-profile/test-user-profile.js
kadams54: Copy-n-paste error. I'll correct.

jujugui/static/gui/src/app/components/user-profile/user-profile.js
kadams54: I was "meh" about how to do this. In the end, I decided to check for the function that was actually used in the subsequent block. I could go either way.

jujugui/static/gui/src/app/components/user-profile/test-user-profile.js
kadams54: OK, now that I've investigated more... it's there because quite a bit of the JSX in creating the expected response runs over the 80 char limit. The same thing is done in the existing "renders empty state" test.

frankban:
QA not ok.
When refreshing after the initial error while logging in I got:
`Warning: Failed propType: Required prop `username was not specified in `Constructor`. Check the render method of `Constructor`.` and other stuff like `cannot retrieve terms: user profile "#pii#frankban#/pii#" not found`. 

When I try to click to the "login to the charmstore" button it opens the new tab, it logs me in but then the profile page does not refresh, doesn't show my charms, doesn't change at all, and keep showing the "We are unable to load your profile at the moment` message.

bac:
I tried the QA but found it to be confusing.  I was eventually (third time) able to 'Login to the charmstore' and the button did go away.  I then got "We are unable to load your profile", which I suspect is an unrelated but unsettling issue.

There was a mess of errors that did not stop as suggested.

screen shot 2016-07-21 at 14 04 20

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5471/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5472/
Test PASSed.

bac:
After "Login to charmstore" I now get a spinner due to a different JS error.

screen shot 2016-07-21 at 15 25 36

bac:
A further attempt worked.  I'm cautiously marking this as QA-OK.

screen shot 2016-07-21 at 16 12 29

jujugui/static/gui/src/app/components/user-profile/user-profile.js
kadams54: After checking into this more, I decided to keep `data.map` in place. If `data` is an empty list, we still want the mapping to happen so that `modelList` ends up being assigned to an empty list. Just checking to see if `data` is truthy doesn't make sense, as we've already checked `data.length` by that point. So I'm sticking with the original code, which verifies that the `map` function is present before calling it.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5473/
Test PASSed.

kadams54:
Thanks all! :shipit:

jujugui:

jujugui/static/gui/src/app/components/user-profile/user-profile.js
hatched: wouldn't `Array.isArray(data)` be more appropriate?

PR #1878, opened 2016-07-26
Create Section load watcher component

hatched:
When we have a number of components that are individually making requests we want to know if all of them return no data, or return an error. This component allows you to easily show an error or empty component if all of the children return an error or empty status via the `broadcastStatus` prop which is augmented onto all of the children.

### DIFF
--- a/jujugui/static/gui/src/app/components/section-load-watcher/section-load-watcher.js
+++ b/jujugui/static/gui/src/app/components/section-load-watcher/section-load-watcher.js
@@ -23,43 +23,90 @@ YUI.add('section-load-watcher', function(Y) {
   juju.components.SectionLoadWatcher = React.createClass({
 
     propTypes: {
-      children: React.PropTypes.func,
+      // The eslint propType sorting requires that capital letters come
+      // before lower case.
       EmptyComponent: React.PropTypes.func,
       ErrorComponent: React.PropTypes.func,
+      children: React.PropTypes.oneOfType([
+        React.PropTypes.object,
+        React.PropTypes.array
+      ]),
       timeout: React.PropTypes.number
+    },
+
+    getInitialState: function() {
+      return {
+        renderEmpty: false,
+        renderError: false
+      };
     },
 
     _validStatuses: ['starting', 'ok', 'empty', 'error'],
 
-    _childrenStatuses: {},
+    _childrenStatuses: new Map(),
+
+    /**
+      When a child status has been updated we need to loop through all of
+      the available statuses to see if they are all 'empty', 'error', or
+      if the timeout has been hit to then decide which component to show.
+
+      @method _checkChildrenStatuses
+    */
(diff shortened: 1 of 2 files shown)


hatched:
#### No QA

It's unlikely that this will pass CI because PhantomJs doesn't support ES6 Maps so this branch cannot land until we remove Phantom.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/5494/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5557/
Test PASSed.

jujugui/static/gui/src/app/components/section-load-watcher/section-load-watcher.js
```diff
@@ -0,0 +1,129 @@
...
+      Checks the state of the component to determine whether it should
+      render the empty, error, or children components.
+
+      @method _renderContent
+    */
+    _renderContent: function() {
+      if (this.state.renderEmpty) {
+        return <this.props.EmptyComponent />;
+      } else if (this.state.renderError) {
+        return <this.props.ErrorComponent />;
```
huwshimi: I'm not sure these should be components, it restricts you to components that don't require props. It might be better to just pass any old html etc. like we do for the extra info for a deployment flow section https://github.com/juju/juju-gui/blob/develop/jujugui/static/gui/src/app/components/deployment-flow/section/section.js#L33

jujugui/static/gui/src/app/components/section-load-watcher/section-load-watcher.js
```diff
@@ -0,0 +1,129 @@
...
+      } else {
+        // Augment the children with the broadcastStatus method
+        const children = React.Children.map(this.props.children, child => {
```
huwshimi: I don't understand what anything here does.

hatched: This clones the child components that were passed into the SLW and adds the `broadcastStatus` prop to it.

jujugui/static/gui/src/app/components/section-load-watcher/section-load-watcher.js
hatched: This is a good point that it doesn't accept props, but I'm not sure I agree that it should only be html. 

I'll investigate how to pass in a component with props.

huwshimi: If you accept anything, like in the extra info for the section, it could include a component or anything else you'd like to display.

kadams54:
Code is :+1: from me.

hatched:
Thanks for the reviews @huwshimi @kadams54 Lets land this now and then do the modifications when we're actually going to use it. :shipit:

jujugui:

PR #1880, opened 2016-07-27
Use real credentials in the deployment flow

huwshimi:


huwshimi:
QA:
- set your guimaas settings
- use the df flag
- load the deployment flow
- choose a cloud
- if you have creds in JEM then you should get the form to select your cred and region
- if not you should get the add credential form
- if you have creds then click on the creds select box
- click on "Add credentail..."
- you should now be shown the form to add credentials
- if you're able to add regions use the CLI then you can test add credentials (this might not currently be working)

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5496/
Test PASSed.

jujugui/static/gui/src/app/components/deployment-flow/credential/credential.js
```diff
@@ -23,12 +23,117 @@ YUI.add('deployment-credential', function() {
...
+      setCredential: React.PropTypes.func.isRequired,
+      setRegion: React.PropTypes.func.isRequired,
+      setTemplate: React.PropTypes.func.isRequired,
+      template: React.PropTypes.string,
+      users: React.PropTypes.object.isRequired,
+      validateForm: React.PropTypes.func.isRequired
     },
 
+    credentialXHR: null,
+    regionsXHR: null,
```
hatched: These two will need to move into the `getInitialState` method as properties defined here are shared across all instances of this component. While here it's not really a problem, it's a good practice to get into. It really belongs in an initializer but this `getInitialState` is the closest we get without using es6 classes.

hatched:
:shipit: I'm shipping this to avoid conflicts but we should still do a follow-up for the above comment.

jujugui:

PR #1887, opened 2016-08-01
Get the list of clouds for the deployment flow

huwshimi:


huwshimi:
QA:
- use the df flag
- open the deployment flow
- you should only see the local cloud (unless you have everything rolled back and set up to use jem and have added a contoller with the cloud option set)

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5512/
Test PASSed.

jujugui/static/gui/src/app/components/deployment-flow/cloud/cloud.js
```diff
@@ -25,34 +25,64 @@ YUI.add('deployment-cloud', function() {
...
+      };
+    },
+
+    componentWillMount: function() {
+      this.props.listClouds((error, clouds) => {
+        if (error) {
+          console.error('Unable to list clouds', error);
+          return;
+        }
+        if (!clouds) {
```
hatched: Instead of doing this you could set it as a default on line 40
```JavaScript
this.props.listClouds((error, clouds=[]) => {
```

huwshimi: Would that work if the API returns `null`?

huwshimi: Yeah, looks like that won't work if the API returns `null`, which just happens to be the thing I'm guarding against.

hatched: Ahh ok, maybe a code comment or explicit check is necessary then, for future us who go back and put in the default value :)

hatched:
I'm going to ship this now to keep things moving along but if you could add that comment in your next branch would be great. Thx! :shipit:

jujugui:

PR #1900, opened 2016-08-04
Expand plans API

hatched:
Add in the missing plans API so they can be used in the subsequent deployment flow UI updates.

### DIFF
--- a/jujugui/static/gui/src/app/assets/javascripts/jujulib/index.js
+++ b/jujugui/static/gui/src/app/assets/javascripts/jujulib/index.js
@@ -70,6 +70,9 @@ var module = module;
       case 'PUT':
         return bakery.sendPutRequest(
             path, JSON.stringify(params), success, failure, redirect);
+      case 'PATCH':
+        return bakery.sendPatchRequest(
+            path, JSON.stringify(params), success, failure, redirect);
       case 'DELETE':
         return bakery.sendDeleteRequest(path, success, failure, redirect);
       default:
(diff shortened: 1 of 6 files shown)


hatched:
No QA necessary as this will be exercised in following branches.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5546/
Test PASSed.

jujugui/static/gui/src/app/assets/javascripts/jujulib/index.js
```diff
@@ -70,6 +70,9 @@ var module = module;
       case 'PUT':
         return bakery.sendPutRequest(
             path, JSON.stringify(params), success, failure, redirect);
+      case 'PATCH':
```
huwshimi: Nice, thanks!

huwshimi:
:+1:

hatched:
:shipit:

jujugui:

jujugui/static/gui/src/app/assets/javascripts/jujulib/plans.js
```diff
@@ -172,6 +172,92 @@ var module = module;
     },
 
     /**
+      Changes the budget to the new limit.
+
+      @method updateBudget
```
frankban: I know this is already merged, but hope this comment can still be useful. I think we should always create a customized, JS friendly, result value from the API server responses. In essence, every time we just `return jujulib._makeRequest(...` we are propagating the server side results to the rest of the application. If the response changes (we know it shouldn't right?) then the application is broken and the error is not intercepted here, but somewhere internally. Moreover, by looking at this code I have basically no idea about what's this `data` passed to the callbacks. I think we should make our assumptions about server side responses more explicit and clear, also considering that we are forcefully using mocks in the tests. For an example of how to do this, see _handlePlan() in this same file.

PR #1924, opened 2016-08-16
API client: ModelGet and DestroyModels calls.

### LINKED ISSUE #1876: Update model methods facade
In Juju master, 3 methods have been moved from the `Client` facade to the `ModelConfig` facade:

- ModelGet
- ModelSet
- ModelUnset

ex) 

```JavaScript
this._send_rpc({
  type: 'Client',
  request: 'ModelGet'
}, cb);
```

now becomes:

```JavaScript
this._send_rpc({
  type: 'ModelConfig',
  request: 'ModelGet'
}, cb);
```

frankban:
The ModelGet call is now done on the ModelConfig facade.
Also, DestroyModels is implemented and will replace the
deprecated DestroyModel (no trailing s) call. The new
endpoint allow for destroying multiple models in the
controller, without being directly connected to them.

Fixes #1876

### DIFF
--- a/jujugui/static/gui/src/app/store/env/api.js
+++ b/jujugui/static/gui/src/app/store/env/api.js
@@ -729,7 +729,7 @@ YUI.add('juju-env-api', function(Y) {
           console.warn('error calling ModelGet API: ' + data.err);
           return;
         }
-        this.set('maasServer', data.config['maas-server']);
+        this.set('maasServer', data.config['maas-server'].value);
       });
     },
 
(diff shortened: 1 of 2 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5606/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5607/
Test PASSed.

frankban:
QA:

- use a recent Juju2 version to bootstrap a controller;
- make dist and upgrade-gui as usual to set up GIJoe with this branch;
- open the GUI in your browser;
- open the JavaScript console;
- define a `var cb = data => {console.log(data)};` callback;
- run `app.env.modelGet(cb);` and check that the callback prints an 
  object with a config field being an object mapping config keys to 
  {source: ..., value: ...} pairs;
- create a couple of new models using the profile page, 
  say "test1" and "test2";
- collect the tags for these new models. To do that, switch to test1 
  and run `var tag1 = app.env.get('modelTag');`. Then switch to
  test2 and run `var tag2 = app.env.get('modelTag');`;
- switch back to test1 and destroy both test1 and test2 models by
  running `app.env.destroyModels([tag1, tag2], cb);`;
- the response printed by the callback should be an object with a 
  single "results" key mapping the tags to null errors;
- from the model switcher, ensure that the two models are gone,
  then switch to the controller model;
- try to destroy a model that doesn't exist: run
  `app.env.destroyModels([tag1], cb);` and this time the callback
  should have received a results with the tag1 key having a 
  "model not found" value;
- try to destroy the controller model: run
  `app.env.destroyModels([app.env.get('modelTag')], cb);`. Note 
  that, surprisingly, this works. Congrats, you just destroyed 
  your controller, and therefore your GUI! Now all you can do 
  is going ahead with 
  `juju destroy-controller --destroy-all-models ...`.

Thank you!

frankban:
This also fixes #1876

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -2803,6 +2799,63 @@ YUI.add('juju-env-api', function(Y) {
...
+      as well, and it's not possible to recover from model's removal.
+      Also note that currently (2016-08-16) nothing prevents this call from
+      destroying the controller model, therefore also disconnecting or even
+      auto-destroying the GUI itself, for instance in the GUI in Juju scenario.
+      For this reason callers are responsible of checking whether a model tag
+      identifies a controller model before calling this method.
+
+      @method destroyModels
+      @param {Array} tags The Juju tags of the models, each one being a string,
+        for instance "model-5bea955d-7a43-47d3-89dd-b02c923e2447".
```
hatched: I don't recall seeing a model uuid with the prefix "model-" is this a requirement for this method? If so can you expand this comment a bit please.

hatched:
Thanks @frankban just one question. 

I'd also like us to move towards using the es6 `let` and `const` instead of using `var`. I won't say you've got to go back and update all these just keep it in mind :)

jujugui/static/gui/src/app/store/env/api.js
frankban: That's a model tag, as the docstring specifies. It's basically a uuid prefixed with "model-" and it's already used in other API calls in api.js.

kadams54:
QA is :ok:. Thanks @frankban!

huwshimi:
Thanks for this branch @frankban :shipit:

jujugui:

PR #1935, opened 2016-08-18
Implement the new ModelInfo method in the sandbox

huwshimi:
This is to stop the warning when the GUI is loaded in sandbox mode.

### DIFF
--- a/jujugui/static/gui/src/app/store/env/sandbox.js
+++ b/jujugui/static/gui/src/app/store/env/sandbox.js
@@ -352,8 +352,8 @@ YUI.add('juju-env-sandbox', function(Y) {
           results: [{
             result: {
               name: 'sandbox',
-              'default-series': 'trusty',
-              'provider-type': 'lxd',
+              'default-series': state.get('defaultSeries'),
+              'provider-type': state.get('providerType'),
               uuid: 'sandboxuuid1',
               'controller-uuid': 'controlleruuid1',
               life: 'alive',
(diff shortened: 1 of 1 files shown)


huwshimi:
QA:
- load the GUI in sandbox mode
- there should be no error in your browser console about a missing `handleModelManagerModelInfo` method

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5626/
Test PASSed.

anthonydillon:
Looks good thanks :+1:

jujugui/static/gui/src/app/store/env/sandbox.js
```diff
@@ -337,6 +337,34 @@ YUI.add('juju-env-sandbox', function(Y) {
...
+    @return {undefined} Side effects only.
+    */
+    handleModelManagerModelInfo: function(data, client, state) {
+      client.receive({
+        'request-id': data['request-id'],
+        response: {
+          results: [{
+            result: {
+              name: 'sandbox',
+              'default-series': 'trusty',
```
frankban: I think we should use state.get('providerType') and state.get('defaultSeries') to get the values here, for consistency.

frankban:
Thanks for this Huw!
👍  with a trivial suggestion and after adding at least a test for the functionality. It can be similar to the existing one for Client.ModelInfo (a call that no longer exists but that I will restore later): see it('can return model information.'.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5627/
Test PASSed.

huwshimi:
Thanks @anthonydillon, @frankban :shipit:

jujugui:

PR #1925, opened 2016-08-16
Use SLW component on UserProfile.

kadams54:

PR #1939, opened 2016-08-19
Fix uncaught exceptions impacting the user profile.

kadams54:


jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/5641/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5642/
Test PASSed.

jujugui/static/gui/src/app/jujulib/index.js
```diff
@@ -57,9 +57,17 @@ var module = module;
       }
       callback(error, data);
     };
-    var failure = function(xhr) {
-      var data = JSON.parse(xhr.target.responseText);
-      var error = data.Message || data.message || data.Error || data.error;
+    var failure = function(xhrOrMessage) {
```
hatched: LETS use LET :)

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -2861,6 +2861,9 @@ YUI.add('juju-env-api', function(Y) {
           var response = data.response;
           transformedData.envs = response['user-models'].map(function(value) {
             var model = value.model;
+            if (!model) {
```
hatched: This shouldn't be possible, can you add a comment as to when this would show up. Also if this is the case I'm not sure returning an array of empty objects is a good idea.

jujugui/static/gui/src/app/jujulib/index.js
hatched: It would probably be nice to have a proper docblock for this fn with a description of the values that could be passed in.

jujugui/static/gui/src/app/jujulib/test-index.js
```diff
@@ -0,0 +1,39 @@
+/* Copyright (C) 2016 Canonical Ltd. */
+
+'use strict';
+
+chai.config.includeStack = true;
+chai.config.truncateThreshold = 0;
+
+describe('jujulib utility functions', function() {
```
hatched: Heh I can't believe we missed that this file wasn't tested.

jujugui/static/gui/src/app/jujulib/index.js
kadams54: Ah yes, good reminder to use `let` and `const` wherever possible. That said, since this is in jujulib (which I don't think is transpiled) I'm inclined to stick with `var` here. I'll add docblocks for both callbacks.

hatched: I moved the jujulib into the tree last week so that it now gets properly transpiled :)

jujugui/static/gui/src/app/jujulib/test-index.js
kadams54: I'm sure if you checked a code coverage tool, the code would be thoroughly exercised via the various `test-jimm.js`, `test-charmstore.js`, etc. The problem was that those aren't proper unit tests of `_makeRequest`. Anyhow, glad we now have a home for direct, isolated testing of an important function.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5644/
Test PASSed.

jujugui/static/gui/src/app/store/env/api.js
kadams54: Tracked this down to bad data being returned by `sandbox.js`, so I removed this logic and made the fix there.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5645/
Test PASSed.

kadams54:
## QA, Part 1

This QA is for this PR specifically. I'm going to add a second set of QA instructions to cover #1925. In Chrome:

1. Load the profile page, `/profile/`, in an incognito window.
2. Pop open your developer tools. Turn on breaks for uncaught exceptions.
3. Right click on the reload button and select "Emtpy cache and hard reload".

You'll see a number of errors in the console, but no uncaught exceptions should trigger breakpoints. Additionally, the user profile page should load completely: no hanging spinners.

hatched:
Thanks for getting the profile page to be resilient!

:shipit:

jujugui:

PR #1955, opened 2016-08-30
Enable model switching via JIMM

hatched:
Add the ability to switch models when using JIMM. This functionality already worked and existed with GiJoe.


hatched:
#### To QA

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/5691/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5694/
Test PASSed.

jujugui/static/gui/src/app/app.js
```diff
@@ -660,6 +660,12 @@ YUI.add('juju-gui', function(Y) {
             this.controllerAPI.connect();
           }
         }
+        // If we have a jimmURL then we need to rely on the controllerAPI
```
frankban: Ok so my impression is that the logic around the GUI connection is starting to be a bit complex.
My impression is that the GUI should not have a concept of jimm: a jimmURL config options smells.
The promise is that we can connect to a Juju controller or to a Jimm in a transparent way, without the client having to know something specific to the latter, i.e. without the client having any special logic related to jimm. The only exception is handling the redirections when logging in to models, but that can be handled generically, as a possibility in the API, without knowing there is a jimm.

I'd really like to see some simplifications and less special cases in the code: we already have 4 config values that we use to build the WebSocket URLs where to connect:
1) apiAddress from which we can parse the base host and port;
2) jujuEnvUUID that holds the UUID;
3) socketTemplate that we can use together with host, port and uuid from points 1 and 2 to build the URL for establishing the model connection;
4) controllerSocketTemplate that we can use together with host and port from point 1 to build the URL for establishing the controller connection.

So connecting to jimm from the GUI perspective is not a special case: it is just a controller connection, so point 4 should be used, no need for introducing a new config value.

Let's consider this code here:
```javascript
         if (isNotGISFSandbox || isNotConnected) {
           this.env.connect();
           if (this.controllerAPI) {
             // We won't have a controller API connection in Juju 1 or in
             // sandbox mode.
              this.controllerAPI.connect();
            }
          }
          if (this.get('jimmURL')) {
           this.controllerAPI.connect();
         }
```
In this snippet, `this.controllerAPI.connect();` could be called twice based on unrelated stuff. Moreover, I am not sure why `this.env.connect()` should not be called in sandbox mode, and it's surely wrong that in sandbox mode we don't have a controller connection. To be clear, IMHO we should remove all special cases for sandbox mode and for jimm.

jujugui/static/gui/src/app/app.js
```diff
@@ -757,15 +763,21 @@ YUI.add('juju-gui', function(Y) {
             return;
           }
           const selectedModel = this._pickModel(modelList);
+          const jimmURL = this.get('jimmURL');
+          let socketTemplate = this.get('socketTemplate');
+
+          if (jimmURL) {
+            socketTemplate = `${jimmURL}${socketTemplate}`;
```
frankban: I am really confused about this: connecting to jimm should be the same as connecting to a controller, so the controllerSocketTemplate should be used. This concatenation of a URL and a template feels wrong, like an hack, and we are doing this hack in two different places in the code, which is also a time bomb. I'd remove all of this, all mentions to jimm, and let the GUI handle the jimm connection with just passing the proper values on the 4 config options that I mentioned above.

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -668,6 +682,21 @@ YUI.add('juju-env-api', function(Y) {
     },
 
     /**
+      Performs the RedirectInfo request.
+
+      @method redirectInfo
+      @param {Function} callback The callback to call with the response.
+    */
+    redirectInfo: function(callback) {
```
frankban: I'd just pass the relevant data to the callback, not the whole response.

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -504,7 +504,11 @@ YUI.add('juju-env-api', function(Y) {
         }
         // Clean up for log out text.
         this.failedAuthentication = false;
-      } else {
+      } else if (
+        !data.error || data.error.indexOf('redirection required') === -1) {
+        // If we are requiring to do a redirect to properly connect to the
+        // model then do not set credentials or the failed authentication
```
frankban: s/do not set/do not reset/ ?

jujugui/static/gui/src/app/app.js
```diff
@@ -856,6 +867,39 @@ YUI.add('juju-gui', function(Y) {
...
+        // request after the controllerAPI call has connected.
+        const selectedModel = this._pickModel(
+          this.get('environmentList'));
+        let socketTemplate = this.get('socketTemplate');
+        // When connecting to a model using JIMM and a redirection, the socket
+        // template base differs from the default so we must provide
+        // it here when creating the socket url.
+        if (this.get('jimmURL')) {
+          socketTemplate =
+            `wss://${server[0].value}:${server[0].port}${socketTemplate}`;
```
frankban: So here the socket template should be set to wss://$host:$port... already, and then the real values rendered there because they are part of the context. Otherwise this will never work on wss proxy cases and it basically breaks the whole concept of socket templates, and the code relying to them.

jujugui/static/gui/src/app/app.js
```diff
@@ -856,6 +867,39 @@ YUI.add('juju-gui', function(Y) {
...
+      @param {Object} resp The response value containing the available server
+        details.
+    */
+    _handleRedirectionRequired: function(err, resp) {
+      // If we have a response and we have a redirection required error
+      // then we need to modify the connection information and try again.
+      // Otherwise we just carry on not doing anything.
+      if (resp && (err && err.indexOf('redirection required') > -1)) {
+        // Loop through the available servers and find the public IP.
+        const server = resp.servers[0].filter(
```
frankban: I think we should just provide the list of servers to callbacks not the response itself. Also, what do you think about keeping the signature of loginWithMacaroon as is (just returning an error, not a response) making the error so that we can have a isRedirectError(err) function and let the client code handle the call to redirectInfo?
Something like the following (I would find that more readable):
```javascript
api.loginWithMacaroon(new Y.juju.environments.web.Bakery({
  webhandler: new Y.juju.environments.web.WebHandler(),
  interactive: true,
  serviceName: 'juju',
  dischargeStore: window.localStorage
}), err => {
  if (!isRedirectError(err)) {
    // notify the user about the actual error;
    return;
  }
  // We need to redirect the model connection to a new URL.
  api.redirectInfo((servers, err) => {
    if (err) {
      // notify the user about the actual error;
      return;
    }
    // Create a new model connection using one of the "servers".
    ...
  });
});
```

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -607,6 +611,16 @@ YUI.add('juju-env-api', function(Y) {
 
       // Define the handler reacting to Juju controller login responses.
       var handleResponse = function(bakery, macaroons, cback, data) {
+        // When using JIMM it's possible that the url we have for the model is
+        // not the correct host and we need to fetch the correct path to be
+        // able to connect to the model.
+        if (data.error && data.error === 'redirection required') {
+          this.redirectInfo(resp =>
```
frankban: As mentioned, I'd let callers do this, this method here already has enough responsibility.

frankban:
Hey Jeff,
thanks for working on this, as you can see I have some concerns and suggestions. I'd like the GUI to remain generic and easy to reason about enough, especially when handling controller vs jimm and real connections vs sandbox. Let's discuss later if anything is not clear. Thank you and sorry for the verbosity of this review!

jujugui/static/gui/src/app/app.js
frankban: I am fixing the fact that the controller connection is not made on sandbox in my current branch.

frankban: The fix is ready for review at https://github.com/juju/juju-gui/pull/1958

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/5719/
Test FAILed.

jujugui/static/gui/src/app/store/env/api.js
hatched: updated

jujugui/static/gui/src/app/store/env/api.js
hatched: done

jujugui/static/gui/src/app/app.js

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/5720/
Test FAILed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/5722/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5730/
Test PASSed.

development.ini
```diff
@@ -18,6 +18,20 @@ jujugui.controllerSocketTemplate = /api
 jujugui.socketTemplate = /model/$uuid/api
 jujugui.gisf = false
 
+; To use a JIMM backend uncomment the following:
+;   The actual websocket address for the JIMM you'd like to connect to.
```
frankban: s/websocket/WebSocket/

development.ini
```diff
@@ -18,6 +18,20 @@ jujugui.controllerSocketTemplate = /api
...
+;jujugui.apiAddress =
+;   JIMM socket template.
+;jujugui.socketTemplate = wss://$server:$port/model/$uuid/api
+;   JIMM controller socket template.
+;jujugui.controllerSocketTemplate = wss://$server:$port/api
+;   Without gisf set to true the GUI will require the username and
+;   password instead of relying on the macaroon auth.
+;jujugui.gisf = true
+;   Setting the sandbox to false will cause the GUI to instantiate a
+;   websocket connection instead of the fakebackend.
```
frankban: s/websocket/WebSocket/

development.ini
```diff
@@ -28,28 +42,24 @@ jujugui.interactive_login = true
 
 # Default  URLS
 jujugui.jem_url =
-jujugui.jimm_url =
```
frankban: \o/

jujugui/static/gui/src/app/app.js
```diff
@@ -652,12 +651,8 @@ YUI.add('juju-gui', function(Y) {
                               (this.get('socket_url') || this.get('sandbox')));
         if (isNotGISFSandbox || isNotConnected) {
           this.env.connect();
-          if (this.controllerAPI) {
-            // We won't have a controller API connection in Juju 1 or in
```
frankban: Why is this comment no longer true? Isn't it still possible that we are on Juju 1 and this.controllerAPI is undefined?

jujugui/static/gui/src/app/app.js
```diff
@@ -748,13 +743,26 @@ YUI.add('juju-gui', function(Y) {
           // user into an uncommitted state.
           if (modelList.length === 0) {
             // XXX Drop the user into the uncommitted state.
+            console.log('No models available, using sandbox mode.');
```
frankban: `console.log('no models available: using unconnected mode');` ?

jujugui/static/gui/src/app/app.js
```diff
@@ -748,13 +743,26 @@ YUI.add('juju-gui', function(Y) {
...
             // leave it be.
             return;
           }
+          // Pick a model to connect to.
           const selectedModel = this._pickModel(modelList);
+          // Set the selected uuid as the active model in the GUI. If this is
+          // not set here then the subsequent code will not know what the
+          // uuid of the model we're supposed to connect to is.
+          if (!selectedModel) {
+            console.log('Cannot select available model, using sandbox mode.');
```
frankban: console.log('cannot select available model: using unconnected mode'); ?

jujugui/static/gui/src/app/app.js
```diff
@@ -854,6 +858,46 @@ YUI.add('juju-gui', function(Y) {
     },
 
     /**
+      Callback handler for the api loginWithMacaroon method which hanadles
```
frankban: s/api/API and also s/hanadles/handles/

jujugui/static/gui/src/app/app.js
```diff
@@ -854,6 +858,46 @@ YUI.add('juju-gui', function(Y) {
...
+      the "redirection required" error message.
+
+      @method _apiLoginHandler
+      @param {Object} api The API that the user is attempting to log into.
```
frankban: ... like this.env or this.controllerAPI

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -595,7 +599,7 @@ YUI.add('juju-env-api', function(Y) {
       var cback = function(err, response) {
         this.handleLogin({error: err, response: response});
         if (callback) {
-          callback(err);
+          callback(err, response);
```
frankban: I think this change is not required, and does not reflect the docstring of this method.

jujugui/static/gui/src/app/app.js
```diff
@@ -854,6 +858,46 @@ YUI.add('juju-gui', function(Y) {
...
+      @param {String} err The login error message, if any.
+      @param {Object} data The login request response data, if any.
```
frankban: Data is not required not provided, please remove.

jujugui/static/gui/src/app/app.js
```diff
@@ -854,6 +858,46 @@ YUI.add('juju-gui', function(Y) {
...
+    */
+    _apiLoginHandler: function(api, err, data) {
```
frankban: Please remove the data parameter here.

jujugui/static/gui/src/app/app.js
```diff
@@ -854,6 +858,46 @@ YUI.add('juju-gui', function(Y) {
...
+      const errorNotify = function(err) {
+        this.db.notifications.add({
+          title: 'Unable to log into model',
```
frankban: Unable to log into Juju? So that we don't have to do model vs controller here.

jujugui/static/gui/src/app/app.js
```diff
@@ -854,6 +858,46 @@ YUI.add('juju-gui', function(Y) {
...
+          message: `unable to log into model: ${err}`,
+          level: 'error'
+        });
+      };
+
+      if (err === 'authentication failed: redirection required') {
```
frankban: I'd create a isRedirectError(err) function here, which encapsulates the error checking logic. String checking is already a weak approach, but it's the best we can do, but at least an external function will be more likely reused and if the message change there will be a single place affected.

jujugui/static/gui/src/app/app.js
frankban: Oh I see now that this is only used when the model connection fails, so nevermind.

jujugui/static/gui/src/app/app.js
```diff
@@ -854,6 +858,46 @@ YUI.add('juju-gui', function(Y) {
...
+        // If the error is that redirection is required then we have to
+        // make a request to get the appropriate model connection information.
+        api.redirectInfo((err, servers) => {
+          if (err) {
+            errorNotify(err);
+            return;
+          }
+          // Loop through the available servers and find the public IP.
+          const server = servers[0].filter(
+            server => server.scope === 'public');
```
frankban: This is very cool, but I am not sure we'll always have a public scoped address on all providers.
I'd try this first and then fall back to juju.chooseAddress() (see net-utils.js) if no address is found.

jujugui/static/gui/src/app/app.js
```diff
@@ -854,6 +858,46 @@ YUI.add('juju-gui', function(Y) {
...
+          // Switch to the redirected model.
+          this.switchEnv(this.createSocketURL(
```
frankban: This is great!

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -504,7 +504,11 @@ YUI.add('juju-env-api', function(Y) {
         }
         // Clean up for log out text.
         this.failedAuthentication = false;
-      } else {
+      } else if (
+        !data.error || data.error.indexOf('redirection required') === -1) {
```
frankban: There you go, second instance of the same error, an external function will be already useful!

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -668,6 +672,27 @@ YUI.add('juju-env-api', function(Y) {
     },
 
     /**
+      Performs the RedirectInfo request.
+
+      @method redirectInfo
+      @param {Function} callback The callback to call with the response.
```
frankban: Please describe what's received by the callback.

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -668,6 +672,27 @@ YUI.add('juju-env-api', function(Y) {
...
+    */
+    redirectInfo: function(callback) {
+      this._send_rpc({
+        type: 'Admin',
+        request: 'RedirectInfo',
+        version: ADMIN_FACADE_VERSION
+      }, resp => {
+        const error = resp.response.error;
+        if (error) {
+          callback(error, resp);
```
frankban: Please keep the two callback argument consistent:
```javascript
if (error) {
  callback(error, []);
  return;
}
```

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -668,6 +672,27 @@ YUI.add('juju-env-api', function(Y) {
...
+      @method redirectInfo
+      @param {Function} callback The callback to call with the response.
+    */
+    redirectInfo: function(callback) {
+      this._send_rpc({
+        type: 'Admin',
+        request: 'RedirectInfo',
+        version: ADMIN_FACADE_VERSION
+      }, resp => {
+        const error = resp.response.error;
```
frankban: I'd start with
```javascript
if (!callback) {
  console.log('data returned by Admin.RedirectInfo API call:', resp);
  return;
}
```

jujugui/static/gui/src/test/test_env_api.js
```diff
@@ -559,6 +559,37 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+          type: 'Admin',
+          version: 3
+        };
+        assert.deepEqual(expected, lastMessage);
+      });
+
+      it('responds with only the server data if no error', function() {
+        const callback = sinon.stub();
+        env.redirectInfo(callback);
+        conn.msg({ 'request-id': 1, response: { servers: 'servers' }});
```
frankban: I'd use `{servers: [{...}, {...}]}` to make the mock more realistic and the test more informative for future readers.

frankban:
This is great stuff Jeff thank you!
I have some comments and suggestions, but nothing major.
👍 with the requested changes.

development.ini
hatched: done

development.ini
hatched: done

jujugui/static/gui/src/app/app.js

jujugui/static/gui/src/app/app.js

jujugui/static/gui/src/app/app.js

jujugui/static/gui/src/app/app.js

jujugui/static/gui/src/app/store/env/api.js
hatched: Good catch, I updated the docstring - I found it very helpful to debug the error case when the response was passed through.

jujugui/static/gui/src/app/app.js
hatched: It is provided, see above comment, but it is not used.

jujugui/static/gui/src/app/app.js
hatched: See above

jujugui/static/gui/src/app/app.js

jujugui/static/gui/src/app/store/env/api.js
hatched: added

jujugui/static/gui/src/app/store/env/api.js
hatched: I find that passing the entire response back to the 'client' in error has been very helpful when debugging issues.

jujugui/static/gui/src/app/store/env/api.js
hatched: done

jujugui/static/gui/src/app/app.js
hatched: good idea, added

jujugui/static/gui/src/app/store/env/api.js
hatched: done

jujugui/static/gui/src/app/app.js
hatched: Good catch, I needed to move this conditional to the line below on 655!

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5743/
Test PASSed.

jujugui/static/gui/src/app/store/env/api.js
frankban: I am not sure I like that the full response is passed to callers. If the goal is only debugging, then we can log the response here in case of errors, or include a serialized version of the response in the returned error. I know the response is not used by callers, but nothing prevents them to use it, and theoretically we are coupling all the places in which this API is called to the response from the server, which should only be handled in one place, here.

hatched: alright I've removed it, I don't feel that strongly about it.

jujugui/static/gui/src/app/app.js
hatched: Good idea, unfortunately because JIMM is down atm I am not able to actually test this so I'll add this as a follow-up

jujugui/static/gui/src/test/test_env_api.js
hatched: Good call

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5748/
Test PASSed.

hatched:
Thanks for the reviews @frankban I'm going to land this in preparation for the 2.1.11 release and create follow-ups for the JIMM items that need to be tackled. 

:shipit:

jujugui:

PR #1962, opened 2016-09-01
Update old tests to use sinon

huwshimi:


huwshimi:
QA: no QA, tests should pass.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5712/
Test PASSed.

frankban:
I believe this is a mechanical branch, on which you did multiple search and replace operations.
In these cases i usually like that the authors describes exactly what they did, for instance "I s/.allArguments()/.args/g" and similar, for future reference.
:+1: with the above added to the PR description. Thank you for cleaning up legacy stuff!

jujugui/static/gui/src/test/utils.js
```diff
@@ -78,21 +36,30 @@ YUI(GlobalConfig).add('juju-tests-utils', function(Y) {
      * @param {String} name the name to be replaced.
      * @return {Function} the new stub function.
      */
-    makeStubMethod: function(context, name) {
-      var responses = Array.prototype.slice.call(arguments, 2);
-      var original = context[name];
-      var f = context[name] = jujuTests.utils.makeStubFunction.apply(
-          jujuTests.utils, responses);
-      f.reset = function() {
+    makeStubMethod: function(context, name, args) {
```
hatched: Is it in the plan to go through and remove this one as well?

hatched:
:+1: Thanks for this @huwshimi you're really cleaning up our codebase :)

kadams54:
First off, this is a-MAY-zing! :tada:

Secondly: I'm curious why as to why `makeStubMethod` is sticking around. No objections to the branch landing, just curious.

:+1: on a job well done. Thanks @huwshimi!

kadams54:
:shipit:

jujugui:

huwshimi:
`makeStubMethod` does two things, mocks a method on an object (now using a sinon.stub) but secondly provides a way to clean up that stub (return it to the original method). We could remove `makeStubMethod` and manually add cleanups each time we replace a method, but it still seems helpful to have a util to handling this. If someone has a better idea we can certainly take another look!

PR #1963, opened 2016-09-01
Remove more YUI

huwshimi:


huwshimi:
QA: No QA, tests should pass.

Note: each commit is logical so you could review one by one if that helps.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/5713/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5714/
Test PASSed.

frankban:
:+1: thanks Huw!
We should also s/Y.JSON/JSON/, parhaps in a follow up?

jujugui/static/gui/src/app/models/bundle.js
```diff
@@ -86,7 +86,7 @@ YUI.add('juju-bundle-models', function(Y) {
     */
     parseNameEmail: function(author) {
       var parts = /^([^<]+?) <(.+)>$/.exec(author);
-      if (Y.Lang.isNull(parts)) {
+      if (!parts) {
```
hatched: technically this would be `parts === null` but I'm not sure that we actually 'need' it to be null here, could you sanity check that for me please.

jujugui/static/gui/src/app/utils/environment-change-set.js
```diff
@@ -92,8 +92,8 @@ YUI.add('environment-change-set', function(Y) {
       var cut;
       // If there is an object after the callback it's the
       // configuration object for ECS.
-      if (Y.Lang.isObject(lastParam) &&
-          Y.Lang.isFunction(args[args.length - 2])) {
+      if (typeof lastParam === 'object' &&
```
hatched: This will have to expand to also have a not null check because `typeof null === 'object'` as well.

`typeof lastParam === 'object' && lastParam !== null && ... `

jujugui/static/gui/src/app/views/topology/relation.js
```diff
@@ -841,15 +837,13 @@ YUI.add('juju-topology-relation', function(Y) {
 
       // Transform endpoints into a list of relatable services (to the
       // service).
-      var possible_relations = Y.Array.map(
-          Y.Array.flatten(Y.Object.values(endpoints)),
-          function(ep) {return ep.service;});
+      var possible_relations = Y.Array.flatten(Y.Object.values(endpoints)).map(
```
hatched: This one still has some YUI :)

jujugui/static/gui/src/app/views/topology/utils.js
```diff
@@ -112,7 +112,7 @@ YUI.add('juju-topology-utils', function(Y) {
     @return {array} A list of coordinate pairs.
   */
   utils.serviceBoxesToVertices = function(serviceBoxes) {
-    return Y.Array.map(Y.Object.values(serviceBoxes), function(box) {
+    return Y.Object.values(serviceBoxes).map(box => {
```
hatched: Some left over YUI

jujugui/static/gui/src/test/test_model.js
```diff
@@ -440,9 +440,9 @@ describe('test_model.js', function() {
                 // This means that it will update the aggregate
                 // relations for databinding
                 handler.detach();
-                var isObject = Y.Lang.isObject;
                 assert.equal(
-                    isObject(service.get('relationChangeTrigger')), true);
+                  typeof service.get('relationChangeTrigger') === 'object',
```
hatched: See above comment about `typeof null === 'object'`

jujugui/static/gui/src/app/models/bundle.js
huwshimi: Yeah, I originally had that as a null check (pretty sure exec returns null if there is no match) but I realised in this case it doesn't actually need to be a null check, just truthy.

jujugui/static/gui/src/app/views/topology/relation.js
huwshimi: I was trying to do these by removing each YUI util one by one, especially with some of these where we will need to create our own util.

jujugui/static/gui/src/app/views/topology/utils.js
huwshimi: As above.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5723/
Test PASSed.

huwshimi:
Thanks @frankban, @hatched.

huwshimi:
:shipit:

jujugui:

PR #1965, opened 2016-09-01
Controller API Cloud facade: UserCredentials and Credential calls.

frankban:
**This is now ready for review**.
**Todo for this branch**:
- [x] enable sandbox mode for the two calls;
- [x] tests for new methods in controller API;
- [x] tests for new sandbox methods.

### DIFF
--- a/jujugui/static/gui/src/app/store/env/controller-api.js
+++ b/jujugui/static/gui/src/app/store/env/controller-api.js
@@ -841,6 +841,138 @@ YUI.add('juju-controller-api', function(Y) {
         storageEndpoint: result['storage-endpoint'] || '',
         regions: regions
       };
+    },
+
+    /**
+      Returns the tags of cloud credentials for a set of users.
+
+      @method getTagsForCloudCredentials
+      @param {Array} userCloudPairs A list of (user-tag, cloud-tag) pairs for
+        which to retrieve the credentials, like
+        [['user-admin', 'cloud-google'], ['user-who@external', 'cloud-lxd']].
+      @param {Function} callback A callable that must be called once the
+        operation is performed. It will receive two parameters: an error (as a
+        string describing the problem) and a sequence of results. In the
+        sequence of results each result refers to the corresponding provided
+        user/cloud pair and holds an object with the following fields:
+        - err: a possible result specific error, in which case all subsequent
+          fields are omitted;
+        - tags: the list of tags that identify cloud credentials corresponding
+          to the user/cloud pair provided as input.
+        If no errors occur, error parameters are null. Otherwise, in case of
+        errors, the second argument is an empty array.
+    */
+    getTagsForCloudCredentials: function(userCloudPairs, callback) {
+      // Decorate the user supplied callback.
+      const handler = function(userCallback, data) {
+        if (!userCallback) {
+          console.log(
+            'data returned by Cloud.UserCredentials API call:', data);
+          return;
+        }
+        if (data.error) {
+          userCallback(data.error, []);
+          return;
+        }
+        const results = data.response.results;
+        if (!results) {
+          userCallback(null, []);
+          return;
+        }
+        const credentials = results.map(result => {
+          const err = result.error && result.error.message;
+          if (err) {
+            return {err: err};
+          }
+          return {tags: result.result || []};
+        });
+        userCallback(null, credentials);
+      }.bind(this, callback);
+      // Send the API request.
+      if (!userCloudPairs.length) {
+        userCloudPairs = [];
+      }
+      const userClouds = userCloudPairs.map(userCloud => {
+        return {'user-tag': userCloud[0], 'cloud-tag': userCloud[1]};
+      });
+      this._send_rpc({
+        type: 'Cloud',
+        request: 'UserCredentials',
+        params: {'user-clouds': userClouds}
+      }, handler);
+    },
+
+    /**
+      Return the specified cloud credentials for each tag, minus secrets.
+
+      @method getCloudCredentials
+      @param {Array} tags The Juju tags of the credentials, each one being a
+        string, for instance "cloudcred-google_dalek@local_google". Tags for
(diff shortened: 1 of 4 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5718/
Test PASSed.

frankban:
QA:
both sandbox and real:

- Run `app.controllerAPI.listClouds(console.log)`.
  This call returns an object whose keys are cloud tags, use one of those in the following instructions, let's say it's "cloud-lxd".
- Run `app.controllerAPI.getTagsForCloudCredentials([['user-admin', 'cloud-lxd']], (err, tags) => {console.log('err:', err, 'tags:', tags)})`.
  After running this call, tags should contain an array of objects (only one in this case) each one having a tags attribute. Pick one tag, for instance "cloudcred-lxd_admin@local_default".
- Run `app.controllerAPI.getCloudCredentials(['cloudcred-lxd_admin@local_default'], (err, creds) => {console.log('err:', err, 'creds:', creds)})`.
  The returned object should have a single "cloudcred-lxd_admin@local_default" key,
  with cloud credentials info as attributes.
- Run `app.controllerAPI.getCloudCredentials(['cloudcred-foo', 'bar'], (err, creds) => {console.log('err:', err, 'creds:', creds)})`.
  Check that errors are returned in the values for both keys.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5726/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5727/
Test PASSed.

jujugui/static/gui/src/app/store/env/controller-api.js
```diff
@@ -841,6 +841,138 @@ YUI.add('juju-controller-api', function(Y) {
         storageEndpoint: result['storage-endpoint'] || '',
         regions: regions
       };
+    },
+
+    /**
+      Returns the tags of cloud credentials for a set of users.
+
+      @method getTagsForCloudCredentials
+      @param {Array} userCloudPairs A list of (user-tag, cloud-tag) pairs for
```
mhilton: This must be the only actual client implementation that makes use of the bulk API. nice.

mhilton:
:+1:

jujugui/static/gui/src/app/store/env/controller-api.js
```diff
@@ -841,6 +841,138 @@ YUI.add('juju-controller-api', function(Y) {
...
+          }
+          return {tags: result.result || []};
+        });
+        userCallback(null, credentials);
+      }.bind(this, callback);
+      // Send the API request.
+      if (!userCloudPairs.length) {
+        userCloudPairs = [];
+      }
+      const userClouds = userCloudPairs.map(function(userCloud) {
```
hatched: In case you weren't aware you could also do implicit returns with arrow fn's
```JavaScript
const userClouds = userCloudPairs.map(userCloud =>
  {'user-tag': userCloud[0], 'cloud-tag': userCloud[1]})
```
I have come to quite like them but I'll leave it up to you if you want to change it.

jujugui/static/gui/src/app/store/env/controller-api.js
```diff
@@ -841,6 +841,138 @@ YUI.add('juju-controller-api', function(Y) {
...
+          };
+          return prev;
+        }, {});
+        userCallback(null, credentials);
+      }.bind(this, callback);
+      // Send the API request.
+      if (!tags.length) {
+        tags = [];
+      }
+      const entities = tags.map(function(tag) {
```
hatched: As I mentioned before about implicit returns I think it's cases like this where it really shines.
```JavaScript
const entities = tags.map(tag => ({tag}));
```
In this case I used two new tricks. First because you're returning an object we have to wrap the curlies in a parens else it'll think it's the opening brackets to the function. The second is that if the key and value are the same string then you can shorten it to just `{tag}` and it'll expand out internally to `{tag: tag}`.

jujugui/static/gui/src/test/test_controller_api.js
```diff
@@ -1358,4 +1358,161 @@ describe('Controller API', function() {
 
   });
 
+  describe('getTagsForCloudCredentials', function() {
+    it('retrieves tags for cloud credentials', function(done) {
+      // Perform the request.
+      const pairs = [
+        ['user-who@gallifrey', 'cloud-lxd'],
+        ['user-dalek@skaro', 'cloud-google'],
+        ['user-rose@earth', 'cloud-google']
```
hatched: haha nice :)

hatched:
Thanks for this @frankban I have a few little comments, do with them what you wish.

I did the QA with Sandbox and Beta 17 and everything works as it should!

jujugui/static/gui/src/app/store/env/controller-api.js
frankban: Done.

jujugui/static/gui/src/app/store/env/controller-api.js
frankban: Hmm... the fact that you need to add parens makes this slightly less readable. I guess I need to get used to that.

jujugui/static/gui/src/app/store/env/controller-api.js
frankban: Actually, the linter complans:  Parsing error: Unexpected token :
I guess I'll need the parens here as well :-/

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5735/
Test PASSed.

frankban:

jujugui:

jujugui/static/gui/src/app/store/env/controller-api.js
hatched: Oops yes sorry my bad

PR #1970, opened 2016-09-05
Fix local charm deploy with juju 2

### LINKED ISSUE #1947: Errors when deploying a local charm with Juju 2
When deploying a local charm with Juju 2 (in my case 2.0-beta16-xenial-amd64) I go the error `Could not add requested charm. Server responded with: only charm store charm URLs are supported, with cs: schema`. The charm did deploy however. There were no errors when adding the charm to the canvas.

The charm id both before deploy and the url sent with the Juju RPC call used `local:trusty/apache2-25`.

huwshimi:
Fixes #1947.

### DIFF
--- a/jujugui/static/gui/src/app/views/ghost-deployer-extension.js
+++ b/jujugui/static/gui/src/app/views/ghost-deployer-extension.js
@@ -74,12 +74,14 @@ YUI.add('ghost-deployer-extension', function(Y) {
       var serviceName = ghostService.get('name');
       var charmId = this._addSeriesToCharmId(charm.get('id'), activeSeries);
       var constraints = {};
-      // TODO frankban: add support for fetching delegatable macaroons that can
-      // be used to add private charms.
-      this.env.addCharm(
-        charmId, null, this._addCharmCallbackHandler.bind(this, charm),
-        // Options used by ECS, ignored by environment.
-        {applicationId: ghostServiceId});
+      if (this.isLegacyJuju() || charm.get('id').indexOf('local:') === -1) {
+        // TODO frankban: add support for fetching delegatable macaroons that
+        // can be used to add private charms.
+        this.env.addCharm(
+          charmId, null, this._addCharmCallbackHandler.bind(this, charm),
+          // Options used by ECS, ignored by environment.
+          {applicationId: ghostServiceId});
+      }
       this.env.deploy(
           charmId,
           activeSeries,
--- a/jujugui/static/gui/src/app/views/ghost-deployer-extension.js
+++ b/jujugui/static/gui/src/app/views/ghost-deployer-extension.js
@@ -74,7 +74,7 @@ YUI.add('ghost-deployer-extension', function(Y) {
       var serviceName = ghostService.get('name');
       var charmId = this._addSeriesToCharmId(charm.get('id'), activeSeries);
       var constraints = {};
-      if (this.isLegacyJuju() || charm.get('id').indexOf('local:') === -1) {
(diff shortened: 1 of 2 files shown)


huwshimi:
QA:
- deploy with juju2
- drop a local charm and deploy
- you should not get an error about not being able to add the charm

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5734/
Test PASSed.

jujugui/static/gui/src/app/views/ghost-deployer-extension.js
```diff
@@ -74,12 +74,14 @@ YUI.add('ghost-deployer-extension', function(Y) {
       var serviceName = ghostService.get('name');
       var charmId = this._addSeriesToCharmId(charm.get('id'), activeSeries);
       var constraints = {};
-      // TODO frankban: add support for fetching delegatable macaroons that can
-      // be used to add private charms.
-      this.env.addCharm(
-        charmId, null, this._addCharmCallbackHandler.bind(this, charm),
-        // Options used by ECS, ignored by environment.
-        {applicationId: ghostServiceId});
+      if (this.isLegacyJuju() || charm.get('id').indexOf('local:') === -1) {
```
frankban: I think the idea here is that if a charm is local then it's been already added (via the HTTP API) and so we don't need to add it again with AddCharm (which uses the charm store and does not support local charms). AFAIK this is not based on the Juju version being used. So I am not sure about the isLegacyJuju() check, I'd just skip adding the charm `if charm.get('id').startsWitth('loca')`.

jujugui/static/gui/src/test/test_ghost_deployer_extension.js
```diff
@@ -145,6 +145,15 @@ describe('Ghost Deployer Extension', function() {
     assert.equal(ghostDeployer.env.deploy.callCount, 1);
   });
 
+  it('does not call addCharm for local charms and juju 2', function() {
```
frankban: Please add two similar tests, one for Juju 1 and one for Juju 2. If you agree with my comment above AddCharm should not be called in both cases.

frankban:
👍 with a suggestion, thanks!

jujugui/static/gui/src/app/views/ghost-deployer-extension.js
anthonydillon: Seems to me that detecting if the charm is local when added to the canvas and set a variable on the charm object would be better. So we have a single source of detection logic throughout the GUI.

anthonydillon:
QA:
- Downloaded the charm (https://jujucharms.com/apache2/)
- Build with the latest Juju
- Dropped the zip apache2 onto the canvas and got no errors
- Deployed the charm
- It went into pending state then errored with hook failed: "config-changed"
- Then deployed wordpress from a local download, which deployed successfully

Marking as QA ok. Believe the issue was with the apache2 charm.

hatched:
Note for QA, it should also be done in the charm with Juju 1 to be sure we've maintained backwards compatibility.

huwshimi:
I've updated this to also not call `addCharm` with Juju 1 (I tested with AWS and it all worked).

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5742/
Test PASSed.

hatched:
Thanks @huwshimi :+1: :ok: on both Juju 1 and 2

:shipit:

jujugui:

PR #1971, opened 2016-09-05
Controller API Cloud facade: (Update|Revoke)Credentials calls.

frankban:
This completes the Cloud facade API client.

### DIFF
--- a/jujugui/static/gui/src/app/store/env/controller-api.js
+++ b/jujugui/static/gui/src/app/store/env/controller-api.js
@@ -973,6 +973,104 @@ YUI.add('juju-controller-api', function(Y) {
         request: 'Credential',
         params: {entities: entities}
       }, handler);
+    },
+
+    /**
+      Create or update a single cloud credential.
+
+      @method updateCloudCredential
+      @param {String} tag The Juju tag of the credential as a string, for
+        instance "cloudcred-google_dalek@local_google". Tags for credentials
+        are usually retrieved by calling the getTagsForCloudCredentials method.
+      @params {String} authType The authentication type, like "userpass",
+        "oauth2" or just "empty".
+      @params {Object} attrs Attributes containing credential values, as an
+        object mapping strings to strings.
+      @param {Function} callback A callable that must be called once the
+        operation is performed. It will receive an error message or null if the
+        credential creation/update succeeded.
+    */
+    updateCloudCredential: function(tag, authType, attrs, callback) {
+      // Decorate the user supplied callback.
+      const handler = function(userCallback, data) {
+        if (!userCallback) {
+          console.log(
+            'data returned by Cloud.UpdateCredentials API call:', data);
+          return;
+        }
+        if (data.error) {
+          userCallback(data.error);
+          return;
+        }
+        const results = data.response.results;
+        if (!results || results.length !== 1) {
+          // This should never happen.
+          userCallback(
+            'invalid results returned by Juju: ' + JSON.stringify(results));
+          return;
+        }
+        const err = results[0].error && results[0].error.message;
+        if (err) {
+          userCallback(err);
+          return;
+        }
+        userCallback(null);
+      }.bind(this, callback);
+      // Send the API request.
+      const credentials = [{
+        tag: tag,
+        credential: {'auth-type': authType || '', attrs: attrs || {}}
+      }];
+      this._send_rpc({
+        type: 'Cloud',
+        request: 'UpdateCredentials',
+        params: {credentials: credentials}
+      }, handler);
+    },
+
+    /**
+      Revoke the cloud credential with the given tag.
+
+      @method revokeCloudCredential
+      @param {String} tag The Juju tag of the credential as a string, for
+        instance "cloudcred-google_dalek@local_google". Tags for credentials
+        are usually retrieved by calling the getTagsForCloudCredentials method.
+      @param {Function} callback A callable that must be called once the
+        operation is performed. It will receive an error message or null if the
+        credential revocation succeeded.
+    */
+    revokeCloudCredential: function(tag, callback) {
+      // Decorate the user supplied callback.
+      const handler = function(userCallback, data) {
+        if (!userCallback) {
+          console.log(
+            'data returned by Cloud.RevokeCredentials API call:', data);
+          return;
+        }
+        if (data.error) {
+          userCallback(data.error);
+          return;
+        }
+        const results = data.response.results;
+        if (!results || results.length !== 1) {
+          // This should never happen.
+          userCallback(
+            'invalid results returned by Juju: ' + JSON.stringify(results));
+          return;
+        }
+        const err = results[0].error && results[0].error.message;
+        if (err) {
+          userCallback(err);
+          return;
+        }
+        userCallback(null);
+      }.bind(this, callback);
+      // Send the API request.
+      this._send_rpc({
+        type: 'Cloud',
+        request: 'RevokeCredentials',
+        params: {entities: [{tag: tag}]}
+      }, handler);
     }
 
   });
--- a/jujugui/static/gui/src/app/store/env/sandbox.js
+++ b/jujugui/static/gui/src/app/store/env/sandbox.js
@@ -26,9 +26,9 @@ Sandbox APIs mimicking communications with the Go and Python backends.
 
 YUI.add('juju-env-sandbox', function(Y) {
 
-  var sandboxModule = Y.namespace('juju.environments.sandbox');
-  var models = Y.namespace('juju.models');
-  var CLOSEDERROR = 'INVALID_STATE_ERR : Connection is closed.';
+  const sandboxModule = Y.namespace('juju.environments.sandbox');
+  const models = Y.namespace('juju.models');
+  const CLOSEDERROR = 'INVALID_STATE_ERR : Connection is closed.';
 
   /**
   A client connection for interacting with a sandbox environment.
@@ -217,6 +217,11 @@ YUI.add('juju-env-sandbox', function(Y) {
     initializer: function() {
       this.connected = false;
       this.wsFailureCount = 0;
+      // Define an in-memory storage for cloud credentials. When the storage is
+      // initialized it holds an object mapping credential tags with the
+      // corresponding credential info. Do not access this attribute directly,
+      // use the _getCredentialsStorage method below.
+      this._cloudCredentialsStorage = null;
     },
 
     /**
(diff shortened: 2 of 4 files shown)


frankban:
QA:
both sandbox and real (in this case lxd, please adjust for your provider type):
- [ ] run `app.controllerAPI.updateCloudCredential('bad-wolf', 'empty', {}, err => {console.log('err:', err)});` and check that the callback reports an error (*bad-wolf is not a valid tag* or similar);
- [ ] run `app.controllerAPI.updateCloudCredential('cloudcred-lxd_admin@local_test', 'empty', {}, err => {console.log('err:', err)});`. This call should return null, indicating no errors occurred and the credential was added.
- [ ] retrieve the newly created credential information with `app.controllerAPI.getCloudCredentials(['cloudcred-lxd_admin@local_test'], console.log);` and check that the data is what you specified when adding the credential;
- [ ] update the credential with `app.controllerAPI.updateCloudCredential('cloudcred-lxd_admin@local_test', 'empty', {answer: '42'}, err => {console.log('err:', err)});` and check a null error is returned;
- [ ] use `app.controllerAPI.getCloudCredentials(['cloudcred-lxd_admin@local_test'], console.log);` again to check that the new *answer* attribute is now included in the result;
- [ ] revoke the new credential by running `app.controllerAPI.revokeCloudCredential('cloudcred-lxd_admin@local_test', err => {console.log('err:', err)});` and check that the error is null;
- [ ] double check the credential is no longer there with `app.controllerAPI.getCloudCredentials(['cloudcred-lxd_admin@local_test'], console.log);`: this time the results should include a *not found* error for the given tag;
- [ ] try to revoke that credential again with `app.controllerAPI.revokeCloudCredential('cloudcred-lxd_admin@local_test', err => {console.log('err:', err)});` and ensure that, even if that record no longer exists in Juju, no errors are returned;
- [ ] finally, try to remove a credential with an invalid tag, like `app.controllerAPI.revokeCloudCredential('bad-wolf', err => {console.log('err:', err)});` and check that an error is returned.

Thank you!

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5738/
Test PASSed.

jujugui/static/gui/src/app/store/env/controller-api.js
```diff
@@ -973,6 +973,104 @@ YUI.add('juju-controller-api', function(Y) {
...
+    */
+    updateCloudCredential: function(tag, authType, attrs, callback) {
+      // Decorate the user supplied callback.
+      const handler = function(userCallback, data) {
+        if (!userCallback) {
+          console.log(
+            'data returned by Cloud.UpdateCredentials API call:', data);
+          return;
+        }
+        if (data.error) {
```
kadams54: This will err if data is undefined or null.

jujugui/static/gui/src/app/store/env/controller-api.js
```diff
@@ -973,6 +973,104 @@ YUI.add('juju-controller-api', function(Y) {
...
+          userCallback(data.error);
+          return;
+        }
+        const results = data.response.results;
```
kadams54: May need validation for type safety: `const results = data.response && data.response.results`

jujugui/static/gui/src/app/store/env/controller-api.js
```diff
@@ -973,6 +973,104 @@ YUI.add('juju-controller-api', function(Y) {
...
+        if (!results || results.length !== 1) {
+          // This should never happen.
+          userCallback(
+            'invalid results returned by Juju: ' + JSON.stringify(results));
```
kadams54: Stringifying here could produce a parse error if results is not valid JSON. Better to do that in a try/catch.

jujugui/static/gui/src/app/store/env/controller-api.js
frankban: Are you sure? AFAIK !!undefined and !!null are both false.

jujugui/static/gui/src/app/store/env/controller-api.js
frankban: Yeah, this is a strong assumption that if Juju does not return an error then it does return a response.

jujugui/static/gui/src/app/store/env/controller-api.js
frankban: I think you mean JSON.parse arguments should be valid JSON. I think strigify accepts quite everything and only errors out for recursive objects, which are impossible in this case, as we are just marshaling a value obtained by a previous JSON.parse (the data returned by Juju).

bac:
QA was good. Ran through your checklist in sandbox and LXD.

FWIW the => syntax is not supported by the console in Safari.

jujugui/static/gui/src/app/store/env/controller-api.js
```diff
@@ -973,6 +973,104 @@ YUI.add('juju-controller-api', function(Y) {
...
+          console.log(
+            'data returned by Cloud.UpdateCredentials API call:', data);
+          return;
+        }
+        if (data.error) {
+          userCallback(data.error);
+          return;
+        }
+        const results = data.response.results;
+        if (!results || results.length !== 1) {
```
kadams54: `if (!results || (results.length && results.length !==1)`

jujugui/static/gui/src/app/store/env/controller-api.js
```diff
@@ -973,6 +973,104 @@ YUI.add('juju-controller-api', function(Y) {
...
+          userCallback(
+            'invalid results returned by Juju: ' + JSON.stringify(results));
+          return;
+        }
+        const err = results[0].error && results[0].error.message;
+        if (err) {
+          userCallback(err);
+          return;
+        }
+        userCallback(null);
```
kadams54: Why null? I was expecting to see `userCallback(results)` somewhere.

jujugui/static/gui/src/app/store/env/controller-api.js
```diff
@@ -973,6 +973,104 @@ YUI.add('juju-controller-api', function(Y) {
...
+    */
+    revokeCloudCredential: function(tag, callback) {
+      // Decorate the user supplied callback.
+      const handler = function(userCallback, data) {
+        if (!userCallback) {
+          console.log(
+            'data returned by Cloud.RevokeCredentials API call:', data);
+          return;
+        }
+        if (data.error) {
```
kadams54: Same type safety concerns as above.

jujugui/static/gui/src/app/store/env/sandbox.js
```diff
@@ -1346,15 +1353,99 @@ YUI.add('juju-env-sandbox', function(Y) {
...
     /**
+      Handle Cloud.UpdateCredentials messages.
+
+      @method handleCloudUpdateCredentials
+      @param {Object} data The contents of the API arguments.
+      @param {Object} client The active ClientConnection.
+      @param {Object} state An instance of FakeBackend.
+    */
+    handleCloudUpdateCredentials: function(data, client, state) {
+      const entries = data.params.credentials;
```
kadams54: Will err if `data` is null/undefined, or doesn't have `params`, or `params` doesn't have `credentials`.

jujugui/static/gui/src/app/store/env/sandbox.js
```diff
@@ -1346,15 +1353,99 @@ YUI.add('juju-env-sandbox', function(Y) {
...
+      if (!entries.length) {
```
kadams54: Will err if `entries` is null/undefined.

jujugui/static/gui/src/app/store/env/sandbox.js
```diff
@@ -1346,15 +1353,99 @@ YUI.add('juju-env-sandbox', function(Y) {
...
+    /**
+      Handle Cloud.RevokeCredentials messages.
+
+      @method handleCloudRevokeCredentials
+      @param {Object} data The contents of the API arguments.
+      @param {Object} client The active ClientConnection.
+      @param {Object} state An instance of FakeBackend.
+    */
+    handleCloudRevokeCredentials: function(data, client, state) {
+      const entities = data.params.entities;
```
kadams54: See above for type safety comments.

kadams54:
@frankban The code looks good, just may need some tweaking to be more resilient. Once those are in place, this will be a +1.

jujugui/static/gui/src/app/store/env/controller-api.js
kadams54: undefined or null won't have a error property, so: `if (data && data.error)`.

jujugui/static/gui/src/app/store/env/controller-api.js
kadams54: Ah yes, you are correct.

jujugui/static/gui/src/app/store/env/controller-api.js
frankban: Ah! Sorry I misread what you suggested. Yes in this case data cannot be undefined or null. If it was, that condition would have been catched up in the stack. This callback is called with this data here because data['request-id'] matches the request id of the original call. So we already know that data is an object.

jujugui/static/gui/src/app/store/env/controller-api.js
frankban: But IIUC the check would fail if results is not an array. The original check instead catches this fact, so an error would be returned every time results.length !== 1, so even in the case results.length is undefined or null.

jujugui/static/gui/src/app/store/env/controller-api.js
frankban: As described in the docstring, the callback only receives an error in this case. There is no valuable data in results (i.e. it is empty if everything went well, and we already parse possible errors).

jujugui/static/gui/src/app/store/env/controller-api.js
frankban: See above, at this point data must be an object.

jujugui/static/gui/src/app/store/env/sandbox.js
frankban: Yes, but since the GUI itself sends the request (which is being parsed here) we can assume request params are properly built with 100% safety.

jujugui/static/gui/src/app/store/env/sandbox.js
frankban: Ditto.

jujugui/static/gui/src/app/store/env/sandbox.js
frankban: As mentioned, in this case this is safe, as we are just parsing the request prepared in controllerAPI.js.

frankban:
Hey, Kyle and Brad,

thank you both for the reviews.
Kyle, thanks for making me think about all the implicit assumptions on the data structures we receive and for focusing on reliability of the code logic. I think that's the right mindset when tackling API code.

:shipit:

jujugui:

PR #1972, opened 2016-09-06
get clouds and credentials from the controller API

huwshimi:


huwshimi:
QA: not possible at this time.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5744/
Test PASSed.

jujugui/static/gui/src/app/components/deployment-flow/credential/credential.js
```diff
@@ -77,30 +75,57 @@ YUI.add('deployment-credential', function() {
       @method _getCredentials
     */
     _getCredentials: function() {
-      this.credentialXHR = this.props.listTemplates(
-        this._getCredentialsCallback);
+      const cloud = this.props.cloud;
+      const user = this.props.users.jem.user;
+      if (user) {
+        this.props.getTagsForCloudCredentials(
+          [[user, `cloud-${cloud}`]],
```
frankban: Cloud tags are already provided by the server. Why are we recalculating them from cloud names here? I know that the loic is quite easy and this will probably work forever, but the code would feel more solid if we just used the tags returned from the server.

jujugui/static/gui/src/app/components/deployment-flow/credential/credential.js
```diff
@@ -77,30 +75,57 @@ YUI.add('deployment-credential', function() {
...
+      @param {Array} tags A list of the tags found.
+    */
+    _getTagsCallback: function(error, tags) {
+      if (error) {
+        console.error('Unable to get tags for credentials', error);
+        return;
+      }
+      const tagList = [];
+      tags.forEach(tag => {
+        tagList = tagList.concat(tag.tags);
```
frankban: Probably a dumb question, but don't we need credentials based on the selected user/cloud pair? ISTM that you are flattening them here.

frankban:
I just have a couple of questions, and I'll leave a more deep review to others for now.
Thank you for this!

jujugui/static/gui/src/app/components/deployment-flow/credential/credential.js
huwshimi: If you get back an array of objects containing objects with arrays of credential tags, how do you know which are the ones you want for this cloud? We provided the user and cloud tags when we called `getTagsForCloudCredentials` so I assumed that only the creds for this user/cloud pair would have been returned. But I'm not sure!

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5757/
Test PASSed.

jujugui/static/gui/src/app/components/deployment-flow/credential/credential.js
frankban: From the docstring of `getTagsForCloudCredentials`: "in the sequence of results each result refers to the corresponding provided user/cloud pair". This is probably not clear enough, and that's my fault, but the meaning is that if you have a call like getTagsForCloudCredentials([pair1, pair2, pair3]) the callback will receive an array of results like [{tags: tags for pair1}, {tags: tags for pair2}, {tags: tags for pair3}], so each results refers to the user/cloud pair at that same index in the input tags. I guess this is less important now because we will be passing a single user/cloud pair, but the code should make that explicit and justify the flattening IMHO.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5758/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5759/
Test PASSed.

hatched:
:+1: Thanks for this @huwshimi 
:shipit:

jujugui:

PR #1973, opened 2016-09-06
Handle new ACLs introduced by beta18.

frankban:
Also extend the ACL interface so that it can be used in
the future when it will be required to do more fine grained
permission checks, like "can the current user create models?"
or "can they share this model?".

**Warning: do not land this branch before GUI 2.1.11 is released.**

### DIFF
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -418,7 +418,6 @@ YUI.add('juju-gui', function(Y) {
         ecs: ecs,
         user: this.get('user'),
         password: this.get('password'),
-        readOnly: this.get('readOnly'),
         conn: this.get('conn'),
         jujuCoreVersion: this.get('jujuCoreVersion')
       };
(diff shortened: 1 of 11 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5746/
Test PASSed.

frankban:
QA:

Run sandbox mode: it should work as usual, and you should be able to deploy charms/bundles, etc.
Also run gijoe (compiled with Juju tip, beta17 does not include the change being tested here): it should work as usual, and you should be able to deploy charms/bundles, etc.
Before destroying the controller, do the following to
check that the GUI works well in read-only mode:
- run `juju add-user who`;
- run `juju change-user-password who` and select a password for the new user;
- run `juju grant who read default` to grant read-only access to the default model to the who user;
- logout and log into the GUI again as user who, and check that the GUI is in read-only mode, meaning, for instance, that you cannot deploy charms or change options for existing applications in the canvas.

anthonydillon:
Code looks good. Does this mean all the components should update the readOnly prop?

frankban:
Component use isReadOnly already, and that API didn't change, only the internal implementation. This means that components can take advantage of the extended API when relevant, for instance the "create new model" component should check `acl.canAddModels()` rather than `acl.isReadOnly()`.

anthonydillon:
Will QA this evening if no one else picks it up.

hatched:
Thanks @frankban this works great

:shipit:

jujugui:

PR #1976, opened 2016-09-06
Fix Juju 1.25 console errors.

### LINKED ISSUE #1974: Juju 1.25 console errors
When logging in:
- Uncaught TypeError: Cannot read property 'setCredentials' of undefined
 - This is caused by https://github.com/juju/juju-gui/blob/develop/jujugui/static/gui/src/app/app.js#L857 sometimes getting passed an undefined value for an api. This is unavoidable so this loop simply needs to check if `api` is undefined and if so then continue on the loop.

- Warning: Failed prop type: Required prop `listModels` was not specified in 'Constructor'. in Constructor
 - It looks like we have `listModels` listed as a required prop in the model switcher but in Juju 1 we do not have list models support.

kadams54:
Fixes #1974.

### DIFF
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -845,20 +845,26 @@ YUI.add('juju-gui', function(Y) {
       credentials, useMacaroons, apis=[this.env, this.controllerAPI]) {
       if (useMacaroons) {
         apis.forEach(api => {
-          api.loginWithMacaroon(new Y.juju.environments.web.Bakery({
-            webhandler: new Y.juju.environments.web.WebHandler(),
-            interactive: true,
-            serviceName: 'juju',
-            dischargeStore: window.localStorage
-          }), this._apiLoginHandler.bind(this, api));
+          // The api may be unset if the model does not support it.
+          if (api) {
+            api.loginWithMacaroon(new Y.juju.environments.web.Bakery({
+              webhandler: new Y.juju.environments.web.WebHandler(),
+              interactive: true,
+              serviceName: 'juju',
+              dischargeStore: window.localStorage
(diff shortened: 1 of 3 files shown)


jujugui/static/gui/src/app/app.js
```diff
@@ -845,20 +845,24 @@ YUI.add('juju-gui', function(Y) {
       credentials, useMacaroons, apis=[this.env, this.controllerAPI]) {
       if (useMacaroons) {
         apis.forEach(api => {
-          api.loginWithMacaroon(new Y.juju.environments.web.Bakery({
-            webhandler: new Y.juju.environments.web.WebHandler(),
-            interactive: true,
-            serviceName: 'juju',
-            dischargeStore: window.localStorage
-          }), this._apiLoginHandler.bind(this, api));
+          if (api) {
```
hatched: Could you add some comments here as to why there might not be an api

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5752/
Test PASSed.

jujugui/static/gui/src/app/app.js
hatched: It's possible that this method is called without specifying the apis. At which point the default values are used. It is however possible at that point that there is no controllerAPI instance because we are in a Juju model which does not support it. This causes the apis array to be `[this.env, undefined]` causing the login's to fail.

hatched:
QA O K !

kadams54:
Per @hatched and I QAing in both Juju 1 & Juju 2, it looks good so... :shipit:

jujugui:

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5753/
Test PASSed.

Issue #1987, opened 2016-09-09
GUI connects to supplied jujuEnvUUID even if user-models is empty/null


hatched:
When the GUI is served up by Juju it includes with it a `jujuEnvUUID` in the config which the GUI immediately uses and tries to connect to that model along side the controller when logging in as an external user via SSO. This happens even if the controller user model list returns returns null or other models. 

This means that the GUI actually connects and then successfully logs into this model but does not have permissions and as such cannot perform any actions.

To remedy this, we shouldn't attempt to log the user into any models, even if one is provided in the `jujuEnvUUID` until the controller connection successfully returns the model list. If at that point it contains the uuid supplied in `jujuEnvUUID` then connect to it.

PR #1992, opened 2016-09-12
Clean up the login process.

frankban:
Fix a bug making anonymous user unsuccessfylly attempt authenticated calls.
Also simplify the login event and make it consistent across all GUI backends.

Note that this require a follow u branch to solve other login problems described at https://github.com/juju/juju-gui/issues/1987

### DIFF
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -585,8 +585,8 @@ YUI.add('juju-gui', function(Y) {
             // minutes of this writing.
             var querystring = this.location.search.substring(1);
             var qs = Y.QueryString.parse(querystring);
-            var authtoken = qs.authtoken;
-            if (Y.Lang.isValue(authtoken)) {
+            var authtoken = qs.authtoken || '';
+            if (authtoken || authtoken.length) {
               // De-dupe if necessary.
               if (Array.isArray(authtoken)) {
                 authtoken = authtoken[0];
@@ -719,7 +719,11 @@ YUI.add('juju-gui', function(Y) {
       controllerAPI.setAttrs({ user, password });
       controllerAPI.setCredentials({ user, password, macaroons });
 
-      controllerAPI.after('login', e => {
+      controllerAPI.after('login', evt => {
+        if (evt.err) {
+          this._renderLogin(evt.err);
+          return;
+        }
         // After logging in trigger the app to dispatch to re-render the
         // components that require an active connection to the controllerAPI.
         this.dispatch();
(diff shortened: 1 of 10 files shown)


frankban:
QA: log into the GUI, with both credentials and USSO.
Also try providing wrong credentials/no credentials.
Thank you!

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5785/
Test PASSed.

jujugui/static/gui/src/app/app.js
```diff
@@ -2042,64 +2041,65 @@ YUI.add('juju-gui', function(Y) {
...
+      // Handle the change set token if provided in the query.
+      // The change set token identifies a collections of changes required
+      // to deploy a bundle. Those changes are assumed to be already
+      // registered in the GUI server (via a ChangeSet:SetChanges request).
+      // Doing that is usually responsibility of a separate system
+      // (most of the times, it is Juju Quickstart).
+      var querystring = this.location.search.substring(1);
+      var qs = Y.QueryString.parse(querystring);
+      var changesToken = qs.changestoken;
+      if (Y.Lang.isValue(changesToken)) {
```
hatched: Would be nice to de-YUI this

hatched:
Thanks for this @frankban I was able to log in as usual with the username and password however when using the USSO login option it appeared to still require a username and password inputted or else it fails with an error:

"authentication failed: invalid entity name or password"

frankban:
@hatched that's weird, I am able to login with USSO without creedntials with this branch.
Did you grant the user login to the controller and some kind of access to one model?

jujugui/static/gui/src/app/app.js
frankban: Done.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/5792/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5793/
Test PASSed.

hatched:
After revisiting this it appears that I must have improperly set up the controller initially causing the previous error. However I now get a new error.

When clicking the Login with USSO option a login error is shown:

`facade "Admin" not supported for controller API connection`

This is because when clicking the USSO button a Login rpc call is made:
`{"type":"Admin","request":"Login","version":3,"request-id":7,"params":{}}`
With the response:
`{"request-id":7,"error":"facade \"Admin\" not supported for controller API connection","error-code":"not supported","response":{}}`
At which point I get the warning about a popup. After permitting the popup and logging in, I'm landed on the canvas as expected without performing any more actions in the GUI tab.

These popups are going to be troublesome going forward. I feel like if the user is going to click the USSO button on the login page we can safely redirect them to the USSO login and then redirect back to the GUI.

frankban:
This is great feedback Jeff! And it allowed me to find the cause of that message and the bad behavior. I bet you encountered that after logging in and then out and then trying the USSO login. That's because when logging out from the GUI we were not actually logging out from the controller, only from the model (which means that the current release of the GUI has a potential security hole!). Therefore, when logging in again, the Admin request was sent through an already authenticated connection, and hence the "facade \"Admin\" not supported" error. I am going to merge this now as this is now a security fix.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5805/
Test PASSed.

frankban:
:shipit:

jujugui:

PR #1998, opened 2016-09-13
Add version.js to a jujugui dist.

### LINKED ISSUE #1986: version for the currently used GUI is not visible
There used to be a method to get the version of the GUI you were looking at by going to the  /juju-ui/version.js url and checking the info in that file. 

When using the GUI in juju:

    juju gui

The url that serves the root GUI is something like:

   https://10.90.136.204:17070/gui/83b3a588-aa29-4d79-847f-ff4b0840bb0f/

The juju-ui/version.js file path does not work to get the version info. I'm aware that there's a gui-archive url that shows all versions of the GUI available and which one is supposed to be shipped to users, however I still think that we need a method of finding out "what is this GUI" because what's intended doesn't always turn out to be what happens. This is useful as I was trying to figure out how to validate that my upgrade-gui command succeeded and that I was getting a new GUI and not cached JS or the like.

kadams54:
Adds a version.js asset to the tree which constains the GUI version and the git sha. It can be fetched by visiting: `/gui/[CONTROLLER UUID]/[MODEL UUID]/static/gui/build/app/version.json`

Fixes #1986.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,9 @@ LSB_RELEASE = $(shell lsb_release -cs)
 SYSDEPS = coreutils g++ git inotify-tools nodejs \
 	python-virtualenv realpath xvfb chromium-browser
 
+CURRENT_VERSION = $(shell sed -n -e '/current_version =/ s/.*\= *// p' .bumpversion.cfg)
+CURRENT_COMMIT = $(shell git rev-parse HEAD)
+
 .PHONY: help
 help:
 	@echo "bumpversion - bump version."
@@ -329,17 +332,15 @@ ci-check: clean-downloadcache deps fast-babel check
 bumpversion: deps
 	bin/bumpversion $(VPART)
 
-CURRENT_VERSION = $(shell sed -n -e '/current_version =/ s/.*\= *// p' .bumpversion.cfg)
-CURRENT_COMMIT = $(shell git rev-parse HEAD)
 .PHONY: version
 version:
-	echo '{ "version": "$(CURRENT_VERSION)", "commit": "$(CURRENT_COMMIT)" }' > $(GUIBUILD)/app/version.json
+	echo "'version: $(CURRENT_VERSION)';\n'commit: $(CURRENT_COMMIT)';" > $(GUIBUILD)/app/version.js
 
 .PHONY: dist
 dist: clean-all fast-dist
 
 .PHONY: fast-dist
-fast-dist: deps fast-babel gui test-deps collect-requirements
+fast-dist: deps fast-babel gui test-deps collect-requirements version
 	python setup.py sdist --formats=bztar\
 
 #######
--- a/Makefile
+++ b/Makefile
@@ -329,9 +329,11 @@ ci-check: clean-downloadcache deps fast-babel check
 bumpversion: deps
 	bin/bumpversion $(VPART)
 
+CURRENT_VERSION = $(shell sed -n -e '/current_version =/ s/.*\= *// p' .bumpversion.cfg)
+CURRENT_COMMIT = $(shell git rev-parse HEAD)
 .PHONY: version
 version:
-	python setup.py --version | sed -e "s/.*/'&'/" > $(GUIBUILD)/app/version.js
+	echo '{ "version": "$(CURRENT_VERSION)", "commit": "$(CURRENT_COMMIT)" }' > $(GUIBUILD)/app/version.json
 
 .PHONY: dist
 dist: clean-all fast-dist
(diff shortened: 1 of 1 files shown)


kadams54:
Right now the version.js is only accessible via the combo loader, so the URL gets ugly fast. For example:

https://107.20.128.188:17070/gui/3e39a40e-58e5-4b40-80fd-36da5aac811e/b6f9b8b96c392cbf609b4431c51aa33753ab3e485c9a7878fdac467af018f368/combo?app/version.js

I'd like to get the file exposed at a slightly nicer route, if possible. Ideas?

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5795/
Test PASSed.

kadams54:
QA

1. Deploy GIJoe.
2. Navigate to `/gui/3e39a40e-58e5-4b40-80fd-36da5aac811e/b6f9b8b96c392cbf609b4431c51aa33753ab3e485c9a7878fdac467af018f368/combo?app/version.json`.
3. You should see the version and the current commit hash.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5880/
Test PASSed.

Makefile
```diff
@@ -329,6 +329,12 @@ ci-check: clean-downloadcache deps fast-babel check
 bumpversion: deps
 	bin/bumpversion $(VPART)
 
+CURRENT_VERSION = $(shell sed -n -e '/current_version =/ s/.*\= *// p' .bumpversion.cfg)
+CURRENT_COMMIT = $(shell git rev-parse HEAD)
+.PHONY: version
+version:
```
hatched: Can we have this generated and inserted into the dists as well plz.

Makefile
```diff
@@ -329,6 +329,12 @@ ci-check: clean-downloadcache deps fast-babel check
 bumpversion: deps
 	bin/bumpversion $(VPART)
 
+CURRENT_VERSION = $(shell sed -n -e '/current_version =/ s/.*\= *// p' .bumpversion.cfg)
+CURRENT_COMMIT = $(shell git rev-parse HEAD)
```
hatched: While these are only used here right now I could see them being useful in the near future in future additions, could you move them to the other definitions please.

kadams54: Done.

Makefile
kadams54: Done.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5887/
Test PASSed.

Makefile
```diff
@@ -329,11 +332,15 @@ ci-check: clean-downloadcache deps fast-babel check
 bumpversion: deps
 	bin/bumpversion $(VPART)
 
+.PHONY: version
+version:
+	echo "'version: $(CURRENT_VERSION)';\n'commit: $(CURRENT_COMMIT)';" > $(GUIBUILD)/app/version.js
```
hatched: Why was this changed from json to js? And invalid js at that?

hatched:
The version file was available but it's content was weird.

```
'version: 2.1.13';
'commit: 1f368802478a23207e94e23b0ad738fc0766dfb7';

/* version.js */
```

Makefile
kadams54: Because the combo loader and build process handle JS just fine but balk at JSON. I spent some time trying to get JSON working but just threw my hands up in the end and switched back to JS. I remember `'foo bar'` used to be valid JS, which is why the strict mode invocation took the form it did (older JS engines wouldn't balk, ignoring it instead), but it seems ECMAScript 5 takes a more stringent view of bare strings in JS files. Meh.

kadams54:
But you were able to see the version and commit info, which is the crucial bit. The formatting of said info within the file is bike shedding at this point, IMO. What about this?

~~~~
const jujuguiVersionInfo = {
  version: '2.1.13',
  commit = '1f368802478a23207e94e23b0ad738fc0766dfb7'
};
~~~~

The version.js should never be loaded with the rest of the files, so I wouldn't expect variable collision to be an issue, but I also chose a name that ought to be sufficiently unique. The other option would be to try to track down all the spots in the build process where JSON needs to be included alongside JS and potentially patch the comboloader. I'd really rather not do that?

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5892/
Test PASSed.

hatched:
Thanks for this @kadams54 

:shipit:

jujugui:

PR #2008, opened 2016-09-15
Actually deploy the model (hopefully, maybe, probably not)

huwshimi:


jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/5824/
Test FAILed.

huwshimi:
QA:
- deploy the gui with a real juju (using tip if this is fixed: https://bugs.launchpad.net/juju/+bug/1623808)
- load `/:flags:/blues/`
- click on your username to open the profile
- create a new model by clicking the 'Create new' button
- add a service
- click 'Deploy changes'
- if you'd like you can add a credential,
- scroll down and click 'Deploy'
- if this bug is fixed: https://bugs.launchpad.net/juju/+bug/1623808 you might have a deployed model otherwise you'll get an error

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5825/
Test PASSed.

jujugui/static/gui/src/app/app.js
```diff
@@ -1134,6 +1133,7 @@ YUI.add('juju-gui', function(Y) {
           name: cloud
         };
       }
+      const user = this.controllerAPI.get('user');
```
frankban: Interesting, what puts the user in the connection instances? Should we use `this.controllerAPI.getCredentials().user` instead?

jujugui/static/gui/src/app/views/utils.js
```diff
@@ -1550,7 +1550,7 @@ YUI.add('juju-view-utils', function(Y) {
 
     @method deploy
     @param {Object} env Reference to the app env.
-    @param {Object} jem Reference to jem.
+    @param {Object} controllerAPI Reference to the controller api.
```
frankban: Well, in JS they are all references to something, correct? ;-) What do you think about `The controller connection as an instance of juju.ControllerAPI.` ?

jujugui/static/gui/src/app/views/utils.js
```diff
@@ -1577,15 +1577,18 @@ YUI.add('juju-view-utils', function(Y) {
       return;
     }
 
-    jem.newModel(
-      users.jem.user,
+    controllerAPI.createModel(
       model,
-      credential,
+      // Check that this is a user tag, if not turn it into one.
```
frankban: I'd like for this logic to be elsewhere, and for functions to clearly define what they receive. A user name and a user tag are two completely different kind of values, and we should be clear here about what's expected, and callers are responsible of providing the right value (either a tag or a name), otherwise we'll end up putting this logic everywhere.

jujugui/static/gui/src/app/views/utils.js
```diff
@@ -1565,8 +1565,8 @@ YUI.add('juju-view-utils', function(Y) {
     @param {String} region The cloud region to deploy to.
   */
   utils.deploy = function(
-    env, jem, users, autoPlaceUnits, createSocketURL, appSet, committed,
-    callback, autoplace=true, model, credential, cloud, region) {
+    env, controllerAPI, autoPlaceUnits, createSocketURL, appSet, committed,
```
frankban: It seems to me that this function really receives too many arguments, and I believe that, when this function is called, understanding the arguments would be pretty hard. I'd like to refactor this so that it takes an object, with clearly defined and documented fields, and some documented default fallbacks in case the field is not specified. That would also be more future proof in cae we want to add another field.

jujugui/static/gui/src/app/views/utils.js
```diff
@@ -1577,15 +1577,18 @@ YUI.add('juju-view-utils', function(Y) {
...
-      users.jem.user,
+    controllerAPI.createModel(
       model,
-      credential,
+      // Check that this is a user tag, if not turn it into one.
+      cloud.indexOf('user-') === 0 ? user : `user-${user}`,
       {
-        cloud: cloud,
+        // Check that this is a credential tag, if not turn it into one.
+        credentialTag: credential.indexOf('cloudcred-') === 0 ?
```
frankban: Same, let's say we receive a credential tag, and let's ensure we always pass tags, and let's remove this logic please.

jujugui/static/gui/src/app/views/utils.js
```diff
@@ -1577,15 +1577,18 @@ YUI.add('juju-view-utils', function(Y) {
...
+          credential : `cloudcred-${credential}`,
+        // Check that this is a cloud tag, if not turn it into one.
+        cloudTag: cloud.indexOf('cloud-') === 0 ? cloud : `cloud-${cloud}`,
```
frankban: Ditto. This demonstrates that this function is much more complex and it is all incidental complexity, so something we can definitely remove.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/5832/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5838/
Test PASSed.

hatched:
Thanks for all this work @huwshimi 

Works as expected! 

:shipit:

jujugui:

PR #2018, opened 2016-09-19
Improve login/logout handling.

frankban:
When logging out now WebSocket connections are properly closed, and the API instances (controller and model) are properly cleaned up in order to avoid any information leak from theoretically logged out connections and from the in-memory database.

Included in this work, a bug prevening login/logout/login again has been fixed by correctly handling the existing credentials when logging out and then logging in again. For instance credentials are now explicitly shared when switching to another model.

This branch also includes extensive sanity improvements to our test suite, especially the old app tests. As part of those improvements, a bug causing the app to dispatch after it's destroyed is also fixed, by cancelling the timer for the method called when the database receives changes.

The _cleanups machinery for the testing suite has been fixed so that at least:
- _cleanups registered functions are called before afterEach (and not after);
- afterEach functions can now use done without the usual "done called more than once" issue (one done was called by the wrapped function, and another from the monkey patched newAfterEach).

Finally, attenuate the horror in the app tests.

### DIFF
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -1842,6 +1842,9 @@ YUI.add('juju-gui', function(Y) {
     @method destructor
     */
     destructor: function() {
+      // Clear the database handler timer. Without this, the application could
+      // dispatch after it is destroyed, resulting in a dirty state and bugs
+      // difficult to debug, so please do not remove this code.
       if (this.dbChangedTimer) {
         clearTimeout(this.dbChangedTimer);
       }
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -582,9 +582,13 @@ YUI.add('juju-gui', function(Y) {
           // Try a token login.
           this.env.tokenLogin(authtoken);
         }
-        // There are no credentials. Do nothing in this case as the controller
-        // login check will have kicked the user to the login prompt already
-        // and we can wait until they have provided the credentials there.
+        // There are no credentials. Do nothing in this case if we are on Juju
+        // >= 2 as the controller login check will have kicked the user to the
+        // login prompt already and we can wait until they have provided the
+        // credentials there. In Juju 1 though we need to display the login.
+        if (!this.controllerAPI) {
+          this._displayLogin();
+        }
       });
 
       // If the database updates, redraw the view (distinct from model updates).
@@ -764,7 +768,12 @@ YUI.add('juju-gui', function(Y) {
         });
       });
 
-      controllerAPI.after('connectedChange', e => {
+      controllerAPI.after('connectedChange', evt => {
+        if (!evt.newVal) {
+          // The controller is not connected, do nothing waiting for a
+          // reconnection.
+          return;
+        }
         const credentials = this.controllerAPI.getCredentials();
         if (!credentials.areAvailable && !this.get('gisf')) {
           this._displayLogin();
@@ -848,10 +857,15 @@ YUI.add('juju-gui', function(Y) {
       }
       apis.forEach(api => {
         // The api may be unset if the current Juju does not support it.
-        if (api && api.get('connected')) {
-          if (credentials) {
-            api.setCredentials(credentials);
-          }
+        if (!api) {
+          return;
+        }
+        if (credentials) {
+          // We set credentials even if the API is not connected: they will be
+          // used when the connection is eventually established.
+          api.setCredentials(credentials);
+        }
+        if (api.get('connected')) {
           console.log(`logging into ${api.name} with user and password`);
           api.login();
         }
@@ -1828,21 +1842,27 @@ YUI.add('juju-gui', function(Y) {
     @method destructor
     */
     destructor: function() {
+      if (this.dbChangedTimer) {
+        clearTimeout(this.dbChangedTimer);
+      }
       if (this.zoomMessageHandler) {
         this.zoomMessageHandler.detach();
       }
       if (this._keybindings) {
         this._keybindings.detach();
       }
-      Y.each(
-          [this.env, this.db, this.endpointsController, this.controllerAPI],
-          function(o) {
-            if (o && o.destroy) {
-              o.detachAll();
-              o.destroy();
-            }
-          }
-      );
+      const toBeDestroyed = [
+        this.env,
+        this.controllerAPI,
+        this.db,
+        this.endpointsController
+      ];
+      toBeDestroyed.forEach(obj => {
+        if (obj && obj.destroy) {
+          obj.detachAll();
+          obj.destroy();
+        }
+      });
       ['dragenter', 'dragover', 'dragleave'].forEach((eventName) => {
         document.removeEventListener(eventName, this._boundAppDragOverHandler);
       });
@@ -1867,11 +1887,12 @@ YUI.add('juju-gui', function(Y) {
      */
     on_database_changed: function(evt) {
       // This timeout helps to reduce the number of needless dispatches from
-      // upwards of 8 to 2. At least until we can move to the model bound views.
+      // upwards of 8 to 2. At least until we can move to the model bound
+      // views.
       if (this.dbChangedTimer) {
-        this.dbChangedTimer.cancel();
+        clearTimeout(this.dbChangedTimer);
       }
-      this.dbChangedTimer = Y.later(100, this, this._dbChangedHandler);
+      this.dbChangedTimer = setTimeout(this._dbChangedHandler.bind(this), 100);
       return;
     },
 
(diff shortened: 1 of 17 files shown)


frankban:
QA:
on Juju1, Juju2 (gijoe) and sandbox, try to use the GUI as usual, log in, log out, log in again, where possible try to log in with macaroons, log out, log in again etc.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5839/
Test PASSed.

jujugui/static/gui/src/app/app.js
```diff
@@ -764,7 +768,12 @@ YUI.add('juju-gui', function(Y) {
         });
       });
 
-      controllerAPI.after('connectedChange', e => {
+      controllerAPI.after('connectedChange', evt => {
+        if (!evt.newVal) {
```
hatched: I'm assuming that this is when the user logs out of the controller?
Could you add this example to the code comment plz.

jujugui/static/gui/src/app/app.js
```diff
@@ -1828,21 +1842,27 @@ YUI.add('juju-gui', function(Y) {
     @method destructor
     */
     destructor: function() {
+      if (this.dbChangedTimer) {
```
hatched: Yay thanks!
This definitely deserves a code comment as to why this is here and to never remove it :smile:

jujugui/static/gui/src/app/app.js
```diff
@@ -1931,12 +1952,31 @@ YUI.add('juju-gui', function(Y) {
...
+      };
+      const controllerAPI = this.controllerAPI;
+      if (controllerAPI) {
+        closeController = controllerAPI.close.bind(controllerAPI);
+      }
+      this.env.close(() => {
+        closeController(() => {
+          if (controllerAPI) {
+            // Juju 2 connects to the controller and gets models from there.
+            controllerAPI.connect();
```
hatched: On logging out from the env we want to reconnect to the controller again?

jujugui/static/gui/src/app/store/env/acl.js
```diff
@@ -43,48 +43,55 @@ YUI.add('acl', function(Y) {
...
-      @return {Boolean} Whether the user can destroy the current model.
-    */
-    canRemoveModel: () => modelAPI.get('modelAccess') === 'admin'
-
-  });
+  juju.generateAcl = (controllerAPI, modelAPI) => {
+    if (!controllerAPI) {
+      // The application is connected to a legacy Juju.
+      controllerAPI = {
+        get: (_) => ''
```
hatched: `() => ''`

jujugui/static/gui/src/app/store/env/acl.js
```diff
@@ -43,48 +43,55 @@ YUI.add('acl', function(Y) {
...
-
-      @function canRemoveModel
-      @return {Boolean} Whether the user can destroy the current model.
-    */
-    canRemoveModel: () => modelAPI.get('modelAccess') === 'admin'
-
-  });
+  juju.generateAcl = (controllerAPI, modelAPI) => {
+    if (!controllerAPI) {
+      // The application is connected to a legacy Juju.
```
hatched: Good catch

jujugui/static/gui/src/app/store/env/base.js
```diff
@@ -266,9 +266,13 @@ YUI.add('juju-env-base', function(Y) {
...
         this.ws = conn;
       } else {
-        this.ws = new jujulib.ReconnectingWebSocket(this.get('socket_url'));
+        const url = this.get('socket_url');
+        console.log('connecting to ' + url);
+        this.ws = new jujulib.ReconnectingWebSocket(url);
+        this._txn_callbacks = {};
       }
       this.ws.debug = this.get('debug');
       this.ws.onmessage = Y.bind(this.on_message, this);
```
hatched: If you wouldn't mind de yui'ing these :)

```JavaScript
this.ws.onmessage = this.on_message.bind(this);
...
```

jujugui/static/gui/src/app/store/env/base.js
```diff
@@ -294,30 +298,39 @@ YUI.add('juju-env-base', function(Y) {
...
+      calling the given callback that effectively closes the WebSocket
+      connection. Concrete implementations are assumed to be idempotent.
+
+      @method cleanup
+      @param {Function} done A callable that must be called by the function and
+        that actually closes the connection.
     */
-    beforeClose: function(callback) {
-      callback();
+    cleanup: function(done) {
```
hatched: I'm not sure about these `done` renamed...but up to you :)

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -697,25 +694,54 @@ YUI.add('juju-env-api', function(Y) {
...
+    cleanup: function(done) {
+      console.log('cleaning up the model API connection');
       if (this._pinger) {
         clearInterval(this._pinger);
         this._pinger = null;
       }
-      this._stopWatching(callback);
+      const callback = () => {
+        // TODO frankban: find a more automated way to clean up attributes.
+        this.setAttrs({
```
hatched: If you wanted to clear every attribute you could iterate over the attrs from `this.getAttrs()`

jujugui/static/gui/src/app/store/env/legacy-api.js
```diff
@@ -614,25 +611,47 @@ YUI.add('juju-env-legacy-api', function(Y) {
...
+    cleanup: function(done) {
+      console.log('cleaning up the legacy model API connection');
       if (this._pinger) {
         clearInterval(this._pinger);
         this._pinger = null;
       }
-      this._stopWatching(callback);
+      const callback = () => {
+        // TODO frankban: find a more automated way to clean up attributes.
+        this.setAttrs({
```
hatched: See above comment about clearing attrs

jujugui/static/gui/src/test/globalconfig.js
```diff
@@ -33,18 +33,24 @@ Mocha.Suite.prototype.beforeEach = function(title, fn) {
 };
 
 Mocha.Suite.prototype.afterEach = function(func) {
-  var newAfterEach = function(done) {
-    func.apply(this, arguments);
+  const newAfterEach = function(done) {
```
hatched: Thanks for this change, does this fix the "done called multiple times" error?

jujugui/static/gui/src/app/app.js
frankban: This is when the network disconnects (connected != authenticated). In here we want to react to changes to the connection only when the change means that it reconnected. We already have a comment there: 
`// The controller is not connected, do nothing waiting for a reconnection.`

jujugui/static/gui/src/app/app.js
frankban: More than a code comment, it deserves a test. But I'll also add a comment ;-)

jujugui/static/gui/src/app/app.js
frankban: Yes, because that's the only way the GUI can make it, the GUI is nothing without a controller connection, being it Juju, Jimm or the fake backend. Maybe you are thinking about authenticating? We are not logging in again, we are just reconnnecting the WebSocket here.

jujugui/static/gui/src/app/store/env/acl.js
frankban: I liked to be clear about the number of expected arguments. I know this is a JS feature, but it feels like a JS bug ;-)

jujugui/static/gui/src/app/store/env/base.js
frankban: Well I renamed those to make it more apparent that this is not something you could (or could not) call, like a usual callback. This is something that you are required to call in order to communicate that you are "done" with the clean up process. How does it sound?

jujugui/static/gui/src/app/store/env/api.js
frankban: I know, I am just not sure about clearing the attributes passed at instantiation time, and I guess there is no easy way of separating attrs defined with set/setAttrs with the ones passed when initializing the instance.

frankban: For a followup I am thinking about a function like this.setLoggedAttr(name, value, initial) that automatically restore the initial value during clean up.

jujugui/static/gui/src/test/globalconfig.js
frankban: It does.

jujugui/static/gui/src/app/store/env/base.js
frankban: Will do.

jujugui/static/gui/src/app/store/env/api.js
hatched: I was really hoping to be able to respond with only emoji but it looks like I actually have to use words this time :laughing: 

This would be a good feature to add to my new-yet-to-have-been-written db library.

jujugui/static/gui/src/app/app.js
frankban: Added the comment.

jujugui/static/gui/src/app/store/env/base.js
frankban: Done.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5842/
Test PASSed.

hatched:
Thanks for putting together all this @frankban this really fixes up the login mechanisms. 

As discussed. This breaks Juju 1 by not presenting the login screen but instead the "connecting to model" screen. 

:+1: :ok: as the Juju 1 bug will be resolved in a follow-up

hatched:
:shipit:

jujugui:

jujugui:
        build url: http://ci.jujugui.org:8080/job/juju-gui-merge/1909

hatched:
It appears that this failure might be spurious.

jujugui:

PR #2034, opened 2016-09-21
Temporarily supply the credentials when creating a model from the user profile page

### LINKED ISSUE #2027: Unable to deploy to new models on GCE
When creating a new model via the user profile page you won't be able to deploy any applications to this model because no credentials are being passed to GCE when creating the model. This prevents any instances from being brought up.

While this is resolved with the new DF the old one simply passed an empty object for args. As a temporary fix we have to fetch the tags, credentials, then supply the first credential in the list when creating the model.

hatched:
When creating a model via the user profile page it'll now pass the correct credentials down. While this won't quite fix 2027 as there is a Juju core component that also needs to be resolved. This ensures that once that fix lands it'll work correctly.

Fixes #2027


hatched:
#### To QA

In LXD and AWS switch to the user profile page and create a model.
Once you're back on the canvas, deploy an application and wait for the machine to spin up.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5859/
Test PASSed.

hatched:
:shipit: 
Shipping to move on with release.

jujugui:

PR #2035, opened 2016-09-22
API layer: improve Juju tags management.

### LINKED ISSUE #2013: env only has a credential tag
If you run `app.env.getAttrs()` you can see that there is a `credentialTag` but it would be nice if it just had the credential.

### LINKED ISSUE #2015: createModel expects tags
The controller `createModel` expects tags, it should turn the values into tags

### LINKED ISSUE #2016: getTagsForCloudCredentials expects tags
`getTagsForCloudCredentials` should turn the values into tags.

frankban:
Now tags are completely handled at the API level, which means that API consumers can ignore the existence of Juju tags.
This is a massive diff because it's mostly the result of grepping and replacing, and all changes must also land all together to avoid consistency issues.

So apologies, but as an attempt to help the reviewer, here is a detailed list of API changes:
```
tags object added to the base model, with tags.build and tags.parse methods

serverTag modelAPI and controllerAPI attrs
    renamed to controllerId
    is now a name, not a tag

modelTag modelAPI attr
    renamed to modelId
    is now a name, not a tag

credentialTag modelAPI and controllerAPI attrs
    renamed to just credential
    is now a name, not a tag

currentModelInfo
    callback: now receives the 2 params (err, result)
    callback: removed all the tags from the second argument

generateTag in modelAPI:
    removed, tags.build is used instead

modelInfo
    tags no longer received (replaced by ids)
    callback: in each returned info
        added id
        removed tag and ownerTag
        renamed serverUuid to controllerUUID

setCredentials / getCredentials now only receive and return user names, not user tags

listModelsWithInfo:
    callback: in each returned info 
        added id
        removed tag and ownerTag 
        renamed serverUuid to controllerUUID

createModel
    receives a user name instead of a tag;
    receives args.cloud instead of args.cloudTag
    receives args.credential instead of args.credentialTag
    callback: removed ownerTag, cloudTag, credentialTag

destroyModels
    receives ids instead of tags
    callback: it is passed a tuple (err, {}) now, and the second argument maps ids to errors, not tags to errors

listModels
    receives a user instead of a user tag
    callback: in each returned info, added id, removed tag and ownerTag

listClouds
    callback: the passed object no longer maps cloud tags to cloud info, it maps cloud names to cloud info instead, and the info value no longer includes the name, as it is now the key

getClouds
    it receives cloud names, not cloud tags
    callback: the passed object no longer maps cloud tags to cloud info, it maps cloud names to cloud info instead, and the info value no longer includes the name, as it is now the key

getTagsForCloudCredentials
    renamed to getCloudCredentialNames
    it receives (user, cloud) pairs, not (user-tag, cloud-tag) pairs
    it returns names, not tags

getCloudCredentials
    it receives credential names, not credential tags
    callback: the passed object no longer maps credential tags to credential info, it maps credential names to credential info instead, and the info value no longer includes the name, as it is now the key

updateCloudCredential
    it receives a name instead of a tag

revokeCloudCredential
    it receives a name instead of a tag
    
getDefaultCloudTag
    renamed to getDefaultCloudName
    callback: it receives a name as second argument, not a tag
```

Fixes #2013
Fixes #2015
Fixes #2016


frankban:
QA must be done on Juju2 (gijoe), sandbox mode and Juju 1.
Since this branch includes lots of API changes (especially related to models management, credentials, clouds and the authentication process) please test the GUI deeply, log in, log out, log in again, use macaroons, create new models from the GUI (where possible), deploy charms in these new models.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5863/
Test PASSed.

jujugui/static/gui/src/app/store/env/base.js
```diff
@@ -27,9 +27,50 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+    /**
+      Parse a Juju entity name from the given tag.
+
+      @method parse
+      @param {String} tag The tag associated with the Juju entity.
+      @param {String} prefix The Juju prefix (see module.prefixes).
+      @return {String} The resulting name of the entity (without the prefix).
+    */
+    parse: (type, tag) => {
+      const exp = new RegExp(`^${type}-`);
```
hatched: I wonder if using `indexof` would be more performant than setting up a regex instance

jujugui/static/gui/src/app/store/env/base.js
```diff
@@ -27,9 +27,50 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+      const name = tag.replace(exp, '');
+      if (name === tag) {
+        throw new Error(`invalid tag of type ${type}: ${tag}`);
```
hatched: Do we want to break the execution of the app if the tag type is invalid? I don't see any try/catches around these method uses.

frankban: This should never happen, so yes we want to know it happened, it should not fail silently, as it means something is wrong in the code.

jujugui/static/gui/src/app/store/env/base.js
hatched: This was more of a pondering question - after doing some micro benchmarks, turns out this is quite slow but because this isn't in a hot path I'm ok with letting it slide :)

hatched:
Thanks for this @frankban such a lot of work here! This really cleans things up.

:+1: :ok:

:shipit:

jujugui:

PR #2050, opened 2016-09-29
Display create new model button even when no models.

### LINKED ISSUE #2049: When user has no models, new model can't be created from profile page
If a user has no models, the model table is not shown and with it the "Create new" button is also missing. Thus, once you're on profile page, you can't go back to the canvas unless you click the back button in the browser.

easiest way to check it; bootstrap with --config identity-url=https://api.jujucharms.com/identity
then "juju grant everyone@external addmodel"
then do "juju gui" and login with sso
go to profile page

kadams54:
Fixes #2049.

### DIFF
--- a/jujugui/static/gui/src/app/components/user-profile/model-list/model-list.js
+++ b/jujugui/static/gui/src/app/components/user-profile/model-list/model-list.js
@@ -24,8 +24,8 @@ YUI.add('user-profile-model-list', function() {
     // broadcastStatus is necessary for communicating loading status back to
     // the parent SectionLoadWatcher.
     propTypes: {
+      acl: React.PropTypes.object,
       broadcastStatus: React.PropTypes.func,
-      canCreateNew: React.PropTypes.bool.isRequired,
       currentModel: React.PropTypes.string,
       listModelsWithInfo: React.PropTypes.func.isRequired,
       switchModel: React.PropTypes.func.isRequired,
@@ -209,17 +209,24 @@ YUI.add('user-profile-model-list', function() {
           </div>
         );
       }
-      const list = this.state.modelList;
-      if (!list || list.length === 0) {
-        return null;
-      }
-      const rows = list.map(this._generateRow);
       const props = this.props;
+      const acl = props.acl;
       let createNewButton;
-      if (props.canCreateNew) {
+      if (acl && acl.canAddModels()) {
         createNewButton = (
           <juju.components.CreateModelButton
             switchModel={this.switchModel} />
+        );
+      }
+      const list = this.state.modelList;
+      let content;
+      if (list && list.length > 0) {
+        const rows = list.map(this._generateRow);
+        content = (
+          <ul className="user-profile__list twelve-col">
+            {this._generateHeader()}
+            {rows}
+          </ul>
         );
       }
       return (
@@ -231,10 +238,7 @@ YUI.add('user-profile-model-list', function() {
             </span>
             {createNewButton}
           </div>
-          <ul className="user-profile__list twelve-col">
-            {this._generateHeader()}
-            {rows}
-          </ul>
+          {content}
         </div>
       );
     }
--- a/jujugui/static/gui/src/app/components/user-profile/model-list/model-list.js
+++ b/jujugui/static/gui/src/app/components/user-profile/model-list/model-list.js
@@ -209,15 +209,17 @@ YUI.add('user-profile-model-list', function() {
           </div>
         );
       }
-      const props = this.props;
-      const acl = props.acl;
       let createNewButton;
-      if (acl && acl.canAddModels()) {
-        createNewButton = (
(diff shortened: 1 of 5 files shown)


kadams54:
## QA

**Note:** This only works if you have charms or bundles or budgets or something else to populate your profile page. Otherwise you'll get the empty profile screen, which already has a "Create new model" button.

Per #2049:

1. Bootstrap with `--config identity-url=https://api.jujucharms.com/identity`.
2. `juju grant everyone@external addmodel`.
3. `juju gui` and login with SSO.
4. Go to the profile page. You should have no models listed and no create button.
5. In your local QA branch, `make dist`.
6. `juju upgrade-gui dist/jujugui-2.1.13.tar.bz2`.
7. Switch back to the browser, pop open the dev tools, then right click on refresh and select "Empty cache and hard reload".
8. You should now see a models section with a count of 0 and a create new button.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5896/
Test PASSed.

jujugui/static/gui/src/app/components/user-profile/model-list/model-list.js
```diff
@@ -222,6 +217,17 @@ YUI.add('user-profile-model-list', function() {
...
+      if (list && list.length > 0) {
+        const rows = list.map(this._generateRow);
+        content = (
+          <ul className="user-profile__list twelve-col">
+            {this._generateHeader()}
+            {rows}
+          </ul>
+        );
+      }
       return (
```
anthonydillon: This is always going to return the model list even if it is empty?

kadams54: @anthonydillon It will return the model list header with the button when the model list is empty. If everything else (bundles, charms, budgets, terms, etc.) are also all empty, the empty user profile page will be displayed instead.

anthonydillon: Ok, thanks :+1:

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5900/
Test PASSed.

jujugui/static/gui/src/app/components/user-profile/model-list/model-list.js
```diff
@@ -209,19 +209,26 @@ YUI.add('user-profile-model-list', function() {
...
-      const list = this.state.modelList;
-      if (!list || list.length === 0) {
-        return null;
-      }
-      const rows = list.map(this._generateRow);
       const props = this.props;
+      const acl = props.acl;
       let createNewButton;
-      if (props.canCreateNew) {
+      if (acl && acl.canAddModels()) {
```
hatched: I'm not entirely sure if canAddModels will return false if readonly is true so to be on the safe side could you expand this to be:

`if (acl && !acl.isReadOnly() && acl.canAddModels()) {`

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5901/
Test PASSed.

jujugui/static/gui/src/app/components/user-profile/model-list/model-list.js
frankban: The original impl with just acl.canAddModels() seems better to me actually.
It is surely possible that a user has read-only access to the current model but she's also able to create new ones.

frankban: The think that I am not sure about is the defensive if (acl && ...
This seems to suggest that acl could be undefined or null, which should never be the case in this code path, so that's just confusing. If it is for testing purposes, let's fix the tests.

kadams54: @frankban I wondered the same about whether `isReadOnly` is needed, after I looked that the `acl.js` code. `isReadOnly` looks at model-level access, whereas `canAddModels` looks at controller-level access. They don't really seem to be related as I'd initially thought.

As to the defensive check: we have two routes here:

1. Make `acl` a required prop, in which case the defensive check isn't needed.
2. Leave it as it is, an optional prop, and leave the defensive check in.

In general I feel like we have way, way too many required props, so my default preference is for optional props.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5902/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5904/
Test PASSed.

bac:
Followed the QA as written. Everything worked!

kadams54:
Woot. :shipit:

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5905/
Test PASSed.

jujugui:

PR #2115, opened 2016-10-10
When removing applications also remove the addCharm call

### LINKED ISSUE #2045: Removing subordinate causes type errors when deploying
Add nrpe to the canvas
Add Python-django to the canvas
Remove nrpe from the canvas
Click deploy

You'll now see a stream of errors and the deployment flow doesn't open.

### LINKED ISSUE #2088: When you delete a single service, [deploy changes (1)] still shows but is inactive
When you delete a single service, the [deploy changes (1)] button still shows but does not progress to the deploy screen - I can't click it, but it looks active. It is confusing to show 1 change on the button when the canvas is empty.

image

If all services on a canvas are destroyed, we should return to the empty canvas state, with the faded button and 0 changes. @urosj would you agree?

huwshimi:
When removing uncommitted applications we also need to remove the addCharm call if there are no other applications using that same addCharm call. This also contains a fix for removing relations when removing uncommitted applications added via a bundle import.

Fixes #2045
Fixes #2088

### DIFF
--- a/jujugui/static/gui/src/test/test_relation_utils.js
+++ b/jujugui/static/gui/src/test/test_relation_utils.js
@@ -427,6 +427,23 @@ describe('RelationUtils', function() {
       assert.strictEqual(result.near.serviceName, 'cs:mysql');
       assert.strictEqual(result.far.serviceName, 'mediawiki');
     });
+
+    it('does not fail if the far application has been removed', function() {
+      db.relations.add({
+        'interface': 'mysql',
+        scope: 'global',
+        endpoints: [
+          ['mysql', {role: 'provider', name: 'mydb'}],
+          ['mediawiki', {role: 'requirer', name: 'db'}]
+        ],
+        'id': 'mediawiki:db mysql:mydb'
+      });
+      db.services.getById = sinon.stub().returns(null);
+      const results = relationUtils.getRelationDataForService(db, service);
+      const result = results[0];
+      assert.strictEqual(result.near.serviceName, 'cs:mysql');
+      assert.isUndefined(result.far);
+    });
   });
 
   describe('DecoratedRelation and RelationCollection', function() {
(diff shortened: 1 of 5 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5974/
Test PASSed.

huwshimi:
QA:
- add an app
- the number of changes should appear in the deploy button
- destroy the app
- the deploy button should now read 0 changes
- add a bundle
- destroy all the apps
- the changes should all be removed except the machines, that issue will have to fixed in a following branch

jujugui/static/gui/src/test/test_relation_utils.js
```diff
@@ -427,6 +427,23 @@ describe('RelationUtils', function() {
       assert.strictEqual(result.near.serviceName, 'cs:mysql');
       assert.strictEqual(result.far.serviceName, 'mediawiki');
     });
+
+    it('does not fail if the far application has been removed', function() {
```
anthonydillon: It doesn't look like there is an application being removed in this test?

anthonydillon:
QA ok :+1: 

- Added MySQL which added 3 change
- Removed MySQL, changes reset to 0
- Added Kibana and Rsyslog resulted in 6 changes
- Removed and and changes went back to 0
- Added `Kubernetes Core` bundle which created 16 changes
- Deleted both applications which resulted in 5 changes (all machines added)

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/5978/
Test PASSed.

jujugui/static/gui/src/test/test_relation_utils.js
huwshimi: Ah yeah, good call, I've added a comment to make it clear what's going on.

fabricematrat:
:shipit:

jujugui:

PR #2151, opened 2016-10-20
Correctly reset the name when switching models

### LINKED ISSUE #2146: in GISF switching models does not show correct model name in Safari
When switch from one model to the other, it is common to see 'untitled-model' temporarily before the selected model name shows up. That's what happens in Chrome. In Safari the canvas populates but 'untitled-model' remains.

screen shot 2016-10-19 at 17 18 22

huwshimi:
Fixes #2146.

### DIFF
--- a/jujugui/static/gui/src/app/store/env/api.js
+++ b/jujugui/static/gui/src/app/store/env/api.js
@@ -228,7 +228,9 @@ YUI.add('juju-env-api', function(Y) {
       // would not be updated as when the model disconnected then the value was
       // not reset. The reason we need the value to be reset is so that when it
       // is set to the new value it fires a value change event, which in turn
-      // updates the name set in the db. See onEnvironmentNameChange.
+      // updates the name set in the db. See onEnvironmentNameChange. For some
+      // reason this can't be set to null as then it displays the name as null
+      // instead of picking up that the name has not been set.
       this.setConnectedAttr('environmentName', undefined);
     },
 
--- a/jujugui/static/gui/src/app/store/env/api.js
+++ b/jujugui/static/gui/src/app/store/env/api.js
@@ -222,6 +222,14 @@ YUI.add('juju-env-api', function(Y) {
       this.pendingLoginResponse = false;
       this.on('_rpc_response', this._handleRpcResponse);
       this._allWatcherBeingStopped = false;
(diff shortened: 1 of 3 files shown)


huwshimi:
QA:
- deploy with GISF
- load /new/
- open the model switcher and click on an existing model (you may have to create one if there are none and then start again)
- the model should load and in the model switch the name should temporarily display 'untitled-model' and then display the name of the model

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/6076/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/6079/
Test PASSed.

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -222,6 +222,14 @@ YUI.add('juju-env-api', function(Y) {
...
       this.on('_rpc_response', this._handleRpcResponse);
       this._allWatcherBeingStopped = false;
+      // When using GISF and switching between models or going from the
+      // unconnected state to a model the env properties are reset, but only
+      // those that were previously set. This meant that the environmentName
+      // would not be updated as when the model disconnected then the value was
+      // not reset. The reason we need the value to be reset is so that when it
+      // is set to the new value it fires a value change event, which in turn
+      // updates the name set in the db. See onEnvironmentNameChange.
+      this.setConnectedAttr('environmentName', undefined);
```
frankban: I am ok with this, so take it or leave it: why not doing this directly in app.js:show_environment? Something like `this.env.setConnectedAttr('environmentName', null);` in the already existing if block used in disconnected state. Also, I'd use null rather undefined, to communicate that the value has been actually set to something.

huwshimi: If the name is `null` then the model name gets displayed as such instead of displaying 'untitled-model'.

As for why it's done there, we only need to set the name when using the juju api, when using the controller api it works fine so we moved this set to the api module.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/6082/
Test PASSed.

hatched:
Closing for alternative PR which doesn't skip the test.

PR #2159, opened 2016-10-25
Add initial dispatch and url generation from new State

hatched:
This branch adds URL generation from the internal application state and the starting of the dispatch functionality.

### DIFF
--- a/jujugui/static/gui/src/app/state/state.js
+++ b/jujugui/static/gui/src/app/state/state.js
@@ -72,7 +72,7 @@ const State = class State {
       @private
       @type {Object}
     */
-    this._location = null;
+    this._location = cfg.location || null;
   }
 
   /**
--- a/jujugui/static/gui/src/app/state/state.js
+++ b/jujugui/static/gui/src/app/state/state.js
@@ -95,17 +95,23 @@ const State = class State {
   }
 
   /**
-    Check the supplied block of path to see if this is a user store path or not.
+    Check the supplied block of path to see if this is a user store path or
+    not. These options are becuase the store path can come in many forms and
+    we need to know if it's a user store path for both parsing and
+    generating the path.
     @return {Boolean}
   */
   _isUserStorePath(block) {
     const seriesList = this.seriesList;
+    // Store path ex) django/bundle/42
     if (block.length === 3 && block[1] === 'bundle') {
       return false;
     }
+    // Store path ex) hatch/ghost
     if (block.length === 2 && !seriesList.includes(block[1])) {
       return Number.isNaN(parseInt(block[1], 10));
     }
+    // Everything else
     return !Number.isNaN(parseInt(block[2], 10)) ||
       seriesList.includes(block[2]);
   }
@@ -226,38 +232,38 @@ const State = class State {
     @return {String} The path representing the current application state.
   */
   generatePath() {
-    let path = '';
+    let path = [];
     const root = this.appState.root;
     if (root) {
-      path = `${path}/${root}`;
+      path.push(root);
     }
     const search = this.appState.search;
     if (search) {
-      path = `${path}/${PATH_DELIMETERS.get('search')}/${search}`;
+      path = path.concat([PATH_DELIMETERS.get('search'), search]);
     }
     const user = this.appState.user || this.appState.profile;
     if (user) {
-      path = `${path}/${PATH_DELIMETERS.get('user')}/${user}`;
+      path = path.concat([PATH_DELIMETERS.get('user'), user]);
     }
     const store = this.appState.store;
     if (store) {
       if (this._isUserStorePath(store.split('/'))) {
-        path = `${path}/u`;
+        path.push(PATH_DELIMETERS.get('user'));
       }
-      path = `${path}/${store}`;
+      path.push(store);
     }
     const gui = this.appState.gui;
     if (gui) {
-      path = `${path}/${PATH_DELIMETERS.get('gui')}`;
+      path.push(PATH_DELIMETERS.get('gui'));
       Object.keys(gui).forEach(key => {
         const value = gui[key];
-        path = `${path}/${key}`;
+        path.push(key);
         if (value !== '') {
-          path = `${path}/${value}`;
+          path.push(value);
         }
       });
     }
-    return path;
+    return '/' + path.join('/');
   }
 
   /**
--- a/jujugui/static/gui/src/app/state/state.js
+++ b/jujugui/static/gui/src/app/state/state.js
@@ -66,6 +66,13 @@ const State = class State {
       @type {Array}
     */
     this.seriesList = cfg.seriesList;
+    /**
+      Internal storage value for the application location. Only used when
+      location is set externally.
+      @private
+      @type {Object}
+    */
+    this._location = null;
   }
 
   /**
@@ -79,6 +86,28 @@ const State = class State {
       .replace(this.baseURL, '')
       // Strip the leading and trailing slashes off the URL.
       .replace(/^\/*/, '').replace(/\/*$/, '');
+  }
+
+  /**
+    The current window location. If set has been called on this property then
+    it will return the externally set option. Typically this option will be used
+    for testing.
+    @type {Object}
+  */
+  get location() {
+    return this._location || window.location;
+  }
+
+  set location(location) {
+    this._location = location;
+  }
+
+  /**
+    Grabs the current location and parses it, building the state, then
+    executing the registered dispatchers.
+  */
+  parseURL() {
+    this.buildState(this.location.href);
   }
 
   /**
--- a/jujugui/static/gui/src/app/state/state.js
+++ b/jujugui/static/gui/src/app/state/state.js
@@ -95,28 +95,6 @@ const State = class State {
   }
 
   /**
-    Check the supplied block of path to see if this is a user store path or
-    not. These options are becuase the store path can come in many forms and
-    we need to know if it's a user store path for both parsing and
-    generating the path.
-    @return {Boolean}
-  */
-  _isUserStorePath(block) {
-    const seriesList = this.seriesList;
-    // Store path ex) django/bundle/42
-    if (block.length === 3 && block[1] === 'bundle') {
-      return false;
-    }
-    // Store path ex) hatch/ghost
-    if (block.length === 2 && !seriesList.includes(block[1])) {
-      return Number.isNaN(parseInt(block[1], 10));
-    }
-    // Everything else
-    return !Number.isNaN(parseInt(block[2], 10)) ||
-      seriesList.includes(block[2]);
-  }
-
-  /**
     The current window location. If set has been called on this property then
     it will return the externally set option. Typically this option will be used
     for testing.
@@ -139,7 +117,7 @@ const State = class State {
   }
 
   /**
-    Grabs the current location and parses it, building the state, then
+    Checks the current location and parses it, building the state, then
     executing the registered dispatchers.
   */
   dispatch() {
@@ -247,9 +225,6 @@ const State = class State {
     }
     const store = this.appState.store;
     if (store) {
-      if (this._isUserStorePath(store.split('/'))) {
-        path.push(PATH_DELIMETERS.get('user'));
-      }
       path.push(store);
     }
     const gui = this.appState.gui;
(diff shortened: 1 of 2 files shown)


hatched:
#### To QA

No qa, if tests pass then it's ok!

jujugui/static/gui/src/app/state/test-state.js
```diff
@@ -19,6 +19,247 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
 'use strict';
 
 describe('State', () => {
+
+  const USER_STATE_URLS = [{
```
hatched: These constants remain the same as were in the previous tests, they were just moved out of the individual tests so that they could be shared between the url parsing and generation suites.

jujugui/static/gui/src/app/state/test-state.js
```diff
@@ -292,379 +584,49 @@ describe('State', () => {
     });
   });
 
-  describe('State.buildState()', () => {
```
hatched: These were not removed, simply collapsed now that the data was abstracted out.

jujugui/static/gui/src/app/state/test-state.js
```diff
@@ -744,12 +706,88 @@ describe('State', () => {
...
+        location: {href: 'foo'}
+      });
+      sinon.stub(state,
+        'generateState', () => ({ error: null, state: {new: 'state'}}));
+      state.dispatch();
+      assert.deepEqual(state._appStateHistory, [{new: 'state'}]);
+    });
+  });
+
+  describe('State.generatePath()', () => {
```
hatched: This uses identical data to the url generation so we only have to add/modify the test data in one place and it'll be tested on both directions of the state system.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/6094/
Test PASSed.

jujugui/static/gui/src/app/state/state.js
```diff
@@ -82,6 +95,70 @@ const State = class State {
...
+    The current window location. If set has been called on this property then
+    it will return the externally set option. Typically this option will be used
+    for testing.
+    @type {Object}
+  */
+  get location() {
+    return this._location || window.location;
+  }
+
+  set location(location) {
```
hatched: I thought so too, but apparently you're only to document one of the set with jsdoc.

jujugui/static/gui/src/app/state/test-state.js
```diff
@@ -33,6 +274,35 @@ describe('State', () => {
     });
   });
 
+  it('can recieve a location value for testing', () => {
```
hatched: fixed

jujugui/static/gui/src/app/state/test-state.js
```diff
@@ -744,12 +706,88 @@ describe('State', () => {
...
+  });
+
+  describe('State.dispatch()', () => {
+    it('passes the current location to generateState', () => {
+      const state = new window.jujugui.State({
+        baseURL: 'http://abc.com:123',
+        seriesList:  ['precise', 'trusty', 'xenial'],
+        location: {href: 'foo'}
+      });
+      console.log(state.generateState);
```
hatched: removed

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/6095/
Test PASSed.

jujugui/static/gui/src/app/state/state.js
```diff
@@ -82,6 +95,76 @@ const State = class State {
...
+  */
+  _isUserStorePath(block) {
+    const seriesList = this.seriesList;
+    // Store path ex) django/bundle/42
+    if (block.length === 3 && block[1] === 'bundle') {
+      return false;
+    }
+    // Store path ex) hatch/ghost
+    if (block.length === 2 && !seriesList.includes(block[1])) {
+      return Number.isNaN(parseInt(block[1], 10));
```
frankban: What's the difference between Number.isNan and just isNan? What's better to use? Personal preference?

jujugui/static/gui/src/app/state/state.js
```diff
@@ -82,6 +95,76 @@ const State = class State {
   }
 
   /**
+    Check the supplied block of path to see if this is a user store path or
+    not. These options are becuase the store path can come in many forms and
+    we need to know if it's a user store path for both parsing and
+    generating the path.
+    @return {Boolean}
+  */
+  _isUserStorePath(block) {
```
frankban: I am not sure this function is useful. Why don't we keep the u/ prefix when parsing the URL? It seems that we are stripping it when saving the block in the state and then we need to add more logic just to check if state.store represents a user owned store entity. Why not considering u/ part of the charm or bundle URL (as it actually is)? At that point we would be able to just remove this function in theory?

jujugui/static/gui/src/app/state/state.js
```diff
@@ -82,6 +95,76 @@ const State = class State {
...
+    const seriesList = this.seriesList;
+    // Store path ex) django/bundle/42
+    if (block.length === 3 && block[1] === 'bundle') {
```
frankban: Why bundle is a special case? It should be a series like all the others and it is included in jujulib.URL.SERIES.

jujugui/static/gui/src/app/state/state.js
```diff
@@ -82,6 +95,76 @@ const State = class State {
...
+    if (error !== null) {
+      return `unable to generate state: ${error}`;
+    }
+    this._dispatch(state);
+  }
+
+  /**
+    Takes the existing app state and then calls the registered dispatchers.
+  */
+  _dispatch() {}
```
frankban: Throw a not implemented error?

jujugui/static/gui/src/app/state/state.js
```diff
@@ -82,6 +95,76 @@ const State = class State {
...
+
+  /**
+    Grabs the current location and parses it, building the state, then
+    executing the registered dispatchers.
+  */
+  dispatch() {
+    let error, state;
+    ({error, state} = this.generateState(this.location.href));
+    this._appStateHistory.push(state);
+    if (error !== null) {
```
frankban: Why returning a string from this function? Maybe throw an error? Or explicitly describe in the docstring that an error string or null is returned. Also, I'd check the error before pushing the state to the history. What do you think?

jujugui/static/gui/src/app/state/state.js
```diff
@@ -82,6 +95,76 @@ const State = class State {
...
+    Takes the existing app state and then calls the registered dispatchers.
+  */
+  _dispatch() {}
+
+  /**
+    Changes the internal state of the app, updating the location and
+    dispatching the app.
+    @param {Object} state - The new state delta to apply to the existing state.
+  */
+  changeState(state) {}
```
frankban: Throw a not implemented error?

jujugui/static/gui/src/app/state/state.js
```diff
@@ -82,6 +95,76 @@ const State = class State {
...
+  /**
+    The object representing the current app state.
+    @type {Object}
+  */
+  get appState() {
+    return this._appStateHistory[this._appStateHistory.length-1];
+  }
+
+  /**
+    Grabs the current location and parses it, building the state, then
```
frankban: This module mixes doc styles, we have both "Check the supplied block of path ..." and third person "Grabs the current location ...". I prefer the former, I suspect you prefer the latter, but it's only important that we are consistent :-)

jujugui/static/gui/src/app/state/test-state.js
frankban: Thanks for this comment, what do you think about renaming this to `const userStateTests` as they are actually "tests" in a table test scenario?

jujugui/static/gui/src/app/state/test-state.js
```diff
@@ -33,6 +274,35 @@ describe('State', () => {
...
+
+  it('uses the window.location if none is provided', () => {
+    // Note that this test is fragile, if any of the test runner execution
+    // parameters change then this test may fail because the location needs
+    // to be updated
+    const state = new window.jujugui.State({
+      baseURL: 'http://abc.com:123',
+      seriesList: ['trusty']
+    });
+    assert.equal(state.location.origin, 'http://0.0.0.0:6544');
```
frankban: Can we just use window.location here?

jujugui/static/gui/src/app/state/test-state.js
frankban: Nice!

jujugui/static/gui/src/app/state/test-state.js
```diff
@@ -292,379 +584,49 @@ describe('State', () => {
...
-
-      urls.forEach(test => {
-        assert.deepEqual(
-          state.buildState(test.path),
-          {error: test.error, state: test.state},
-          `${test.path} did not properly generate the state object: ` +
-          JSON.stringify(test.state));
+  describe('State.generateState()', () => {
+    // Generate all of the test cases for the various states.
+    [{
```
frankban:
```javascript
{
  title: 'blah blah',
  tests: rootStateTests
}
```
? [shrug]

jujugui/static/gui/src/app/state/state.js
hatched: This method is used for both parsing of the path and generating the path from state so in the latter case there is no 'u'

jujugui/static/gui/src/app/state/state.js
hatched: In ES2016 it's just a good practice to use it off of `Number`

jujugui/static/gui/src/app/state/state.js
hatched: Updated

jujugui/static/gui/src/app/state/state.js
hatched: Honestly, this function is still very much in development. I started writing it then switched to the URL generation.

Because the parser at least makes an attempt to generate the state when there is an error I was thinking that it would still try and push whatever state was outputted. Or would you say that if there is a state the user never comes off of the 'root' view of the GUI?

jujugui/static/gui/src/app/state/state.js
hatched: because nobody will use them until they are :)

jujugui/static/gui/src/app/state/state.js
hatched: because nobody will use them until they are :)

jujugui/static/gui/src/app/state/test-state.js
hatched: sure

jujugui/static/gui/src/app/state/test-state.js
hatched: Good call. it'll be less fragile then

jujugui/static/gui/src/app/state/test-state.js
hatched: sure

jujugui/static/gui/src/app/state/state.js
hatched: Because when it's coming from state this could be `hatch/mongodb/trusty` so the 1 spot is not a series.

jujugui/static/gui/src/app/state/state.js
frankban: I guess what I am suggesting is that state.store includes the u/ prefix in the case of user owned store entities: that would make this function unnecessary.

jujugui/static/gui/src/app/state/state.js
frankban: So seriesList.includes() should also match bundle right?

jujugui/static/gui/src/app/state/state.js
frankban: I'd just assume that, in case of errors, no usable state is ever returned. Otherwise it could be confusin, instead "it's either a good state or a good error" is more simple to explain. What do you think?

jujugui/static/gui/src/app/state/state.js
frankban: :-)

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/6100/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/6102/
Test PASSed.

hatched:

@frankban I'm going to hold off on some of the public API naming until it's ready to be integrated. At that point we can try a few colours on the bike shed :)

jujugui:

PR #2187, opened 2016-11-08
Applications with resources now get deployed with them

### LINKED ISSUE #2166: We should list resources when adding a charm
GUI doesn't deploy charms when they have resources. Deploying "kubernetes-core" for example fails with GUI, but works with CLI. The proper way to add a charm is to list resources with it. Check out code for conjure-up:
https://github.com/conjure-up/conjure-up/blob/master/conjureup/juju.py#L385-L403

hatched:
When deploying an application that has resources in the charm store we need to call a Juju api method `addPendingResources` prior to deploy to make Juju know about these resources.

Fixes #2166

### DIFF
--- a/jujugui/static/gui/src/app/store/env/api.js
+++ b/jujugui/static/gui/src/app/store/env/api.js
@@ -1076,6 +1076,22 @@ YUI.add('juju-env-api', function(Y) {
     },
 
     /**
+      Calls the environment's _addPendingResources method or creates a
+      new addPendingResources record in the ECS queue.
+
+      Parameters match the parameters for the _addPendingResources method
+      below. The only new parameter is the last one (ECS options).
+    */
+    addPendingResources: function(args, callback, options) {
+      if (options && options.immediate) {
+        // Call the deploy method right away bypassing the queue.
+        this._addPendingResources(args, callback);
+        return;
+      }
+      this.get('ecs')._lazyAddPendingResources(arguments);
+    },
+
+    /**
       Sends the provided resource info up to Juju without making it available
       yet.
 
@@ -1104,7 +1120,6 @@ YUI.add('juju-env-api', function(Y) {
           }, deployCallback);
         });
 
-      @method addPendingResources
       @param {Object} args The arguments required to make the call, including:
         - applicationName: the name of the application to be associated with
           the resources;
@@ -1120,7 +1135,7 @@ YUI.add('juju-env-api', function(Y) {
           or an empty object if the operation failed. The returned map can be
           used as is in the "resources" argument of the deploy call.
     */
-    addPendingResources: function(args, callback) {
+    _addPendingResources: function(args, callback) {
       // Validate the arguments.
       const resources = args.resources;
       if (!(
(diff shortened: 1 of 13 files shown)


hatched:
#### To QA
- Deploy the kubernetes-core bundle, it should deploy as expected.
- In a new model deploy a single non-resources charm, it should deploy as expected.
- In a new model deploy a single charm with resources (easyrsa works), it should deploy as expected.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/6164/
Test PASSed.

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -1200,6 +1200,20 @@ YUI.add('juju-env-api', function(Y) {
         return;
       }
       const ecs = this.get('ecs');
+      if (args.charmResources) {
```
frankban: I think callers have the responsibility to do this, and injecting an extraneous deploy argument like this can be confusing.

jujugui/static/gui/src/app/utils/environment-change-set.js
```diff
@@ -592,6 +592,18 @@ YUI.add('environment-change-set', function(Y) {
     },
 
     /**
+      Creates a new entry in the queue to add the pending resources.
+      @param {Array} The arguments to add the pending resources with.
+    */
+    _lazyAddPendingResources: function(args) {
```
frankban: Where are we collecting the response for this call?

hatched: We aren't, it didn't appear to be necessary.

jujugui/static/gui/src/app/store/env/api.js
hatched: Originally I was thinking the same but ended up this way. I can revert back.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/6169/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/6171/
Test PASSed.

jujugui/static/gui/src/app/utils/environment-change-set.js
```diff
@@ -638,6 +661,11 @@ YUI.add('environment-change-set', function(Y) {
               delete config[key];
             }
           });
+          // Check if we have any resources, if we do, then add them.
+          const resourceIds = ghostService.get('resourceIds');
```
frankban: Having one call executed after the other without the need for storing the pending ids in a state would have been better, but I understand that this may require changes in the ecs we don't want to do now. Are we sure this little state here will be cleaned up later? I guess the whole ghost service will be destroyed after firing the ecs.

hatched: I didn't have any plans on removing this. All other commands which change records have updated the corresponding record in the model. While this is adding a new record I don't think that it is to far from the conventions we've set forward. I actually think it would have been worse to create some kind of an internal ecs database to store this as we'd then have to make sure that never got out of sync.

hatched:
Thanks for the review @frankban were you also able to do a QA?

urosj:
QA 👍

hatched:
Thanks all! :shipit:

jujugui:

PR #2203, opened 2016-11-17
Update canvas state

huwshimi:


jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/6203/
Test FAILed.

jujugui/static/gui/src/app/app.js
```diff
@@ -1817,6 +1817,8 @@ YUI.add('juju-gui', function(Y) {
           this._clearDeployment.bind(this)]
       ]);
 
+      // Listen for change state events from the canvas components.
+      this.on('*:changeState', this._changeState, this);
```
hatched: Lets remove this and the `_changeState` method entirely and instead rely on passing the method around. This way we remove another piece of our dependence on YUI.

jujugui/static/gui/src/app/views/environment.js
```diff
@@ -197,12 +197,7 @@ YUI.add('juju-view-environment', function(Y) {
     */
     _handlePlusClick: function(e) {
       this.fire('changeState', {
```
hatched: Instead of firing the event here. in the `show_environment` method in app.js, pass in the changeState method and then call it directly here.

jujugui/static/gui/src/app/views/topology/relation.js
```diff
@@ -1159,9 +1159,8 @@ YUI.add('juju-topology-relation', function(Y) {
       var endpoint = Y.one(this).getAttribute('data-endpoint');
       var serviceId = endpoint.split(':')[0].trim();
       topo.fire('changeState', {
```
hatched: In `environment.js` pass the changeState method into the `setAttrs` call in `createTopology` and then call it directly.

jujugui/static/gui/src/app/views/topology/service.js
```diff
@@ -885,9 +885,8 @@ YUI.add('juju-topology-service', function(Y) {
     _deployLocalCharm: function(file, env, db) {
       var topo = this.get('component');
       topo.fire('changeState', {
```
hatched: Change all of these `fire` to call the new `changeState` method directly.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/6217/
Test FAILed.

huwshimi:
@hatched I've updated everything to pass in the state.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/6218/
Test FAILed.

hatched:
Thanks @huwshimi So it looks like there is a bug while qa'ing. If I have the inspector open and then mouse over the icon on the canvas, it closes the inspector but the state doesn't update. We can land this as the bug was there before, but maybe you could dig into that as you've been in this area.

PR #2242, opened 2016-11-23
Update state to handle search filters

huwshimi:


jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/6240/
Test FAILed.

jujugui/static/gui/src/app/state/state.js
```diff
@@ -449,7 +449,24 @@ const State = class State {
...
     if (search) {
-      path = path.concat([PATH_DELIMETERS.get('search'), search]);
+      path.push(PATH_DELIMETERS.get('search'));
+      // Append the text if it is truthy, i.e. not a blank string etc.
+      if (search.text) {
+        path.push(search.text);
+      }
+      // Everything that is not the 'text' param should be appened as a query
+      // string.
+      delete search.text;
```
hatched: This is deleting the key from the actual state object which we don't want to do because this will effectively be modifying history. Instead It's best to check in the loop below and continue if the `key === 'text'`

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/6242/
Test FAILed.

PR #2246, opened 2016-11-24
Validate the input field when leaving the field

anthonydillon:


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/6250/
Test PASSed.

anthonydillon:
## QA
- Hook up JIMM in `development.ini`
- Deploy an application
- Begin to enter an incorrect name for a model (such as, "-abc-") 
- See that the validation does not appear until you click off the input
- Check the input to something validate (such as "abc") Click off and see validation passes
- Do the same for the credentials name
- Leave an empty input and you should now only get the "This is required" validation instead of the character check too

jujugui/static/gui/src/app/components/deployment-flow/credential/add/add.js
```diff
@@ -277,7 +277,7 @@ YUI.add('deployment-credential-add', function() {
                   regex: /\S+/,
                   error: 'This field is required.'
                 }, {
-                  regex: /^[a-zA-Z0-9]([a-zA-Z0-9.-]*[a-zA-Z0-9])?$/,
+                  regex: /^([a-zA-Z0-9]([a-zA-Z0-9.-]*[a-zA-Z0-9])?)?$/,
```
hatched: IIRC this regex was grabbed from the Juju source - Can you confirm that this is still correct?

anthonydillon:
By adding the brackets it means the regex no passes on empty string. The
validation above it catches empty string.

urosj:
QA 🆗

urosj:
:shipit:

jujugui:

PR #2290, opened 2016-12-12
Export uncommitted changes.

huwshimi:


huwshimi:
QA:
- load the gui
- add some apps and machines and relations and config etc.
- export the model
- check the yaml file, all your changes should appear
- refresh the gui
- import your yaml file
- your model should load as it was previously

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/juju-gui/6340/
Test FAILed.

jujugui/static/gui/src/app/models/models.js
```diff
@@ -2343,22 +2343,29 @@ YUI.add('juju-models', function(Y) {
...
-            db.units.filterByMachine(machine.id).length === 0) {
-          return false;
-        }
-        return true;
-      });
+      if (includeUncommitted) {
+        machines = machineList.toArray();
+      } else {
+        // Strip uncommitted machines and containers from the machine list.
+        var db = this; // machineList.filter does not respect bindscope.
```
hatched: You could do this a couple ways. Because this is YUI the `filter` function takes a second argument which is the execution context. The other (and preferred) is to use an error function :)

jujugui/static/gui/src/app/models/models.js
```diff
@@ -2374,11 +2381,16 @@ YUI.add('juju-models', function(Y) {
       });
 
       machines.forEach(function(machine) {
-        // We're only intested in committed units on the machine.
-        var units = this.units.filterByMachine(machine.id).filter(
+        let units;
+        if (includeUncommitted) {
+          units = this.units.filterByMachine(machine.id);
+        } else {
+          // We're only intested in committed units on the machine.
```
hatched: "If we're only..."

jujugui/static/gui/src/app/views/utils.js
```diff
@@ -621,8 +621,18 @@ YUI.add('juju-view-utils', function(Y) {
...
   */
   utils.exportEnvironmentFile = function(db, legacyServicesKey) {
-    var result = db.exportDeployer(legacyServicesKey);
+    const apps = db.services.toArray();
+    const idMap = new Map();
+    // Store a map of all the temporary app ids to the real ids.
+    apps.forEach(app => {
+      idMap.set(app.get('id'), app.get('name'));
+    });
+    var result = db.exportDeployer(legacyServicesKey, true);
```
hatched: You have added the option for exporting uncommitted but then hard coded the value to `true` Do we maybe want this to be user definable when they click export and there are uncommitted items?

hatched:
CI Failure looks like a lint error.

jujugui/static/gui/src/app/views/utils.js
huwshimi: That's an interesting idea, I did this using the hardcoded value so that we can change it in the future without having to add the code back in. I guess we could do a followup to add the confirm if you think it'd be nice?

jujugui:
Build finished.

jujugui:
http://ci.jujugui.org:8080//job/juju-gui/6549/

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/juju-gui/6348/
Test PASSed.

jujugui/static/gui/src/app/models/models.js
```diff
@@ -2472,15 +2490,17 @@ YUI.add('juju-models', function(Y) {
...
-          return;
+        if (!includeUncommitted) {
+          // Exclude this service if it is a ghost or if it is named "juju-gui".
+          // This way we prevent the Juju GUI service to be exported when the
+          // bundle is created from a live environment. Note that this is a weak
+          // check: in theory, each deployed charm can be named "juju-gui", but
+          // we still assume this convention since there are no other (more
+          // solid) ways to exclude the Juju GUI service.
+          if (service.get('pending') === true ||
+          serviceName === JUJU_GUI_APPLICATION_NAME) {
```
frankban: I guess we should not include the Juju GUI application even in the case includeUncommitted is true.
So, something like
```javascript
if (
  (service.get('pending') && !includeUncommitted) || 
  (serviceName === JUJU_GUI_APPLICATION_NAME)
) {
  return;
}
```
Please also add a test for includeUncommitted === true and serviceName === JUJU_GUI_APPLICATION_NAME.

hatched: I was actually thinking we should remove this check. Now that the GUI is included by default in the model if the user has added the GUI it's an explicit part of their model.

hatched:
@frankban what advantage would enabling the functionality only if there is uncommitted stuff in the canvas get us? Wouldn't the functionality be identical?

jujugui/static/gui/src/app/models/models.js
huwshimi: Maybe it should check the Juju version and not include the juju-gui service when using Juju 1.

bac:
# QA Notes

## Chrome

I added a bundle and a charm as shown.

screen shot 2017-01-09 at 11 49 28

The YAML file created by the export only had:

```
$ cat sandbox-2017-01-09.yaml
series: trusty
applications: {}
relations: []
machines: {}
```

## Safari

Deploying the same bundle and charm as above, the export leads to a new tab containing
```
series: trusty
applications: 
  mediawiki: 
    charm: "cs:trusty/mediawiki-3"
    num_units: 1
    annotations: 
      "gui-x": "609"
      "gui-y": "-15"
    to: 
      - "0"
  mysql: 
    charm: "cs:trusty/mysql-29"
    num_units: 1
    options: 
      "query-cache-size": -1
    annotations: 
      "gui-x": "610"
      "gui-y": "255"
    to: 
      - "1"
  kibana: 
    charm: "cs:trusty/kibana-11"
    num_units: 1
relations: 
  - - "mediawiki:db"
    - "mysql:db"
machines: 
  "0": 
    series: trusty
  "1": 
    series: trusty
```

which looks correct. Manually saving it to a file (another bug in Safari) and re-importing it worked as expected.

huwshimi:
@bac strangely when I tried this with Chrome (Version 55.0.2883.87 (64-bit) on Ubuntu 16.10) I got:
```
series: trusty
applications: 
  kibana: 
    charm: "cs:trusty/kibana-11"
    num_units: 1
    to: 
      - "2"
  mediawiki: 
    charm: "cs:trusty/mediawiki-3"
    num_units: 1
    annotations: 
      "gui-x": "609"
      "gui-y": "-15"
    to: 
      - "0"
  mysql: 
    charm: "cs:trusty/mysql-29"
    num_units: 1
    options: 
      "query-cache-size": -1
    annotations: 
      "gui-x": "610"
      "gui-y": "255"
    to: 
      - "1"
relations: 
  - - "mediawiki:db"
    - "mysql:db"
machines: 
  "0": 
    series: trusty
  "1": 
    series: trusty
  "2": 
    series: trusty
```

jujugui/static/gui/src/app/models/models.js
huwshimi: Any further thoughts about this?

bac:
I did the QA again and this time the uncommitted model exported and imported just fine. It probably was just a caching issue the first time around. Thanks for the changes @huwshimi and sorry for the noise.

bac:
:shipit:

jujugui:

Issue #2297, opened 2016-12-14
Cannot commit k8s bundle in sandbox mode


hatched:
Add and deploy: https://demo.jujucharms.com/?store=cs:bundle/canonical-kubernetes-14
You'll see that most of the relations are never created

frankban:
Unfortunately this also affects the deployment of kubernetes on real models. Same issue, some relations are properly established, most of them are not, and many units are still pending even after committing the changes. No apparent JavaScript errors in the console.

frankban:
I was able to reproduce this issue also using the remove-yui-router branch on google cloud.

PR #2300, opened 2016-12-14
Simplify ECS code, and fix pending expose callbacks.

frankban:
This fixes https://github.com/juju/juju-gui/issues/2297 by addressing the underlaying issue: not providing callbacks to some API calls causes the ECS to not being able to complete.
In that scenario, for instance, expose callbacks are never called because the existing code failed to identify the missing argument and react accordingly.
This branch makes calling the ECS lazy methods explicit, so that we are more in control of what arguments are passed. As a consequence, some magic is also removed, which is always good.

To QA this, deploy the canonical k8s in sandbox mode and on real models. Also please exercise all the API calls and ECS interactons, as this branch introduces major changes in the logic there.

### DIFF
--- a/jujugui/static/gui/src/app/utils/environment-change-set.js
+++ b/jujugui/static/gui/src/app/utils/environment-change-set.js
@@ -81,27 +81,6 @@ YUI.add('environment-change-set', function(Y) {
     },
 
     /**
-      Returns an array of the parameters this method was called with skipping
-      the final parameter because it's the public method options.
-
-      @method _getArgs
-      @return {Array} An array of the arguments passed to this method.
-    */
-    _getArgs: function(args) {
-      var lastParam = args[args.length - 1];
-      var cut;
-      // If there is an object after the callback it's the
-      // configuration object for ECS.
-      if (typeof lastParam === 'object' && lastParam !== null &&
-          typeof args[args.length - 2] === 'function') {
-        cut = -1;
-      }
-      // Deep copy the resulting array of arguments, in order to prevent
-      // changeset functions to mutate the given data structures.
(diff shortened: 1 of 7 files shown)


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6557/

jujugui/static/gui/src/app/utils/environment-change-set.js
```diff
@@ -572,33 +538,35 @@ YUI.add('environment-change-set', function(Y) {
...
       // If there is an existing record for this charm then don't add another.
       if (existing) {
         return;
       }
-      var command = {
-        method: '_addCharm',
-        args: this._getArgs(args),
-        options: args[3]
-      };
+      const command = {method: '_addCharm', args: args, options: options};
```
hatched: Since you are doing edits for these commands, you can use object shorthand for the repeated properties (and the calls below)
```JavaScript
const command = {method: '_addCharm', args, options};
```

jujugui/static/gui/src/app/utils/environment-change-set.js
```diff
@@ -185,31 +164,19 @@ YUI.add('environment-change-set', function(Y) {
...
-      // Possible strict violation.
-      var self = this;
-      /**
-        Wrapper function for the supplied callback for the command.
-        @method _callbackWrapper
-      */
-      function _callbackWrapper() {
+      const args = record.command.args;
+      // By convention, the callback is the last argument.
+      const callback = args[args.length - 1];
```
hatched: This assumes that the user provided the callback - we should probably check that this is indeed a function.

hatched:
QA OK
:shipit:

jujugui:

PR #2319, opened 2017-01-06
Retrieve bundle changes from bundle service when in anonymous mode.

frankban:
This way it's still possible to add bundles to the canvas even when the GUI is in connected to the controller but not authenticated yet.

### DIFF
--- a/jujugui/static/gui/src/app/store/env/controller-api.js
+++ b/jujugui/static/gui/src/app/store/env/controller-api.js
@@ -350,7 +350,8 @@ YUI.add('juju-controller-api', function(Y) {
 
     /**
       Make a WebSocket request to retrieve the list of changes required to
-      deploy a bundle, given the bundle YAML content.
+      deploy a bundle, given the bundle YAML content. If the current connection
+      is not authenticated, fall back to using the restful bundle service API.
 
       @method getBundleChanges
       @param {String} bundleYAML The bundle YAML file contents.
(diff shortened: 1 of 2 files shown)


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6585/

frankban:
QA:
- switch to production jimm;
- run `make run` on this branch;
- run `guiproxy -uuid disconnected` so that gisf mode is simulated;
- visit the proxy address and ensure you are logged out (anonymous user);
- add a bundle to the canvas: it should load the uncommitted state as usual.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6586/

frankban:
Thanks for the reviews!
:shipit:

jujugui:

jujugui/static/gui/src/app/store/env/controller-api.js
```diff
@@ -387,6 +393,28 @@ YUI.add('juju-controller-api', function(Y) {
...
+        and a list of bundle changes.
+    */
+    _getBundleChangesFromBundleService: function(bundleYAML, callback) {
+      this.get('bundleService').getBundleChangesFromYAML(
+        bundleYAML, (error, changes) => {
+          if (error) {
+            callback([error], []);
+            return;
+          }
+          callback([], changes);
```
hatched: If there are no errors then the error position in the callback should be `null`. I know that this callback expects an array but that fix should be propagated up and down.

PR #2346, opened 2017-01-24
Model API: implement the KeyManager.ListKeys endpoint.

frankban:
This way we are able to retrieve existing keys and potentially use them when creating new models.

### DIFF
--- a/jujugui/static/gui/src/app/store/env/api.js
+++ b/jujugui/static/gui/src/app/store/env/api.js
@@ -2506,8 +2506,8 @@ YUI.add('juju-env-api', function(Y) {
 
       @method listKeys
       @param {String} user The user for whom the keys must be listed.
-      @param {Boolean} requestFullKeys Whether to pass full keys to the
-        callback, or just the SSH key fingerprints.
+      @param {Boolean} requestFullKeys Whether to request full keys or just the
+        SSH key fingerprints to be returned and passed to the callback.
       @param {Function} callback A callable that must be called once the
         operation is performed. It will receive two arguments: an error string
         and the list of keys as strings. If the operation succeeds, the error
--- a/jujugui/static/gui/src/app/store/env/api.js
+++ b/jujugui/static/gui/src/app/store/env/api.js
@@ -2502,6 +2502,59 @@ YUI.add('juju-env-api', function(Y) {
     },
 
     /**
+      Return the authorized SSH keys for the specified user.
+
+      @method listKeys
+      @param {String} user The user for whom the keys must be listed.
+      @param {Boolean} requestFullKeys Whether to pass full keys to the
+        callback, or just the SSH key fingerprints.
+      @param {Function} callback A callable that must be called once the
+        operation is performed. It will receive two arguments: an error string
+        and the list of keys as strings. If the operation succeeds, the error
+        is null.
+    */
+    listKeys: function(user, requestFullKeys, callback) {
+      if (!user) {
+        callback('no user provided', []);
+        return;
+      }
+      const handler = data => {
+        if (!callback) {
+          console.log('data returned by KeyManager.ListKeys API call:', data);
+          return;
+        }
+        if (data.error) {
+          callback(data.error, []);
+          return;
+        }
+        const results = data.response && data.response.results;
+        if (!results || results.length !== 1) {
+          // This should never happen.
+          callback('unexpected results: ' + JSON.stringify(results), []);
+          return;
+        }
+        const result = results[0];
+        const err = result.error && result.error.message;
+        if (err) {
+          callback('cannot list keys: ' + err, []);
+          return;
+        }
+        callback(null, result.result);
+      };
+
+      // Send the API call.
+      const entityTag = tags.build(tags.USER, user);
+      this._send_rpc({
+        type: 'KeyManager',
+        request: 'ListKeys',
+        params: {
+          entities: {entities: [{tag: entityTag}]},
+          mode: !!requestFullKeys
+        }
+      }, handler);
+    },
+
+    /**
       Make application endpoints available for consumption.
 
       @method offer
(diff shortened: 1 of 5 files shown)


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6609/

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -2502,6 +2502,59 @@ YUI.add('juju-env-api', function(Y) {
     },
 
     /**
+      Return the authorized SSH keys for the specified user.
+
+      @method listKeys
+      @param {String} user The user for whom the keys must be listed.
+      @param {Boolean} fullKeys Whether to pass full keys to the callback, or
```
mitechie: fullKeys feels like a list of keys or something, maybe something more boolean centric like justFingerprints or isFingerprint?

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -2502,6 +2502,59 @@ YUI.add('juju-env-api', function(Y) {
...
+        just the SSH key fingerprints.
+      @param {Function} callback A callable that must be called once the
+        operation is performed. It will receive two arguments: an error string
+        and the list of keys as strings. If the operation succeeds, the error
+        is null.
+    */
+    listKeys: function(user, fullKeys, callback) {
+      if (!user) {
+        callback('no user provided', []);
```
mitechie: hah, I see the Go in your JavaScript

jujugui/static/gui/src/app/store/env/sandbox.js
```diff
@@ -1517,6 +1518,23 @@ YUI.add('juju-env-sandbox', function(Y) {
...
+      @method handleKeyManagerListKeys
+      @param {Object} data The contents of the API arguments.
+      @param {Object} client The active ClientConnection.
+      @param {Object} state An instance of FakeBackend.
+    */
+    handleKeyManagerListKeys: function(data, client, state) {
+      client.receive({
+        'request-id': data['request-id'],
+        response: {results: [{
+          result: ['sandbox SSH key']
```
mitechie: should this handle the fullkey/fingerprint difference so that it can be replicated a bit in testing and such?

frankban:
QA:
- bootstrap a controller;
- upgrade-gui or use guiproxy with that controller;
- open the GUI and the JS console;
- run the following:
```javascript
var cb = (err, data) => {console.log(err, data)};
app.env.listKeys('admin', false, cb);
app.env.listKeys('admin', true, cb);
```
- do the same in sandbox mode.

jujugui/static/gui/src/app/store/env/sandbox.js
frankban: I think the GUI will always request the full keys, and, even if it makes sense to have the real API generic enough to support both cases, I'd not propagate unrequired complexity to stuff like sandbox,  at least until it's proven to be required.

jujugui/static/gui/src/app/store/env/api.js
frankban: The other way around would be scary ;-)

jujugui/static/gui/src/app/store/env/api.js
frankban: Done, I used requestFullKeys.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6610/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6611/

jujugui/static/gui/src/app/store/env/api.js
hatched: fwiw this pattern is also a convention in node ;)

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -2502,6 +2502,59 @@ YUI.add('juju-env-api', function(Y) {
     },
 
     /**
+      Return the authorized SSH keys for the specified user.
+
+      @method listKeys
+      @param {String} user The user for whom the keys must be listed.
+      @param {Boolean} requestFullKeys Whether to pass full keys to the
+        callback, or just the SSH key fingerprints.
```
hatched: While the effect of this argument is that the calback gets the fingerprint or the full key, the root cause is that it's requesting it differently from the api.

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -2502,6 +2502,59 @@ YUI.add('juju-env-api', function(Y) {
...
+        if (!callback) {
+          console.log('data returned by KeyManager.ListKeys API call:', data);
+          return;
+        }
+        if (data.error) {
+          callback(data.error, []);
+          return;
+        }
+        const results = data.response && data.response.results;
+        if (!results || results.length !== 1) {
```
hatched: Why wouldn't the length of the results not be 1? Could you add a comment to this effect plz

jujugui/static/gui/src/test/test_env_api.js
```diff
@@ -3056,6 +3057,116 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+          assert.deepEqual(keys, ['fingerprint']);
+          const msg = conn.last_message();
+          assert.deepEqual(msg, {
+            'request-id': 1,
+            type: 'KeyManager',
+            request: 'ListKeys',
+            version: 77,
+            params: {
+              entities: {entities: [{tag: 'user-dalek'}]},
+              mode: false
```
hatched: heh `mode: <Boolean>` how descriptive ;)

jujugui/static/gui/src/app/store/env/api.js
frankban: The comment is already there: that should never happen, but we are being defensive, as this is data provided by the network.

jujugui/static/gui/src/test/test_env_api.js
frankban: Yeah.

jujugui/static/gui/src/app/store/env/api.js
frankban: Sounds good.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6612/

hatched:
QA :ok: Thanks @frankban 


:shipit:

jujugui:

PR #2377, opened 2017-02-02
Fix opening deployment flow when there are no new apps

### LINKED ISSUE #2376: Get terms call fails when destroying machine
My model only had a single machine deployed into it (no applications/units etc)

I switched to the machine view, clicked to delete it, then clicked the 'commit' button. before the D opened it threw an error:

`TypeError: Cannot convert undefined or null to object`

https://github.com/juju/juju-gui/blob/remove-yui-router/jujugui/static/gui/src/app/components/deployment-flow/deployment-flow.js#L310 

```JavaScript
JSON.stringify(this.props.groupedChanges)
"{"_destroyMachines":{"destroyMachines-924":{"id":"destroyMachines-924","index":1,"parents":[],"executed":false,"command":{"method":"_destroyMachines","args":[["0"],true,null]},"timestamp":1485982670101}}}"
```
So you can see that there isn't a `_deploy` change which causes the `Object.keys()` to fail.
```JavaScript
      const deployCommands = this.props.groupedChanges['_deploy'];
      Object.keys(deployCommands).forEach(key => {
        appIds.push(deployCommands[key].command.args[0].charmURL);
      }) || [];
```

huwshimi:
Fixes #2376

### DIFF
--- a/jujugui/static/gui/src/app/components/deployment-flow/deployment-flow.js
+++ b/jujugui/static/gui/src/app/components/deployment-flow/deployment-flow.js
@@ -308,6 +308,9 @@ YUI.add('deployment-flow', function() {
       const appIds = [];
       // Find the undeployed app IDs.
       const deployCommands = this.props.groupedChanges['_deploy'];
+      if (!deployCommands) {
+        return;
+      }
       Object.keys(deployCommands).forEach(key => {
         appIds.push(deployCommands[key].command.args[0].charmURL);
       }) || [];
@@ -724,7 +727,8 @@ YUI.add('deployment-flow', function() {
         return false;
       }
       // Check that any terms have been agreed to.
-      if (this.state.terms.length > 0 && !this.state.termsAgreed) {
+      const terms = this.state.terms;
+      if (terms && terms.length > 0 && !this.state.termsAgreed) {
         return false;
       }
       // That's all we need if the model already exists.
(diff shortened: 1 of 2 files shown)


huwshimi:
QA:
- add a machine using the machine view
- click deploy
- the deployment flow should open

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6667/

jujugui/static/gui/src/app/components/deployment-flow/deployment-flow.js
```diff
@@ -308,6 +308,9 @@ YUI.add('deployment-flow', function() {
       const appIds = [];
       // Find the undeployed app IDs.
       const deployCommands = this.props.groupedChanges['_deploy'];
+      if (!deployCommands) {
```
hatched: Could you add a comment here as to why we only care about deploys

jujugui/static/gui/src/app/components/deployment-flow/deployment-flow.js
```diff
@@ -724,7 +727,8 @@ YUI.add('deployment-flow', function() {
         return false;
       }
       // Check that any terms have been agreed to.
-      if (this.state.terms.length > 0 && !this.state.termsAgreed) {
+      const terms = this.state.terms;
+      if (terms && terms.length > 0 && !this.state.termsAgreed) {
```
hatched: Good catch, thx

frankban:
:shipit:

jujugui:

PR #2409, opened 2017-02-08
Switch to the Cloud.Cloud API for finding the MAAS server

makyo:
QA will involve using the GUI with a MAAS server hosting the model. I was able to poke at it a little on guimaas, but would appreciate lots more eyes on it, as the cloud API is more convoluted than I initially thought.

### DIFF
--- a/development.ini
+++ b/development.ini
@@ -11,7 +11,7 @@ pyramid.default_locale_name = en
 
 jujugui.sandbox = true
 jujugui.raw = true
-jujugui.combine = true
+jujugui.combine = false
 jujugui.gzip = true
 jujugui.cachebuster = dev
 jujugui.controllerSocketTemplate = /api
--- a/development.ini
+++ b/development.ini
@@ -11,7 +11,7 @@ pyramid.default_locale_name = en
 
 jujugui.sandbox = true
 jujugui.raw = true
-jujugui.combine = false
+jujugui.combine = true
 jujugui.gzip = true
 jujugui.cachebuster = dev
 jujugui.controllerSocketTemplate = /api
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -586,13 +586,15 @@ YUI.add('juju-gui', function(Y) {
 
       // Once we know about MAAS server, update the header accordingly.
       let maasServer = this.env.get('maasServer');
-      if (!maasServer) {
+      if (!maasServer && this.controllerAPI) {
         maasServer = this.controllerAPI.get('maasServer');
       }
       if (maasServer) {
         this._displayMaasLink(maasServer);
       } else {
-        this.controllerAPI.once('maasServerChange', this._onMaasServer, this);
+        if (this.controllerAPI) {
+          this.controllerAPI.once('maasServerChange', this._onMaasServer, this);
+        }
         this.env.once('maasServerChange', this._onMaasServer, this);
       }
 
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -2591,18 +2591,6 @@ YUI.add('juju-gui', function(Y) {
           );
         }, this);
       }
-      // Retrieve maas credentials should they exist.
-      // NB: this assumes that a maas cloud cannot be added to a multi-cloud
-      // controller. If this changes in the future, this code will need to be
-      // called somewhere else. - Makyo 2017-02-16
-      this.controllerAPI.getDefaultCloudName((error, name) => {
-        this.controllerAPI.getClouds([name], (error, data) => {
-          if (data['.names']) {
-            this.controllerAPI.set('maasServer',
-                data[data['.names'][0]].endpoint);
-          }
-        });
-      });
       // Handle the change set token if provided in the query.
       // The change set token identifies a collections of changes required
       // to deploy a bundle. Those changes are assumed to be already
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -585,11 +585,15 @@ YUI.add('juju-gui', function(Y) {
       this.env.after('login', this.onLogin, this);
 
       // Once we know about MAAS server, update the header accordingly.
-      var maasServer = this.env.get('maasServer');
-      if (maasServer === undefined) {
-        this.env.once('maasServerChange', this._onMaasServer, this);
-      } else {
+      let maasServer = this.env.get('maasServer');
+      if (!maasServer) {
+        maasServer = this.controllerAPI.get('maasServer');
+      }
+      if (maasServer) {
         this._displayMaasLink(maasServer);
+      } else {
+        this.controllerAPI.once('maasServerChange', this._onMaasServer, this);
+        this.env.once('maasServerChange', this._onMaasServer, this);
       }
 
       // Feed environment changes directly into the database.
@@ -2585,6 +2589,18 @@ YUI.add('juju-gui', function(Y) {
           );
         }, this);
       }
+      // Retrieve maas credentials should they exist.
+      // NB: this assumes that a maas cloud cannot be added to a multi-cloud
+      // controller. If this changes in the future, this code will need to be
+      // called somewhere else. - Makyo 2017-02-16
+      this.controllerAPI.getDefaultCloudName((error, name) => {
+        this.controllerAPI.getClouds([name], (error, data) => {
+          if (data['.names']) {
+            this.controllerAPI.set('maasServer',
+                data[data['.names'][0]].endpoint);
+          }
+        });
+      });
       // Handle the change set token if provided in the query.
       // The change set token identifies a collections of changes required
       // to deploy a bundle. Those changes are assumed to be already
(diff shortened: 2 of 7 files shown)


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6719/

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -810,24 +810,50 @@ YUI.add('juju-env-api', function(Y) {
...
 
     /**
+      Fetch the cloud information from the cloud facade. This will include
+      MaaS connection information.
+
+      @method cloudInfo
+      @param {String} cloud The cloud name retrieve information about
+      @param {Function} callback The callback function for handling data
+    */
+    cloudInfo: function(cloud, callback) {
```
frankban: We started implementing the calls so that two args are always passed to callbacks: an error string (or null if no error occurred) and the actual data (a cloud object in this case). For an example, see the listKeys call in api.js.

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -810,24 +810,50 @@ YUI.add('juju-env-api', function(Y) {
         this.setConnectedAttr('maasServer', null);
         return;
       }
-      this.modelGet(data => {
-        if (data.err) {
-          console.warn('error calling ModelGet API: ' + data.err);
+      this.cloudInfo(data.cloud, data => {
+        if (!data) {
```
frankban: Let's make this impossible by providing (err, cloud) to the callback.
So, for instance
```javascript
this.cloudInfo(data.cloud, (err, cloud) => {
  if (err) {
    console.log('cannot retrieve maas server from cloud info: ' + err);
    this.setConnectedAttr('maasServer', null);
    return;
  }
  this.setConnectedAttr('maasServer', cloud.server);
}
```
We could make the provided cloud a js friendly object, so that we don't have to do the .value stuff.

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -810,24 +810,50 @@ YUI.add('juju-env-api', function(Y) {
...
+          console.log('cannot retrieve maas-server from Cloud call');
+          this.setConnectedAttr('maasServer', null);
           return;
         }
-        const maasServer = data.config['maas-server'];
+        if (data.error) {
+          console.warn('error calling Cloud API: ' + data.error);
```
frankban: Since this seems to come from response.results[0], the error field here is an object with a message attribute.
But, as suggested above, we can change this so that we always return an error string, as in listKeys.

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -810,24 +810,50 @@ YUI.add('juju-env-api', function(Y) {
...
+    cloudInfo: function(cloud, callback) {
+      // Send the API request.
+      this._send_rpc({
+        type: 'Cloud',
+        request: 'Cloud',
+        params: {
+          entities: [tags.build(tags.CLOUD, cloud)]
+        }
+      }, data => {
+        callback(data.response.results[0]);
```
frankban: Note that data can also have an error field for global API errors that we are not checking here.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6733/

mitechie:
QA: getting an error on the console that A.endpoint is undefined:

                void this.setConnectedAttr('maasServer', null)) : void this.setConnectedAttr('maasServer', A.endpoint)

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6790/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6792/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6793/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6794/

makyo:
Tangentially (and a bit amusing): ran into [this bug](https://bugs.chromium.org/p/chromium/issues/detail?id=552753#c15) in Chrome, which may be worth keeping an eye on - debugging arrow funcs is strange, at the moment.

development.ini
```diff
@@ -11,7 +11,7 @@ pyramid.default_locale_name = en
 
 jujugui.sandbox = true
 jujugui.raw = true
-jujugui.combine = true
+jujugui.combine = false
```
frankban: Leftover?

jujugui/static/gui/src/app/app.js
```diff
@@ -2585,6 +2591,18 @@ YUI.add('juju-gui', function(Y) {
           );
         }, this);
       }
+      // Retrieve maas credentials should they exist.
```
frankban: IIUC this onLogin in the app is called after the user logs into the model. What we really want I think is running this after the controller connection is authenticated. I'd suggest to move this code to controller-api.js in the handleLogin method, if the auth succeeds.

jujugui/static/gui/src/app/app.js
```diff
@@ -2585,6 +2591,18 @@ YUI.add('juju-gui', function(Y) {
...
+      // NB: this assumes that a maas cloud cannot be added to a multi-cloud
+      // controller. If this changes in the future, this code will need to be
+      // called somewhere else. - Makyo 2017-02-16
+      this.controllerAPI.getDefaultCloudName((error, name) => {
+        this.controllerAPI.getClouds([name], (error, data) => {
```
frankban: Before doing the second call to getClouds we need to check if there is an error returned by the first call (getDefaultCloudName), in which case I'd just stop the execution and log the error in the console.

jujugui/static/gui/src/app/store/env/controller-api.js
```diff
@@ -948,6 +948,11 @@ YUI.add('juju-controller-api', function(Y) {
             return prev;
           }
           prev[name] = this._parseCloudResult(result.cloud);
+          if (!prev['.names'] || !prev['.names'].length) {
```
frankban: I guess here you could use `if (!index)` (and document this attr in the docstring), but I am not sure this is really required.
Are we doing this because we want to keep the order? But isn't the order already known by the caller, which calls this method with a list of names?

jujugui/static/gui/src/test/test_controller_api.js
```diff
@@ -1607,7 +1607,9 @@ describe('Controller API', function() {
       // Perform the request.
       controllerAPI.getClouds(['lxd', 'google', 'no-such'], (err, clouds) => {
         assert.strictEqual(err, null);
+        console.log(clouds);
```
frankban: Leftover.

jujugui/static/gui/src/app/app.js
frankban: Also, data seems to generic, those are cloud objects, so maybe `clouds`?

jujugui/static/gui/src/app/app.js
```diff
@@ -2585,6 +2591,18 @@ YUI.add('juju-gui', function(Y) {
...
+          if (data['.names']) {
```
frankban: We should check the error here as well, for the second call, maybe
```javascript
this.controllerAPI.getDefaultCloudName((error, name) => {
  if (error) {
    console.error('cannot retrieve cloud name for maas: ' + error);
    return;
  }
  this.controllerAPI.getClouds([name], (error, clouds) => {
    const err = error || clouds[name].err;
    if (err) {
      console.error('cannot retrieve clouds for maas: ' + err);
      return;
    }
    this.controllerAPI.set('maasServer', clouds[name].endpoint);
  });
});
```
What do you think?

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -2929,6 +2929,12 @@ YUI.add('juju-env-api', function(Y) {
       }, handler);
     }
 
+  }, {
+    ATTRS: {
```
frankban: Thanks!

jujugui/static/gui/src/app/store/env/controller-api.js
makyo: May have been left over due to scoping issues we were running into in the debugger (I want to get `data[name]`, but `name` didn't appear available in the inner arrow func, so I worked around it). I'll research if that's still needed.

jujugui/static/gui/src/app/app.js
makyo: Fair enough.

development.ini
makyo: Thanks.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6802/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6803/

development.ini
frankban: You are welcome, but it's still there :-D

jujugui/static/gui/src/app/store/env/controller-api.js
```diff
@@ -152,6 +152,24 @@ YUI.add('juju-controller-api', function(Y) {
         }
         // Clean up for log out text.
         this.failedAuthentication = false;
+        // Retrieve maas credentials should they exist.
+        // NB: this assumes that a maas cloud cannot be added to a multi-cloud
+        // controller. If this changes in the future, this code will need to be
+        // called somewhere else. - Makyo 2017-02-16
+        this.getDefaultCloudName((error, name) => {
```
frankban: Much better thanks!

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6811/

mitechie:
QA ok, no errors in console that aren't expected it looks like. QA'd on 2.1rc2 controller and love the urls

https://10.0.0.155:17070/gui/u/rharding/mymodel

makyo:
TY! :shipit:

jujugui:

Issue #2414, opened 2017-02-09
Weird pop up when logging out


frankban:
When trying to log out as a USSO user, the GUI correctly logs me out, but a pop up is opened to the idm and it says that I am logged in.

PR #2423, opened 2017-02-09
When logging out, clean up the state.

frankban:
So that we don't have dispatched components starting macaroon acquisitions because they don't have a user, and therefore we prevent popups when logging out.
Also, defer the state change to the end of the log out process, to avoid weird connection stages when redispatching.
Fixes https://github.com/juju/juju-gui/issues/2414

Note that, while cleaning up the state is desirable, we'll also need to pay more attention at authentication assumptions in our components, especially before calling out to macaroon protected services.

### DIFF
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -2424,9 +2424,6 @@ YUI.add('juju-gui', function(Y) {
       if (controllerAPI) {
         closeController = controllerAPI.close.bind(controllerAPI);
       }
-      this.state.changeState({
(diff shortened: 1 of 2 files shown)


frankban:
QA: log in (gijoe), go to the profile page, log out. You should be redirected to the root without any additional popup opened.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6743/

hatched:
QA OK Thanks @frankban :shipit:

jujugui:

PR #2447, opened 2017-02-14
Add method to get agreements by term

hatched:
With the `terms.getAgreementsByTerms` method you can now pass it a list of terms and it'll return the terms which need to be agreed to still.

### DIFF
--- a/jujugui/static/gui/src/app/jujulib/terms.js
+++ b/jujugui/static/gui/src/app/jujulib/terms.js
@@ -87,10 +87,13 @@ var module = module;
     _formatTerm: function(term) {
       var milliseconds = Date.parse(term['created-on']);
       return {
-        user: term.user,
-        term: term.term,
+        content: term.content,
+        createdAt: new Date(milliseconds),
+        name: term.name,
+        owner: term.owner,
         revision: term.revision,
-        createdAt: new Date(milliseconds)
+        term: term.term,
+        user: term.user
       };
     },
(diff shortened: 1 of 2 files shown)


hatched:
#### To QA

- Connect to an environment where you have agreed to terms, get those terms by running `app.getAgreements((e, t) => console.log(t))`
- Then taking one of those agreements run `app.getAgreementsByTerms([<term-name>], (e, t) => console.log(t))` and it should return null, or the term if you haven't yet agreed to it. On jujugui.org you can use the `hatch/test-term1` term if you need one you haven't yet agreed to.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6769/

jujugui/static/gui/src/app/jujulib/terms.js
```diff
@@ -161,6 +164,42 @@ var module = module;
       };
       var url = this.url + '/agreements';
       return jujulib._makeRequest(this.bakery, url, 'GET', null, handler);
+    },
+
+    /**
+      Calls to check if the supplied terms have been agreed to. If they have
+      it returns an empty array, else it'll return the terms which need to be
```
huwshimi: I might be mistaken, but doesn't line 171 say this will return null?

huwshimi:
Thanks @hatched :shipit:

jujugui:

PR #2481, opened 2017-02-20
Fix the wrapping of long entity titles

### LINKED ISSUE #2453: Long charm/bundle names are not wrapped properly.
The design (showing how things should look with a title that wraps): Good wrapping

And reality:

Bad wrapping

huwshimi:
Fixes #2453.

### DIFF
--- a/jujugui/static/gui/src/app/components/entity-details/header/_header.scss
+++ b/jujugui/static/gui/src/app/components/entity-details/header/_header.scss
@@ -118,18 +118,17 @@
   }
 
   &__icon {
+    float: left;
     border-radius: 50%;
     height: 96px;
     width: 96px;
-    display: inline-block;
     vertical-align: top;
     margin: 0 30px 0 0;
   }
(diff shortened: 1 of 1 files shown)


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6819/

huwshimi:
QA:
- find a charm or bundle with a long name or view an entity and use your browser tools to modify the name
- check that the title wraps correctly onto a new line

Lukewh:
First QA instruction won't work properly as the height of the container is set via the JavaScripts, so changing the title after it's loaded won't refresh the height (which is also pretty bad if you increase font-size in browser). 

Therefore QA-OK, however your branch doesn't seem to include the revision number in the header?

frankban:
:shipit: 
@huwshimi , please consider the suggestion from @Lukewh for a follow-up branch.

jujugui:

Issue #2476, opened 2017-02-17
Can't visit /u/team or /u/user from the GUI with URL


urosj:
Try visiting a list of charms from another team when signed in to GUI. It redirects you back to your profile page. It should try to connect to the model and if that doesn't exist, try to show the content of store.

frankban:
Clicking on charms/bundles on the profile pages also does not work: it raises errors as it tries to use the old style charm store entity URLs. I'll fix that as part of this same branch.

Also note that I requested a quick UX workaround for the fact that, after visiting someone else's profile, there is no way to go back to my profile: we are missing some navigation capabilities.

kasiag:
@frankban @urosj @anthonydillon  Here's summary of a hangout between me, Elvi, Luke and Francesco as well as actions that came out of it:
since this bug is a release blocker we went with a solution of linking the juju logo with user's profile. Therefore regardless of where the user is, in their canvas or looking at someone else's models/ profile if they click on the juju logo, they will be taken back to their original profile. 

We've also proposed to re-visit this implementation at the JAAS sprint w/c 6th March for any changes.

PR #2491, opened 2017-02-21
Support for visiting someone else's profile page.

frankban:
This include groups, and it must work also when the URL is directly referenced.
Also fix navigating to user owned charms/bundles from the profile view.
The profile of other users now include models those users shared with you.
Refreshing on a model page is now more reliable, and less affected by possible races.

Also fix a JS error when logging out and trying to show the Juju logo, and a problem in which the maas server button is always displayed, even when not in a maas cloud.

Fixes https://github.com/juju/juju-gui/issues/2476

### DIFF
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -803,14 +803,18 @@ YUI.add('juju-gui', function(Y) {
           this.maskVisibility(false);
           const isLogin = current.root === 'login';
           const newState = {
-            root: isLogin ? null : current.root,
+            profile: current.profile,
+            root: isLogin ? null : current.root
           };
           // If the current root is 'login' after logging into the controller,
-          // or there is no root defined and no store defined then we want to
-          // render the users profile.
-          if (!current.store &&
-              (isLogin || !current.root) &&
-              this.get('gisf')) {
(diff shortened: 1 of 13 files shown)


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6831/

frankban:
**QA** must be done in both gijoe and gisf scenarios.
For gisf, at this stage you can use `guiproxy -config 'gisf: true, jujuEnvUUID: ""'`.

- try to visit your and someone else's profile page, both navigating the GUI or directly pointing your browser to, for instance, /u/hatch or /u/frankban. Check that everything works. Check, for instance, that when connected to the production jimm, you can switch to the juju-gui model owned by Uros, then click to his name in the breadcrumb and correctly see his profile with that single model shared with you.

- ensure you can correctly log into the charm store from your profile page in gijoe, not from other users' profiles. After you logged in you can see your charms and bundles, and you can click on them to correctly navigate to the entity details page.

- ensure that the maas UI button on the top right of the page is only visible when in maas.

Lukewh:
QA OK

frankban:
Thank you both for the reviews!
:shipit:

jujugui:

Issue #2511, opened 2017-02-22
No actions just keyboard shortcuts. 


elvirafigueras:

Lukewh:
Now that keyboard the keyboard shortcuts overlay is being surfaced, we should hide the settings on that overlay as they're not part of the same context.

hatched:
@elvirafigueras we can't just remove functionality that people use without moving it somewhere else. I'm ok with having the options not under a "Keyboard Shortcuts" heading but not removed entirely.

frankban:
This is now tracked at https://github.com/juju/juju-gui/issues/2526

PR #2513, opened 2017-02-22
Quick fix to hide general settings from the keyboard shortcuts overlay

Lukewh:
Fixes https://github.com/juju/juju-gui/issues/2511


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6837/

hatched:
Thanks for putting in this work @Lukewh, while I agree with it on principle, we can't just remove functionality (see my comment on the issue). So I'm going to block this until we have a proper plan forward here.

hatched:
Closing this so that another PR with these options moved can be created in its wake

PR #2517, opened 2017-02-22
Fixes automatic charmstore login from the canvas usso login link

jaycee:


jaycee:
QA:

In a private window with cache disabled spin up the gui with guiproxy, using `-config 'gisf:true, jujuEnvUUID:""'`. Go to /new/. Click the "Login" link in the upper right hand window, and allow popups to go through the full login process. You should then see your username in the breadcrumb in the left.

Click it to go to your profile. There should be no "Log in to charm store" button and you should see your charms/bundles if you have any.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6838/

hatched:
QA Notes:

After following the instructions sometimes I would close the popup window and I would already be on the profile, sometimes I'd be on the canvas. Both times a new popup would be trying to load to discharge to identity.

hatched:
After discussion with @jaycee - it is working when configured properly :)

However there does appear to be a race condition in state but that's separate from this branch.

:shipit:

jujugui:

PR #2527, opened 2017-02-27
To avoid unsetting values on empty config save, use configYAML field.

### LINKED ISSUE #2325: landscape bundle error when deployed via gui (KeyError in config changed hook in haproxy charm)
From https://bugs.launchpad.net/juju/+bug/1651260:

I ran into this bug while testing matrix, but it also seems to affect the gui.

Here's how to reproduce it:

1) Setup a model on aws, and fire up the gui.
2) Add the "Landscape Dense-Mass" bundle to the canvas (https://54.159.236.52:17070/gui/b13bbd14-2a27-4fbf-827e-35116de20180/?store=cs:bundle/landscape-dense-maas-1)
3) Commit, and wait for the deploy to settle.
4) Note that haproxy is in an errored state. Inspecting the logs, you will see a KeyError fetching the "services" key from the config during the config changed hook.

The error does not occur when one deploys the landscape bundle via the command line client. This bug possibly occurs because the haproxy charm is fairly old; updating the charmhelpers in that bundle might fix it. It would be nice to fix this in the gui (and in matrix!), though, as the juju client seems to be working around the bug, and other charms and bundles might break because of it.

Note that the "services" option is specified as a blank string in the bundle.yaml; it is possible that the juju client is doing something special with falsey values which the gui is not.

I'm filing this here because I suspect that we're running into some undocumented behavior on the part of core or the client, and I'd like to get some information about the fix here, before I attempt to apply the fix to matrix/python-libjuju, or file a bug with the gui.

(All this happens in a freshly bootstrapped controller, on aws, under juju 2.1 beta3.)

hatched:
Fixes #2325

### DIFF
--- a/jujugui/static/gui/src/app/store/env/api.js
+++ b/jujugui/static/gui/src/app/store/env/api.js
@@ -1302,13 +1302,14 @@ YUI.add('juju-env-api', function(Y) {
       const params = {
         application: args.applicationName,
         'charm-url': args.charmURL,
-        // In order to maintain backwards compatibility the Juju API implemented
-        // a new param, configYAML, so that empty configuration values to
-        // not 'unset' keys. This value is parsed by the Go yaml2 package
-        // So as long as we pass a key/value JSON map it will be able to
-        // parse it as YAML.
-        // In the event this needs to be reverted the key can simply be changed
-        // to 'config'.
+        // In order to maintain backwards compatibility, the Juju API
+        // implemented a new param, configYAML, so that an empty configuration
+        // value does not 'unset' keys. This value is parsed by the Go yaml2
+        // package So as long as we pass a key/value JSON map it will be able
+        // to parse it as YAML.
+        // In the event that making this modification causes issues setting
+        // configuration values due to the JSON>YAML conversion and needs to be
+        // reverted the key can simply be changed to 'config'.
         configYAML: stringifyObjectValues(args.config || {}),
         'config-yaml': args.configRaw || '',
         constraints: constraints,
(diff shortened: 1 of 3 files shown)


hatched:
#### To QA

Deploy charms, modifying the configs.
Switch to the CLI to confirm that the config values have been set appropriately. 
Be sure to use use PosgreSQL and landscape dense bundle

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6852/

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -1302,7 +1302,14 @@ YUI.add('juju-env-api', function(Y) {
       const params = {
         application: args.applicationName,
         'charm-url': args.charmURL,
-        config: stringifyObjectValues(args.config || {}),
+        // In order to maintain backwards compatibility the Juju API implemented
+        // a new param, configYAML, so that empty configuration values to
```
hatched: updated

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -1302,7 +1302,14 @@ YUI.add('juju-env-api', function(Y) {
...
+        // not 'unset' keys. This value is parsed by the Go yaml2 package
+        // So as long as we pass a key/value JSON map it will be able to
+        // parse it as YAML.
+        // In the event this needs to be reverted the key can simply be changed
```
hatched: updated

hatched:
Thanks! :shipit:

jujugui:

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6860/

PR #2533, opened 2017-02-28
Entity details pages have add to model name instead of canvas

### LINKED ISSUE #2466: "add to canvas" should be "add to model".
I think it'd be useful to s/canvas/model throughout the gui and site. Now that there's not really the idea of the demo GUI which is not a real model I think referring to a model is best. 

If there's a currently named or active model I think that the button can/should name that model. Add to model XXX or the like so that context is clear and meaningful to the end user.

hatched:
Fixes #2466

### DIFF
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -1767,10 +1767,7 @@ YUI.add('juju-gui', function(Y) {
         }
         charmstore.getEntity(url.legacyPath(), callback);
       };
-      const getModelName = () => {
-        const name = this.db.environment.get('name');
-        return name === 'untitled-model' ? undefined : name;
-      };
+      const getModelName = () => this.env.get('environmentName');
       ReactDOM.render(
         <window.juju.components.Charmbrowser
           acl={this.acl}
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -1767,6 +1767,10 @@ YUI.add('juju-gui', function(Y) {
         }
         charmstore.getEntity(url.legacyPath(), callback);
       };
+      const getModelName = () => {
+        const name = this.db.environment.get('name');
+        return name === 'untitled-model' ? undefined : name;
+      };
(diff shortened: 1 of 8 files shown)


hatched:
#### To QA
Visit a charm/bundle details page and it should now say "Add to model" instead of "Add to canvas".
Try this in unconnected mode and connected to a model. When connected to a model it should say "Add to {modelname}"

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6858/

jujugui/static/gui/src/app/app.js
```diff
@@ -1767,6 +1767,10 @@ YUI.add('juju-gui', function(Y) {
         }
         charmstore.getEntity(url.legacyPath(), callback);
       };
+      const getModelName = () => {
+        const name = this.db.environment.get('name');
+        return name === 'untitled-model' ? undefined : name;
```
huwshimi: I think you could just use `this.env.get('environmentName')` as it doesn't get set to "untitled-model".

jujugui/static/gui/src/app/app.js
```diff
@@ -1780,6 +1784,7 @@ YUI.add('juju-gui', function(Y) {
           getEntity={getEntity}
           getFile={charmstore.getFile.bind(charmstore)}
           getDiagramURL={charmstore.getDiagramURL.bind(charmstore)}
+          getModelName={getModelName}
```
huwshimi: So then this here could just be `modelName={this.env.get('environmentName')}`. This is how we do it in the deployment flow.

jujugui/static/gui/src/app/app.js
hatched: I thought the db was the source of truth, but you're correct that the `environmentName` is. Updating

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6861/

hatched:
Thanks @huwshimi :shipit:

jujugui:

PR #2577, opened 2017-03-07
Hide sharing icon until model API calls can be made.

### LINKED ISSUE #2549: Pressing the sharing button before a model full loads produces an error

kadams54:
Fixes #2549.

### DIFF
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -1495,9 +1495,6 @@ YUI.add('juju-gui', function(Y) {
       const db = this.db;
       const utils = views.utils;
       const env = this.env;
-      const modelConnected = () => {
-        return env.get('connected') && this.get('modelUUID');
-      };
       ReactDOM.render(
         <window.juju.components.ModelActions
           acl={this.acl}
@@ -1511,7 +1508,7 @@ YUI.add('juju-gui', function(Y) {
           hideDragOverNotification={this._hideDragOverNotification.bind(this)}
           importBundleFile={this.bundleImporter.importBundleFile.bind(
             this.bundleImporter)}
-          modelConnected={modelConnected}
+          modelAvailable={env.get('available')}
           renderDragOverNotification={
             this._renderDragOverNotification.bind(this)}
           sharingVisibility={this._sharingVisibility.bind(this)}/>,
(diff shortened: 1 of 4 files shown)


kadams54:
## QA

Follow the "to reproduce" instructions on #2549 and verify that you do not get the error. Additionally you should not see the button on /new, on the profile, or when we're in the middle of switching between models.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6880/

jujugui/static/gui/src/app/app.js
```diff
@@ -2803,6 +2806,7 @@ YUI.add('juju-gui', function(Y) {
         console.log('switching models is not supported in sandbox');
       }
       console.log('switching model connection');
+      this.env.set('available', false);
```
frankban: Why setting this as a YUI attribute? Do we plan to have subscribers for "available" changes? Also, why is this an attribute of the model connection? As the logic around model availability is only relevant at the app level, what do you think about making this an app._modelIsAvailable attribute?

kadams54: @frankban We already have `this.env.get('connected')`; however, that was inadequate as it only signals that the GUI is connected to the model. API calls will still fail because we haven't yet logged in, so we needed a second flag, alongsite `connected`, that would represent that the model is both connected and logged in. Hence `available`.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6882/

jujugui/static/gui/src/app/app.js
```diff
@@ -1168,6 +1168,12 @@ YUI.add('juju-gui', function(Y) {
         document.getElementById('top-page-container'));
       // The model name should not be visible when viewing the profile.
       this._renderBreadcrumb({ showEnvSwitcher: false });
+      // Model actions should not be visible when viewing the profile.
+      ReactDOM.unmountComponentAtNode(
```
hatched: We should have methods which handle these unmounts like we do for the other component renderings.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6883/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6884/

hatched:
QA OK Thanks @kadams54 
:shipit:

jujugui:

PR #2612, opened 2017-03-08
Match CLI functionality for creating relations

### LINKED ISSUE #1055: Relation logic doesn't allow multiple relations between two services sharing an endpoint
The GUI currently allows the user to create multiple relations between the same two services as long as they differ by at least one endpoint and that the services have not yet been committed to the environment. However once the services have been deployed you are no longer able to add multiple relations between two services where an interface on one service is already being used.

This is because the logic that was originally written back in 2012 did not allow multiple relations on the same endpoint. 

This only works by chance with the ghost services because the relations do not have an interface defined and as such the comparison function simply returns when checking if they match (`'http' !== undefined`). A complete fix here would be to make sure that ghost relations are provided with the appropriate relation interface data and then update the logic in the comparison function to allow many to one style relations.

### LINKED ISSUE #2485: GUI doesn't allow subordinate charm to have non-subordinate relationship
Deploy the limeds-bigdata bundle: https://demo.jujucharms.com/applications/?store=cs:~tengu-team/bundle/limeds-bigdata-1

You see that `limeds-mongodb` and `mongodb` have a relationship. This relationship can be added using the cli and works correctly. This relationship cannot be added using the Juju GUI.

You can verify this by removing and re-adding the relationship.

hatched:
Match the CLI functionality for creating relations. This fixes the long standing issue of explicitly blocking multiple relations to a `requires` endpoint and properly respecting the subordinate scope when creating relations with multi-series charms.

Fixes #2485 
Fixes #1055

### DIFF
--- a/jujugui/static/gui/src/app/models/endpoints.js
+++ b/jujugui/static/gui/src/app/models/endpoints.js
@@ -53,34 +53,29 @@ YUI.add('juju-endpoints', function(Y) {
   },
 
   /**
-   * Find available relation targets for a service.
-   *
-   * @method getEndpoints
-   * @param {Object} application A service object.
-   * @param {Object} controller The endpoints controller.
-   *
-   * @return {Object} A mapping with keys of valid relation service targets
-   *   and values consisting of a list of valid endpoints for each.
-   */
+    Find available relation targets for an application.
+    @param {Object} application An application object.
+    @param {Object} controller The endpoints controller.
+    @return {Object} A mapping with keys of valid relation application targets
+      and values consisting of a list of valid endpoints for each.
+  */
   models.getEndpoints = function(application, controller) {
-    const targets = {},
-        requires = [],
-        provides = [],
-        appId = application.get('id'),
-        db = controller.get('db'),
-        endpointsMap = controller.endpointsMap;
+    const targets = {};
+    const appId = application.get('id');
+    const db = controller.get('db');
+    const endpointsMap = controller.endpointsMap;
+    const appEndpoints = endpointsMap[appId];
     const appIsSubordinate = application.get('subordinate');
-    // Bail out if the map doesn't yet exist for this service.  The charm may
+    const appSeries = models._getSeries(db, application);
+    // Bail out if the map doesn't yet exist for this application. The charm may
     // not be loaded yet.
-    if (!endpointsMap[appId]) {
-      return targets;
+    if (!appEndpoints) {
+      return {};
     }
     /**
-     * Convert a service name and its relation endpoint info into a
-     * valid relation target endpoint, ie. including service name.
-     *
-     * @method getEndpoints.convert
-     */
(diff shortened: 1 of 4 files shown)


hatched:
#### To QA
Add `limeds-mongodb` and `mongodb` to the canvas and relate/deploy them.
It should allow the relation.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6898/

frankban:
QA ok.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6940/

hatched:
This has been totally rewritten since the last reviews, @huwshimi @frankban could you please take another look and 2 qa's :)

jujugui/static/gui/src/app/models/endpoints.js
```diff
@@ -53,127 +53,83 @@ YUI.add('juju-endpoints', function(Y) {
...
+      return {};
     }
     /**
-     * Convert a service name and its relation endpoint info into a
-     * valid relation target endpoint, ie. including service name.
-     *
-     * @method getEndpoints.convert
-     */
+      Convert an application name and its relation endpoint info into a
+      valid relation target endpoint, ie. including application name.
```
huwshimi: Needs `@params` etc?

jujugui/static/gui/src/app/models/endpoints.js
```diff
@@ -53,127 +53,83 @@ YUI.add('juju-endpoints', function(Y) {
...
-     * Store endpoints for a relation to target the given service.
-     *
-     * @method getEndpoints.add
-     * @param {Object} targetEndpoint Target endpoint.
-     * @param {Object} originEndpoint Origin endpoint.
-     */
+      Store endpoints for a relation to target the given application.
+      @param {Object} targetEndpoint Target endpoint.
+      @param {Object} originEndpoint Origin endpoint.
+    */
```
huwshimi: Missing `@applicationName` (I know this was an existing issue, but since you're here...)

jujugui/static/gui/src/app/utils/relation-utils.js
```diff
@@ -533,45 +531,50 @@ YUI.add('relation-utils', function(Y) {
...
+          const existingRelations = RelationUtils.getRelationDataForService(
+            db, subordinate);
+          // If there is an existing relation then we don't want to update the
+          // relation otherwise it will not match the existing application.
+          if (existingRelations.length === 0) {
+            // Update the subordinate series to match the application. We know
+            // that the subordinate is a multi-series charm with a series that
+            // matches the application otherwise we would not have got matching
+            // endpoints.
+            subordinate.set('series', appSeries);
```
huwshimi: If the subordinate is connected to multiple apps with different series, will this just update the series to the last connected app? Does that matter?

frankban:
QA ok thanks for these fixes!

Issue #2598, opened 2017-03-08
Put public adresses on top


galgalesh:
Request for small enhancement.

Users will be mostly interested in the public address of a unit, however, for the moment, the private address is on top. It might be better to put the public address of a unit on top since it is the information most relevant to the user.

![image](https://cloud.githubusercontent.com/assets/1492981/23701676/00bae700-03f8-11e7-8fac-ef467f5120d9.png)

Issue #2596, opened 2017-03-08
Open port ranges aren't shown correctly, makes GUI hang.


galgalesh:
The[ Eclipse Che charm](https://jujucharms.com/login/eclipse-che) exposes the port range `32768-65535/tcp`. This range is shown correctly in the cli, but the GUI shows each port individually. Clicking on the eclipse-che/0 unit hangs the GUI for ~10 seconds while the GUI generates >30.000 open port links.

The Charm needs that many exposed ports, otherwise the Charm won't be accessible in a public cloud.

Juju CLI:

```
Model      Controller  Cloud/Region   Version
dev-model  jujucharms  aws/us-east-1  2.0.3

App          Version  Status  Scale  Charm        Store       Rev  OS      Notes
eclipse-che           active      1  eclipse-che  jujucharms    0  ubuntu  

Unit            Workload  Agent  Machine  Public address  Ports                     Message
eclipse-che/0*  active    idle   0        54.84.173.104   8080/tcp,32768-65535/tcp  Ready (eclipse/che)

Machine  State    DNS            Inst id              Series  AZ
0        started  54.84.173.104  i-00c7c62a2b8e3499e  xenial  us-east-1b

```

Juju GUI: 

![image](https://cloud.githubusercontent.com/assets/1492981/23700077/0505ca06-03f2-11e7-90d0-d906fbce64cc.png)

frankban:
I believe we have the port range information on the mega-watcher, and there is surely area for improvement here. Thanks for reporting!

Issue #2570, opened 2017-03-03
inspector regression: change version does not work


frankban:
A message is displayed to users:
*There was a problem while loading the entity details. You could try searching for another charm or bundle or go back.*

In the console:
*cannot fetch the entity:Error: invalid charm/bundle name: "nagios-14"*
(also note the missing space after entity: in the console error).

Issue #2307, opened 2016-12-19
IP Address shows as None 


jrwren:
![screen shot 2016-12-19 at 9 32 34 am](https://cloud.githubusercontent.com/assets/106443/21316316/286bc3fe-c5ce-11e6-8c0e-e63dfc12b967.png)

$ juju status
MODEL       CONTROLLER  CLOUD/REGION   VERSION
parse-test  jem         aws/us-east-1  2.0.2.1

APP           VERSION  STATUS   SCALE  CHARM         STORE       REV  OS      NOTES
parse-server           unknown      1  parse-server  jujucharms    1  ubuntu

UNIT             WORKLOAD  AGENT  MACHINE  PUBLIC-ADDRESS  PORTS  MESSAGE
parse-server/0*  unknown   idle   0        54.82.203.49

MACHINE  STATE    DNS           INS-ID               SERIES  AZ
0        started  54.82.203.49  i-0e33d8b95d5e27d0f  xenial  us-east-1a

frankban:
This affects other browsers as well.

jrwren:
Confirmed in Chrome(57) and Safari(10)

jrwren:
Retrieving environment info.
Connecting to wss://jimm.jujucharms.com:443/model/12bcaa00-accd-4029-81ae-51180d36dd91/api.

--> #1 Admin:Login
{'params': {'auth-tag': 'user-admin', 'credentials': 'aaa'},
 'request': 'Login',
 'request-id': 1,
 'type': 'Admin',
 'version': 3}
<-- #1 response
{'error': 'redirection required',
 'error-code': 'redirection required',
 'request-id': 1,
 'response': {}}

--------------------------------------------------------------------------------
Traceback (most recent call last):
  File "megawatcher.py", line 137, in <module>
    start(connect())
  File "megawatcher.py", line 41, in connect
    conn.login(password)
  File "megawatcher.py", line 99, in login
    facades = resp['response']['facades']
KeyError: 'facades'

jrwren:
At some point, gui did display this in a red box in the lower left.

cannot update credentials: updating cloud credentials: read tcp 172.31.7.200:58318->172.31.36.230:37017: i/o timeout

Note the reserved ip addresses.

richardsith:
same problem if I try to deploy Postgresql, here is the screen

[...]

PR #2626, opened 2017-03-09
Use port ranges instead of ports when possible.

frankban:
Only single ports outside of ranges are turned into links.
Fixes https://github.com/juju/juju-gui/issues/2596

Also reorganize the view so that public ports are on top.
Fixes https://github.com/juju/juju-gui/issues/2598

Avoid showing "none" when no ports are still available but we've got addresses.
Fixes https://github.com/juju/juju-gui/issues/2307

Additionally, fix the inspector behavior when changing charm versions. Now it displays the new-style charm URLs and clicking on those actually works.
Fixes https://github.com/juju/juju-gui/issues/2570

In order to QA port ranges, you can use the eclipse-che charm, which anyway does not work in LXD. Also, since that charm has many ports open, the WebSocket message is too large and cannot be handled by the WebSocket library used by guiproxy (this will change when guiproxy starts using gorilla websockets). So I guess the best way to QA this is with a gijoe connected to a model in a public cloud.

### DIFF
--- a/jujugui/static/gui/src/app/store/env/controller-api.js
+++ b/jujugui/static/gui/src/app/store/env/controller-api.js
@@ -1308,7 +1308,6 @@ YUI.add('juju-controller-api', function(Y) {
       Modify (grant or revoke) user access to the specified model.
 
       @method _modifyModelAccess
-      @method _modifyModelAccess
       @param {String} modelId The UUID for the model on which the access is
                               being changed.
       @param {Array} users The usernames of the users who need modified access.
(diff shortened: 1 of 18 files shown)


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6918/

jujugui/static/gui/src/app/store/env/controller-api.js
```diff
@@ -1308,6 +1308,7 @@ YUI.add('juju-controller-api', function(Y) {
       Modify (grant or revoke) user access to the specified model.
 
       @method _modifyModelAccess
+      @method _modifyModelAccess
```
hatched: duplicate line?

frankban: Fixed, good catch ;-)

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6919/

hatched:
Thanks @frankban QA OK on all of the issues.

:shipit:

jujugui:

PR #2635, opened 2017-03-13
A bunch more YUI removal

huwshimi:
Contains:
- Replace Y.QueryString.parse with a util.
- Replace Y.JSON.stringify with JSON.stringify.
- Replace Y.JSON.parse with JSON.parse.
- Replaced Y.Lang.now with new Date().getTime().
- Replaced Y.Lang.now with new Date().getTime().
- Replace Y.Lang.isNumber with isNaN.

### DIFF
--- a/jujugui/static/gui/src/app/d3-components/d3-components.js
+++ b/jujugui/static/gui/src/app/d3-components/d3-components.js
@@ -27,7 +27,8 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 YUI.add('d3-components', function(Y) {
   var d3 = Y.namespace('d3'),
-      ns = Y.namespace('d3-components');
+      ns = Y.namespace('d3-components'),
+      utils = Y.namespace('juju.views.utils');
 
   var Module = Y.Base.create('Module', Y.Base, [], {
     _defaultEvents: {
@@ -232,7 +233,7 @@ YUI.add('d3-components', function(Y) {
         Object.keys(handlers).forEach(trigger => {
           const handler = self._normalizeHandler(
             handlers[trigger], module, selector);
-          if (L.isValue(handler)) {
+          if (utils.isValue(handler)) {
             _bindEvent(trigger, handler.callback,
                        container, selector, handler.context);
           }
@@ -544,4 +545,6 @@ YUI.add('d3-components', function(Y) {
     'base',
     'array-extras',
     'event',
-    'event-resize']});
+    'event-resize',
+    'juju-view-utils'
+]});
--- a/jujugui/static/gui/src/app/d3-components/d3-components.js
+++ b/jujugui/static/gui/src/app/d3-components/d3-components.js
@@ -27,8 +27,7 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 YUI.add('d3-components', function(Y) {
   var d3 = Y.namespace('d3'),
-      ns = Y.namespace('d3-components'),
-      L = Y.Lang;
+      ns = Y.namespace('d3-components');
 
   var Module = Y.Base.create('Module', Y.Base, [], {
     _defaultEvents: {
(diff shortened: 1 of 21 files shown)


huwshimi:
QA:
- have a play around with some of the areas that were modified, everything should work as it did

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6934/

jujugui/static/gui/src/app/models/charm.js
```diff
@@ -402,29 +403,6 @@ YUI.add('juju-charm-models', function(Y) {
       changelog: {
         value: {}
       },
-      //XXX jcsackett Aug 7 2013 This attribute is only needed until we turn
```
hatched: Hehe 2013 :D

jujugui/static/gui/src/app/models/charm.js
```diff
@@ -712,7 +690,7 @@ YUI.add('juju-charm-models', function(Y) {
          * @method scheme.setter
          */
         setter: function(val) {
-          if (!Y.Lang.isValue(val)) {
+          if (val === undefined && val === null && isNaN(val)) {
```
hatched: Why wasn't this one converted to the new `isValue` util?

jujugui/static/gui/src/app/d3-components/d3-components.js
```diff
@@ -162,25 +162,25 @@ YUI.add('d3-components', function(Y) {
...
       var result = {};
 
-      if (L.isString(handler)) {
+      if (typeof(handler) === 'string') {
         result.callback = module[handler];
         result.phase = 'on';
       }
 
-      if (L.isObject(handler)) {
+      if (typeof(handler) === 'object' && handler !== null) {
```
hatched: if handler is an array it'll also return `object`. While that's pretty unlikely, it might be worth a check here as well.

Lukewh: `!Array.isArray(handler)` I guess?

Can we create these as functions like YUI did? `utils.isString` `utils.isObject` etc. That also checks `!== null`?

jujugui/static/gui/src/app/store/env/base.js
```diff
@@ -441,7 +441,7 @@ YUI.add('juju-env-base', function(Y) {
      * @return {undefined} Fire an event only.
      */
     on_message: function(evt) {
-      this.fire('msg', Y.JSON.parse(evt.data));
+      this.fire('msg', JSON.parse(evt.data));
```
Lukewh: 👏

jujugui/static/gui/src/app/d3-components/d3-components.js
Lukewh: Or better starts with a `isValue`

jujugui/static/gui/src/app/views/utils.js
```diff
@@ -1524,6 +1524,52 @@ YUI.add('juju-view-utils', function(Y) {
...
+
+  /**
+    Parse a URL for the querystring and return it as an object.
+
+    @method parseQueryString
+    @param {String} URL The URL to get the query string from.
+    @returns {Object} The parsed query string..
+  */
+  utils.parseQueryString = URL => {
+    const querystring = URL.split('?')[1];
```
Lukewh: We should never get double `?` in the url, but we should probably defend against it?
`URL.split('?'); querystring.shift(); querystring.join('?');` Probably unnecessary, but urls are scary things.

bac:
# QA Notes

I tried doing the QA via GIJoe on AWS and ran into the problem filed as issue #2637

jujugui/static/gui/src/app/models/charm.js
huwshimi: Hah, yeah when I read that I just removed it (we weren't using the attribute).

jujugui/static/gui/src/app/models/charm.js
huwshimi: Oops!

jujugui/static/gui/src/app/d3-components/d3-components.js
huwshimi: @Lukewh Do you think we need the utils when the built in check works? Most of the checks e.g. typeof('string') === 'string' don't need to do anything more than check the type, and I wonder if the util gives us anything?

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6941/

hatched:
Thanks all :shipit:

jujugui:

Issue #2594, opened 2017-03-07
cannot click on add credential, console error


mitechie:
In the console I see:

Unable to add credential bad request: "cloudcred-aws_rick-harding+guibot@external_myaws" is not a valid cloudcred tag

I got no notification to the UI that there was an issue and that this failed at all.

bac:
The GUI has a bug that needs to be fixed in that it is eating the error.

The credential name should be valid, so the underlying error is in juju/names. It has been reported at https://bugs.launchpad.net/juju/+bug/1671915

PR #2642, opened 2017-03-15
Notify user of error when adding credentials.

kadams54:
Fixes https://github.com/juju/juju-gui/issues/2594.

It's a partial fix, as the real problem is the underlying Juju bug (see https://bugs.launchpad.net/juju/+bug/1671915) that does not recognize '+' as a legit part of an e-mail address. That said, now the user isn't left wondering why the app did nothing.

### DIFF
--- a/jujugui/static/gui/src/app/components/deployment-flow/credential/add/add.js
+++ b/jujugui/static/gui/src/app/components/deployment-flow/credential/add/add.js
@@ -23,6 +23,7 @@ YUI.add('deployment-credential-add', function() {
   juju.components.DeploymentCredentialAdd = React.createClass({
     propTypes: {
       acl: React.PropTypes.object.isRequired,
+      addNotification: React.PropTypes.func.isRequired,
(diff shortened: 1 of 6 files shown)


kadams54:
## QA

1. Fire up `guiproxy -controller jimm.jujucharms.com:443 -config 'gisf: true, jujuEnvUUID:""'`
2. `make clean-gui fast-babel gui run`
3. Open `http://localhost:8042/` in an incognito window.
4. Login as guibot. Message me if you need the credentials.
5. Add a charm to the model.
6. Commit the changes.
7. Type in a dummy model name (doesn't matter, we won't actually create one).
8. Select AWS for the cloud provider.
9. Select "Add credential..." in the credential dropdown.
10. Enter dummy data for the credential name and keys; again, it won't matter.
11. Click on the Add Credential button.
12. Verify you get back an error notification about the credential tag being invalid.

kadams54:
Sweet. Thanks @hatched! :shipit:

jujugui:

PR #2656, opened 2017-03-20
Improve header when loading models and on the profile.

### LINKED ISSUE #2350: Model action buttons are displayed in the header when in the profile.
Import & export are one-off, model-specific actions. When in the user profile, the user isn't in any particular model, at least not from a UX perspective. Consequently it doesn't make sense to continue displaying these buttons for the profile. It's unclear which model will be exported or what exactly will happen on import (Add to the last model viewed? Create a new model?).

selection_001

kadams54:
Fixes #2350.

### DIFF
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -650,7 +650,7 @@ YUI.add('juju-gui', function(Y) {
       // When the connection resets, reset the db, re-login (a delta will
       // arrive with successful authentication), and redispatch.
       this.env.after('connectedChange', evt => {
-        this._clearProviderLogo();
+        this._renderProviderLogo();
         if (!evt.newVal) {
           // The model is not connected, do nothing waiting for a reconnection.
           console.log('model disconnected');
@@ -1194,11 +1194,11 @@ YUI.add('juju-gui', function(Y) {
         />,
         document.getElementById('top-page-container'));
       // The model name should not be visible when viewing the profile.
-      this._renderBreadcrumb({ showEnvSwitcher: false });
+      //this._renderBreadcrumb({ showEnvSwitcher: false });
       // Model actions should not be visible when viewing the profile.
-      this._clearModelActions();
+      //this._clearModelActions();
       // Provider logo should not be visible when viewing the profile.
-      this._clearProviderLogo();
+      //this._clearProviderLogo();
     },
 
     /**
@@ -1583,6 +1583,7 @@ YUI.add('juju-gui', function(Y) {
       ReactDOM.render(
         <window.juju.components.ModelActions
           acl={this.acl}
+          appState={this.state}
           changeState={this.state.changeState.bind(this.state)}
           currentChangeSet={env.get('ecs').getCurrentChangeSet()}
           exportEnvironmentFile={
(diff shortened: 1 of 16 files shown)


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6956/

kadams54:
## QA

**Note:** this PR should be tested in *both* gisf and gijoe.

1. `make dist`
2. Either test as gijoe (`juju upgrade-gui dist/jujugui-2.4.3.tar.bz2 && juju gui --show-credentials`) or as gisf running against guiproxy (`guiproxy -controller jimm.jujucharms.com:443 -config 'gisf: true, jujuEnvUUID:""'`).
3. Verify that elements displayed in the header (breadcrumb, model switcher, import/export buttons, share button, cloud provider logo) are displayed when in a model.
4. Verify that only the breadcrumb is present when in a profile.
5. Verify that the elements that are show/hidden do so smoothly, particularly when switching between models with the model switcher, or when moving in and out of the profile.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6957/

frankban:
QA is ok, with one exception, that can be done in a follow up. The sharing button should not be visible when creating a new model. Animations are cool, thanks!

bac:
Hey @frankban try to remember to check and use labels, especially wrt to QA.  Thanks.

frankban:
I've seen that you were doing QA, didn't want to stop that, I judged this branch worth doubling that effort. Sorry, I'll try to make that clearer next time.

bac:
Oh, ok, thanks @frankban.  I just assumed you did a full QA and that was sufficient.  I'll have a look now too, just to give it extra coverage.

bac:
This looks good @kadams54 . When using GIJoe running AWS over a not terribly fast connection (thanks Time Warner) I saw some odd behavior going from the profile page back to a canvas.  The model name would appear by itself, then fade out, then fade back in with the import/export/sharing buttons. This was observed in Chrome on macOS.

Overall, QA-OK but thought you'd like to know about that little oddity.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6962/

jujugui/static/gui/src/app/app.js
```diff
@@ -2927,6 +2901,7 @@ YUI.add('juju-gui', function(Y) {
         console.log('switching models is not supported in sandbox');
       }
       console.log('switching model connection');
+      this.env.loading = true;
```
hatched: So when we call switch env we set this  property to `true` to mark it as loading. But the only way we can get it to be in a not loading state is if root is `new`? Or is there somewhere else that modifies this property that I don't see?

kadams54: The "somewhere else" is here: https://github.com/juju/juju-gui/pull/2656/files#diff-21dd467cb411bdd922556d1af7461864R2919

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6964/

kadams54:
The `/new` transition is still abrupt, but after some investigation, @hatched and I discovered that this was going to be too hard to fix. Maybe when the [Web Animation API](https://developer.mozilla.org/en-US/docs/Web/API/animation) is actually supported by the browsers.

:shipit:

jujugui:

PR #2674, opened 2017-03-23
User react.min.js instead of react-with-addons.min.js

hatched:
Using the proper build of React saves us 12KB


hatched:
#### To QA

Make sure the GUI still works :)

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/6994/

hatched:
After some investigation, the -with-addons is indeed required for the tests to run. They should not have passed with these changes...I'll close this for now and we can revisit again in the future.

PR #2678, opened 2017-03-27
More payment api methods

huwshimi:


huwshimi:
QA: see me, unless you know what you're doing.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7004/

jujugui/templates/index.html.mako
```diff
@@ -465,6 +465,6 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
     }
     </script>
     <!-- End Google Tag Manager -->
-
+    <script type="text/javascript" src="https://js.stripe.com/v2/"></script>
```
hatched: The instantiation of some of the code included above here requires this aiui so putting it down here means that it's possible that it won't be available when needed.

jujugui/static/gui/src/app/jujulib/register.js
```diff
@@ -171,6 +173,123 @@ var module = module;
...
+        exp_year: card.expYear,
+        name: card.name,
+        address_line1: card.addressLine1,
+        address_line2: card.addressLine2,
+        address_city: card.addressCity,
+        address_state: card.addressState,
+        address_zip: card.addressZip,
+        address_country: card.addressCountry
+      };
+      this.stripe.card.createToken(data, handler);
```
hatched: this assumes a third party library loaded successfully, which we can't be sure about as the GUI can be run in a network restricted environment. We'll need to know about this so that we can update the UI accordingly elsewhere so maybe you should add another method which is called at the start of every method requiring stripe.

```Javascript
stripeInitialized: function(callback) {
  if (!this.stripe) {
    callback('Stripe does not exist, cannot perform request.');
    return false;
  }
  return true;
}
``` 

And then this would change to

```JavaScript
if (this.stripeInitialized(callback)) {
  this.stripe.card.createToken(data, handler);
}
```

jujugui/static/gui/src/app/jujulib/register.js
```diff
@@ -23,12 +23,14 @@ var module = module;
...
       User instance.
+    @param stripe {Object} A reference to the Stripe API Object.
     @returns {Object} A client object for making payment API calls.
   */
-  function payment(url, bakery) {
+  function payment(url, bakery, stripe) {
     // Store the API URL (including version) handling missing trailing slash.
     this.url = url.replace(/\/?$/, '/') + paymentAPIVersion;
     this.bakery = bakery;
+    this.stripe = stripe;
```
hatched: In the event `stripe` is undefined

```JavaScript
if (!stripe) {
  console.error('Stripe not provided, some api requests will not complete successfully.');
}
this.stripe = stripe;
```

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7017/

huwshimi:
I've updated this to load the external js dynamically. If you have the payFlag on and have a look at your network tab and do `app.stripe._getStripe(() => {console.log(Stripe)})` you should see that the script gets loaded when that method is run. Running it again should not load the script a second time.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7018/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7019/

huwshimi:
Thanks @hatched, @fabricematrat :shipit:

jujugui:

PR #2699, opened 2017-03-30
Adds new user class and updates controller creds to use it

jaycee:


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7037/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7038/

jaycee:
QA:

Run the gui w/ guiproxy against the production jimm and `-config 'gisf:true, jujuEnvUUID:""'`; everything should auto login properly, and you should see any models of yours on the profile page. You should have access to the jujugui.org model, so click manage through it and ensure it comes up right.

If you have other models, check those and check switching between models.

Create a new model, do a deployment, and ensure all is well.

Bootstrap another model, run guiproxy against it without any special config. Make sure you can switch between models, make new models, deploy, etc.

jujugui/static/gui/src/app/app.js
```diff
@@ -929,6 +917,9 @@ YUI.add('juju-gui', function(Y) {
         return window.localStorage.getItem('discharge-token');
       };
       const charmstore = this.get('charmstore');
+      const setCreds = function(creds) {
```
hatched: I'm surprised this works because the `this` should be pointing to the component and not `app`. I'd like some investigation done here, but I feel that you'll probably want to change this to be.

```JavaScript
const setCreds = creds => {
  this.user.controller = creds;
}
```

jujugui/static/gui/src/app/app.js
```diff
@@ -1135,6 +1126,9 @@ YUI.add('juju-gui', function(Y) {
         // If the controller isn't ready yet then don't render anything.
         return;
       }
+      const getCreds = function() {
```
hatched: See above comment about context.

```JavaScript
const getCreds = () => this.user.controller.user;
```

jujugui/static/gui/src/app/app.js
```diff
@@ -2705,12 +2699,12 @@ YUI.add('juju-gui', function(Y) {
         // We don't always get a new username and password when switching
         // environments; only set new credentials if we've actually gotten them.
         // The GUI will automatically log in when we switch.
-        this.env.setCredentials({
+        this.user.controller = {
```
hatched: I know this follows the original pattern but won't this clear the macaroons if any were set?

jujugui/static/gui/src/app/app.js
```diff
@@ -2727,8 +2721,7 @@ YUI.add('juju-gui', function(Y) {
       const setUpModel = model => {
         // Tell the model to use the new socket URL when reconnecting.
         model.set('socket_url', socketUrl);
-        // Store the existing credentials so that they can be possibly reused.
-        model.setCredentials(credentials);
+        this.user.controller = credentials;
```
hatched: I'm not sure this is necessary any longer. I believe it's just going to set it to the same value it was at 2707

jujugui/static/gui/src/app/app.js
```diff
@@ -3117,6 +3110,10 @@ YUI.add('juju-gui', function(Y) {
       socketTemplate: {value: '/model/$uuid/api'},
 
       /**
+       The user auth object.
```
hatched: Can you expand on this

jujugui/static/gui/src/app/store/env/base.js
```diff
@@ -337,12 +282,11 @@ YUI.add('juju-env-base', function(Y) {
...
-      // Populate our credentials if they don't already exist.
-      const credentials = this.getCredentials();
+      const credentials = this.get('user').controller;
       if (!credentials.areAvailable) {
-        credentials.user = this.get('user') || '';
-        credentials.password = this.get('password') || '';
-        this.setCredentials(credentials);
+        credentials.user = '';
+        credentials.password = '';
+        this.get('user').controller = credentials;
```
hatched: I am not sure 285-289 is necessary as the `user.controller` will return empty strings for the `user` and `password` fields if none are provided anyways.

jujugui/static/gui/src/app/user/test-user.js
```diff
@@ -0,0 +1,29 @@
...
+ You should have received a copy of the GNU Affero General Public License along
+ with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+'use strict';
+
+chai.config.includeStack = true;
+chai.config.truncateThreshold = 0;
+
+describe('user auth class', () => {
```
hatched: I feel like you're missing some tests here :)

jujugui/static/gui/src/app/user/user.js
```diff
@@ -0,0 +1,88 @@
...
+    this.storage = cfg.storage || sessionStorage;
+  }
+
+  // TODO get username
+
+  // TODO get/set identity creds
+
+  // TODO get/set charmstore creds
+
+  get controller() {
```
hatched: There are a number of improvements I'd like to see here but as discussed this just moves the existing functionality.

jujugui/static/gui/src/test/test_app.js
```diff
@@ -97,30 +106,29 @@ describe('App', function() {
...
       });
       config.container = container;
       config.viewContainer = container;
+      console.error('MAKING THE APP');
       app = new Y.juju.App(Y.mix(config, {
         baseUrl: 'http://0.0.0.0:6543/',
         consoleEnabled: true,
         socketTemplate: '/model/$uuid/api',
-        controllerSocketTemplate: '/api'
+        controllerSocketTemplate: '/api',
```
hatched: trailing comma

jujugui/static/gui/src/app/app.js
jaycee: It does, but so would the original functionality. I'd like to put something smarter in that sorts out keys, but I think that's the next phase.

jujugui/static/gui/src/test/test_app.js
```diff
@@ -1,5 +1,5 @@
 /*
-This file is part of the Juju GUI, which lets users view and manage Juju
+ This file is part of the Juju GUI, which lets users view and manage Juju
```
frankban: Revert this?

jujugui/static/gui/src/test/test_controller_api.js
```diff
@@ -32,10 +32,22 @@ describe('Controller API', function() {
...
+      return {
+        store: {},
+        setItem: function(name, val) { this.store['name'] = val; },
+        getItem: function(name) { return this.store['name'] || null; }
+      };
+    };
+  };
+
   beforeEach(function() {
+    const userClass = new window.jujugui.User({storage: getMockStorage()});
```
frankban: s/userClass/user/ ? This is not a class, this is the actual instance AFAICT.

frankban:
This seems to work well with both macaroons and userpass. QA ok.

jujugui/static/gui/src/app/app.js
```diff
@@ -471,7 +470,7 @@ YUI.add('juju-gui', function(Y) {
         }
         this.state = this._setupState(baseURL);
       }.bind(this);
-
+      this.user = this.get('user') || new window.jujugui.User();
```
jaycee: Done.

jujugui/static/gui/src/app/app.js
```diff
@@ -527,24 +525,16 @@ YUI.add('juju-gui', function(Y) {
...
-          macaroons = credentials.macaroons;
-        }
+      var credentials = this.user.controller;
+      if (credentials.areAvailable) {
+        user = credentials.user;
+        password = credentials.password;
+        macaroons = credentials.macaroons;
       }
-      modelAPI.setCredentials({ user, password, macaroons });
+      this.user.controller = { user, password, macaroons };
```
jaycee: We can't quite; user, pass &c are used separately below. We can remove this reassign though.

jaycee: In the next pass we can also standarize on null or '' representing no data for user/pass fields and then make sure the controller always returns whichever and everything else accepts that. I lean towards null myself.

jujugui/static/gui/src/app/app.js
```diff
@@ -637,7 +627,7 @@ YUI.add('juju-gui', function(Y) {
         console.log('model connected');
         this.env.userIsAuthenticated = false;
         // Attempt to log in if we already have credentials available.
-        var credentials = this.env.getCredentials();
+        var credentials = this.user.controller;
```
jaycee: Done.

jujugui/static/gui/src/app/app.js
jaycee: This works b/c the Login component no longer has the property at all. :smile:

I've deleted the whole bit.

jujugui/static/gui/src/app/app.js
jaycee: This actually would fail b/c it's a func, and the component wants a string--so we're just passing in the username now.

jujugui/static/gui/src/app/app.js
jaycee: I've removed it.

jujugui/static/gui/src/app/app.js
jaycee: Done.

jujugui/static/gui/src/app/store/env/base.js
jaycee: So I don't know why, but removing that bit breaks model switching. I'm leaving it in for this branch and we can see about making it cleaner/removing it in another pass.

jujugui/static/gui/src/app/user/test-user.js
jaycee: Why yes. Yes I am. I'll put those in.

jujugui/static/gui/src/app/app.js
```diff
@@ -527,24 +525,16 @@ YUI.add('juju-gui', function(Y) {
...
-      if (!user || !password) {
-        // No user and password credentials provided in config: proceed with
-        // usual credentials handling.
-        var credentials = modelAPI.getCredentials();
-        if (credentials.areAvailable) {
-          user = credentials.user;
-          password = credentials.password;
-          macaroons = credentials.macaroons;
-        }
+      var credentials = this.user.controller;
```
jaycee: Fixed.

jujugui/static/gui/src/app/user/user.js
```diff
@@ -0,0 +1,88 @@
...
+ with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+'use strict';
+
+if (typeof this.jujugui === 'undefined') {
+  this.jujugui = {};
+}
+
+/** Class representing a user's authorizations in the GUI **/
```
jaycee: Fixed with better docstring per @hatched's request.

jujugui/static/gui/src/app/user/user.js
```diff
@@ -0,0 +1,88 @@
...
+const User = class User {
+
+  constructor(cfg = {}) {
```
jaycee: I think it's quite likely that as we add the expiration stuff and the other credentials to this object cfg is going to grow. If it doesn't it's easy to change later.

jujugui/static/gui/src/app/user/user.js
jaycee: Docstring added. Not sure about rename--we're going to be doing a second pass to clean this up, now that all the functionality is moved. When we do that I'll change the name, but I'm not sure about juju. I've added an XXX to that effect as well.

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -605,7 +605,7 @@ YUI.add('juju-env-api', function(Y) {
       if (this.pendingLoginResponse) {
         return;
       }
-      var credentials = this.getCredentials();
+      var credentials = this.get('user').controller;
```
jaycee: Done.

jujugui/static/gui/src/test/test_app.js
jaycee: Fixed.

jujugui/static/gui/src/test/test_controller_api.js
jaycee: Done.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7055/

jaycee:
:shipit:

jujugui:

Issue #2604, opened 2017-03-08
credential name shown to me in the DF makes no sense to me


mitechie:
I created a model from the cli and it grabbed my credential. Now in the UX I see the name in the DF is "aws_rharding@external_personal"

I don't know what credential that is and it doesn't mean anything to me. It's very confusing verifying what's being used on my new model.

frankban:
Yes, that's how the credential is called in Juju, and we should provide a more human friendly name (in the example case I guess `personal` would be good enough).

mitechie:
Yes, when I added the model from the CLI the output was that it was "using your cached credential 'personal'...

PR #2708, opened 2017-04-04
Display friendly credential names

huwshimi:
Fixes: #2604.

### DIFF
--- a/jujugui/static/gui/src/app/store/env/controller-api.js
+++ b/jujugui/static/gui/src/app/store/env/controller-api.js
@@ -1107,6 +1107,7 @@ YUI.add('juju-controller-api', function(Y) {
         - err: a possible result specific error, in which case all subsequent
           fields are omitted;
         - names: the list of names that identify cloud credentials
+        - displayNames: the list of credential names extracted from the full id
           corresponding to the user/cloud pair provided as input.
         If no errors occur, error parameters are null. Otherwise, in case of
         errors, the second argument is an empty array.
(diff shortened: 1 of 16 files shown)


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7067/

jujugui/static/gui/src/app/store/env/controller-api.js
```diff
@@ -1121,7 +1137,12 @@ YUI.add('juju-controller-api', function(Y) {
           const names = credentialTags.map(credentialTag => {
             return tags.parse(tags.CREDENTIAL, credentialTag);
           });
-          return {names: names};
+          const displayNames = names.map(
```
frankban: Please update the docstring of this method to reflect the presence of a new field.

jujugui/static/gui/src/app/store/env/controller-api.js
```diff
@@ -1192,6 +1213,7 @@ YUI.add('juju-controller-api', function(Y) {
           prev[name] = {
             authType: entry['auth-type'] || '',
             attrs: entry.attrs || {},
+            displayName: this._parseCredentialName(name),
```
frankban: Ditto: update docstring by including displayName.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7068/

kadams54:
QA is OK, but we now have another place we display credentials: in the expanded row of the user profile's model list. It would be nice to have a follow-up card that brings the displayName to the user profile as well. I'm not sure the best way to do that: call `_parseCredentialName` when storing the name with the model in `listModelInfo`? That has the disadvantage of the model not having the full credential name/id. Pass `_parseCredentialName` up to the `UserProfileEntity` component? Then the ID can be parsed into a user-friendly version at render time, but the downside is having to pass another prop into a deeply-buried component.

huwshimi:
@kadams54 good spot, I've added an issue to keep track of it here: #2711.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7076/

huwshimi:
Thanks @kadams54 @frankban :shipit:

jujugui:

PR #2716, opened 2017-04-05
Add a `next` state property to the state system.

hatched:
In order to redirect a user after logging in this branch adds a `next` property to the state system and handles redirecting to it on controller connect on load.

### DIFF
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -707,10 +707,16 @@ YUI.add('juju-gui', function(Y) {
           console.log('redirecting to "next" state', next);
           const {error, state: newState} = state.generateState(next, false);
           if (error === null) {
-            newState.root = null;
+            // The root at this point will be 'login' and because the `next`
+            // url may not explicitly define a new root path we have to set it
+            // to null to clear 'login' from the url.
+            if (!newState.root) {
+              newState.root = null;
+            }
             newState.special = null;
             this.maskVisibility(false);
-            return state.changeState(newState);
+            state.changeState(newState);
+            return;
           }
           console.error('error redirecting to previous state', error);
           return;
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -397,18 +397,6 @@ YUI.add('juju-gui', function(Y) {
 
       this.renderEnvironment = true;
 
-      // If this property has a value other than '/' then
-      // navigate to it after logging in.
-      this.redirectPath = '/';
-
-      // This attribute is used by the namespaced URL tracker.
-      // _routeSeen is part of a mechanism to prevent non-namespaced routes
-      // from being processed multiple times when multiple namespaces are
-      // present in the URL.  The data structure is reset for each URL (in
-      // _dispatch).  It holds a mapping between route callback uids and a
-      // flag to indicate that the callback has been used.
-      this._routeSeen = {};
-
       // When a user drags a file over the browser we show notifications which
       // are drop targets to illustrate what they can do with their selected
       // file. This array keeps track of those masks and their respective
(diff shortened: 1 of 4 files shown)


hatched:
#### To QA

jujugui/static/gui/src/app/app.js
```diff
@@ -701,20 +689,38 @@ YUI.add('juju-gui', function(Y) {
...
+        // If state has a `next` property then that overrides all defaults.
+        const specialState = current.special;
+        const next = specialState && specialState.next;
+        // There should never be a `next` state if we aren't on login but
+        // adding the state login check here just to be sure.
+        if (state.current.root === 'login' && next) {
+          console.log('redirecting to "next" state', next);
+          const {error, state: newState} = state.generateState(next, false);
+          if (error === null) {
+            newState.root = null;
```
hatched: Right, so this is because at this point `root` is going to be "login". If it's not set to `null` then the url will still have the `/login` until the model connects. This was handled by fluke elsewhere in the app :) I'll add a comment.

jujugui/static/gui/src/app/app.js
```diff
@@ -701,20 +689,38 @@ YUI.add('juju-gui', function(Y) {
...
+            newState.special = null;
+            this.maskVisibility(false);
+            return state.changeState(newState);
```
hatched: Yeah, sorry about that, I've been doing this style lately and I don't know why :D

jujugui/static/gui/src/app/state/state.js
```diff
@@ -661,17 +661,24 @@ const State = class State {
...
-      if (!state.special) {
-        state.special = {};
-      }
-      state.special.deployTarget = query['deploy-target'];
+    if (!query) {
+      return state;
+    }
+    let deployTarget = query['deploy-target'];
+    if (deployTarget) {
+      if (!state.special) { state.special = {}; }
```
hatched: done

jujugui/static/gui/src/app/state/state.js
```diff
@@ -661,17 +661,24 @@ const State = class State {
...
+      state.special.deployTarget = deployTarget;
       // Push the state so that any special query string is removed now,
       // therefore avoiding multiple dispatches of the same special callback.
       if (allowStateModifications) {
         this._pushState();
       }
     }
+    let next = query.next;
+    if (next) {
+      if (!state.special) { state.special = {}; }
```
hatched: done

jujugui/static/gui/src/app/app.js
hatched: Actually, this is could be a bug - if the `next` state supplies a new root then this will ignore it. I'll also fix that.

hatched:
Thanks for the review and qa @frankban. I totally agree with the follow-up, I was going to tackle that here, however in an effort to make sure this can get in for Garek I didn't want to wake any sleeping dragons :)

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7082/

hatched:
Follow-up tracking issue https://github.com/juju/juju-gui/issues/2717

jujugui/static/gui/src/app/app.js
```diff
@@ -701,20 +689,44 @@ YUI.add('juju-gui', function(Y) {
...
+            // to null to clear 'login' from the url.
+            if (!newState.root) {
+              newState.root = null;
+            }
+            newState.special = null;
+            this.maskVisibility(false);
+            state.changeState(newState);
+            return;
+          }
+          console.error('error redirecting to previous state', error);
```
kadams54: This is more of a personal preference thing, but I prefer to put the error condition in the if block and make the happy path the default, e.g.,:

~~~~
if (error) {
  ...
}
// happy path
~~~~

kadams54:
QA is OK.

jujugui/static/gui/src/app/app.js
hatched: I usually agree with that, but in this case I felt that the execution flow of the code made more sense if it went the opposite here.

hatched:
Thanks all :shipit:

jujugui:

Issue #2697, opened 2017-03-30
Hide the Destroy model button when you don't have admin rights


anthonydillon:
On the profile page. All models have a Destroy model button but error once you click on it. This button should either be hidden or disabled when you only have read access.

PR #2727, opened 2017-04-06
Only display the destroy model button when the user has enough access.

frankban:
Fixes https://github.com/juju/juju-gui/issues/2697

### DIFF
--- a/jujugui/static/gui/src/app/components/user-profile/entity/entity.js
+++ b/jujugui/static/gui/src/app/components/user-profile/entity/entity.js
@@ -24,6 +24,7 @@ YUI.add('user-profile-entity', function() {
     displayName: 'UserProfileEntity',
 
     propTypes: {
+      acl: React.PropTypes.object,
       changeState: React.PropTypes.func,
       children: React.PropTypes.oneOfType([
         React.PropTypes.object,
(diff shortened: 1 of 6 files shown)


frankban:
QA, in gisf, check that the destroy button on the profile page is only displayed for your own models.
Also check that the profile still works in general.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7087/

jujugui/static/gui/src/app/components/user-profile/entity/entity.js
```diff
@@ -257,7 +258,7 @@ YUI.add('user-profile-entity', function() {
       const region = model.region || 'no region';
       let owner = '--';
       if (model.owner) {
-        owner = model.owner = model.owner.split('@')[0];
+        owner = model.owner.split('@')[0];
```
hatched: Thanks for cleaning that up

jujugui/static/gui/src/app/components/user-profile/entity/entity.js
```diff
@@ -318,6 +352,7 @@ YUI.add('user-profile-entity', function() {
           name: name,
           owner: entity.owner
         });
+        destroyButton = this._generateDestroyButton(entity);
```
hatched: because the `_generateDestroyButton` returns `null` could you not just put this method call on line 378 directly?

frankban: Since we already have the isModel check in place for other stuff, I took advantage of that to do one less function call for each entity that are not models.

kadams54:
QA is OK. Thanks @frankban!

frankban:
Thank you all for the reviews!
:shipit:

jujugui:

Issue #2730, opened 2017-04-06
Put the account page behind a feature flag


frankban:
If possible, also fix the error raised when going from an uncommitted state to the account page.

PR #2735, opened 2017-04-07
Improve handling of account page in the case of uncommitted changes.

frankban:
Now it works and it does not throw an error.

Also hide the menu item for the time being, the accountFlag must be active to see the account menu entry.
Also correctly hide the model switcher, cloud logo and model actions when in the account page, as technically we are not in a model at that point, and practically could not be in a model (for instance if /account is accessed directly or from the profile page).

Fixes https://github.com/juju/juju-gui/issues/2730

### DIFF
--- a/jujugui/static/gui/src/app/components/user-menu/user-menu.js
+++ b/jujugui/static/gui/src/app/components/user-menu/user-menu.js
@@ -70,43 +70,51 @@ YUI.add('user-menu', function() {
     },
 
     /**
-     Generate menu based on whether the button has been clicked.
+      Generate menu based on whether the button has been clicked.
 
       @method generateUserMenu
     */
     _generateUserMenu: function() {
-      if (this.state.showUserMenu) {
-        const logoutLink = this.props.LogoutLink;
-        return (
-          <juju.components.Panel
-            instanceName="header-menu__menu"
-            visible={true}>
-              <ul className="header-menu__menu-list" role="menubar">
-                <li className="header-menu__menu-list-item
-                  header-menu__menu-list-item-with-link"
-                  role="menuitem" tabIndex="0">
-                  <a role="button"
-                    onClick={this._handleProfileClick}>Pofile</a>
-                </li>
-                <li className="header-menu__menu-list-item
-                  header-menu__menu-list-item-with-link"
-                  role="menuitem" tabIndex="0">
-                  <a role="button"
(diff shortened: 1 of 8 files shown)


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7099/

frankban:
QA:
- use guiproxy with `-config 'gisf: true, jujuEnvUUID: ""'` and visit the GUI;
- ensure that the account page is NOT visible when clicking the user menu (top right);
- ensure the GUI works as usual (profile page, a model, etc);
- use guiproxy with `-config 'gisf: true, jujuEnvUUID: ""', accountFlag: true` and visit the GUI;
- ensure that the account page is visible when clicking the user menu (top right);
- ensure the GUI works as usual (profile page, a model, etc);
- go to the account page and ensure that it works, and that the cloud logo, the model actions and switcher are NOT visible (for instance they are removed when going to the account from a model canvas);
- switch to a model, and create an uncommitted state, for instance by adding a charm to the canvas (without committing), then go to the account using the menu, ensure the uncommitted alert is displayed, and it works.

kadams54:
QA is OK, thanks!

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7101/

jujugui/static/gui/src/app/components/user-menu/user-menu.js
```diff
@@ -70,43 +70,51 @@ YUI.add('user-menu', function() {
...
       }
-      return '';
+      const logoutLink = this.props.LogoutLink;
+      let accountEntry = null;
+      // For the time being, weakly protect the account page via a flag.
+      // The account page can still be accessed directly by people knowing its
+      // "/account" path. Otherwise, the user menu voice can be activated, for
+      // instance, by running the following guiproxy command:
+      //   guiproxy -config 'gisf: true, jujuEnvUUID: "", accountFlag: true'
+      if (window.juju_config && window.juju_config.accountFlag) {
```
hatched: Typically a component shouldn't rely on globals but have everything passed in. Because this is only going to be very short lived I suppose it's ok ;)

frankban: Yeah I'd not extend that rule to temporary feature flags.

frankban:
Thanks for the reviews!
:shipit:

jujugui:

PR #2743, opened 2017-04-10
Provide timely feedback when destroying a model.

### LINKED ISSUE #2566: Destroying a model in profile view doesn't update model status until you click "Destroy model" a second time
From https://github.com/CanonicalLtd/blues_browser/issues/1442:

Upon clicking destroy model, the model remains in the model view until you click destroy model again. Then the status message of it being destroyed is displayed rather than an option to manage the model.

kadams54:
Fixes #2566.

### DIFF
--- a/jujugui/static/gui/src/app/components/user-profile/model-list/model-list.js
+++ b/jujugui/static/gui/src/app/components/user-profile/model-list/model-list.js
@@ -21,6 +21,8 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
 YUI.add('user-profile-model-list', function() {
 
   juju.components.UserProfileModelList = React.createClass({
+    displayName: 'UserProfileModelList',
+
     // broadcastStatus is necessary for communicating loading status back to
     // the parent SectionLoadWatcher.
     propTypes: {
@@ -45,11 +47,10 @@ YUI.add('user-profile-model-list', function() {
     },
 
     getInitialState: function() {
-      this.xhrs = [];
-
       return {
+        destroyingModels: [],
         modelList: [],
-        loadingModels: false,
+        loadingModels: false
       };
     },
 
@@ -63,12 +64,6 @@ YUI.add('user-profile-model-list', function() {
 
     componentWillMount: function() {
       this._fetchModels(this.props.facadesExist);
-    },
-
-    componentWillUnmount: function() {
-      this.xhrs.forEach((xhr) => {
-        xhr && xhr.abort && xhr.abort();
-      });
     },
 
     componentWillReceiveProps: function(nextProps) {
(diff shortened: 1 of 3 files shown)


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7110/

kadams54:
## QA

1. Fire up `make run` and `guiproxy -controller jimm.jujucharms.com:443 -config 'gisf: true, jujuEnvUUID:""'`.
2. Create a new model and deploy something to it.
3. Go to your profile page.
4. If you're in Chrome, go to the Network tab in your dev tools and throttle your connection to something slow (3G or less). See https://developers.google.com/web/tools/chrome-devtools/network-performance/network-conditions for more details on how to do this. We want to make sure the "destroy model" request doesn't return too quickly. Note that you may need to reload the page after throttling, as I've had the setting not take effect until page reload.
5. Destroy the just-created model.
6. Verify that the model switches to "Requesting that <my-model> be destroyed..." in the list.
7. Verify that the model is removed from the list once the API call returns.

jujugui/static/gui/src/app/components/user-profile/model-list/model-list.js
```diff
@@ -45,11 +47,10 @@ YUI.add('user-profile-model-list', function() {
     },
 
     getInitialState: function() {
-      this.xhrs = [];
```
kadams54: Pulling xhr stuff out was sort of a drive-by cleanup. The controller API calls happen via a websocket, which doesn't return any value, so we're essentially storing an array of nulls right now. I've talked with @hatched and @frankban about how to gracefully cancel any in-flight websockets in response to a componentWillUnmount(), but until that is in place, there's no point having this code for the time being.

jujugui/static/gui/src/app/components/user-profile/model-list/model-list.js
```diff
@@ -193,56 +187,65 @@ YUI.add('user-profile-model-list', function() {
...
       const addNotification = this.props.addNotification;
-      // Handle global errors.
-      if (err) {
+      // Handle global errors or model-specific errors.
+      const error = err || results[uuid];
+      if (error) {
         addNotification({
           title: 'Model destruction failed',
-          message: 'The model failed to be destroyed: ' + err,
+          message: `Could not destroy model "${modelName}": ${error}`,
```
kadams54: Decided to make the messages here a little more user-friendly, as long as I'm improving the feedback.

jaycee:
QA OK. Thanks @kadams54.

jujugui/static/gui/src/app/components/user-profile/model-list/model-list.js
```diff
@@ -193,56 +187,65 @@ YUI.add('user-profile-model-list', function() {
...
+        });
+        destroyedModel.isAlive = false;
       }
-      // Handle a successful deletion.
-      addNotification({
-        title: 'Model destroyed',
-        message: 'The model is currently being destroyed.',
-        level: 'important'
+      // Remove the model UUID from the list of in-flight deletion requests.
+      const destroyingModels = this.state.destroyingModels.filter(id => {
```
mitechie: do we need to do this as well in case of an error? It's not still destroying any more if we've got an error condition is it?

kadams54: I changed the logic flow around within this function so that this bit happens for both errors and success.

kadams54:
Thanks! :shipit:

jujugui:

Issue #2710, opened 2017-04-04
Sandbox mode is gone. Fix 'make run' output.


bac:
In the GUI if you run `make run` you get

```
Starting subprocess with file monitor
Starting server in PID 20877.
serving on http://0.0.0.0:6543
```

Since sandbox mode is gone, hitting that URL will not work. It would be helpful if the output message were changed to give guidance to use guiproxy.

hatched:
This is a good idea @bac 

You are supposed to get an alert when you visit that page https://github.com/juju/juju-gui/blob/e5c0c7987b6484311028fa13843061a2c62a99d4/jujugui/static/gui/src/app/app.js#L2648 but it looks like due to recent changes, there is a typeerror instead.

huwshimi:
In #2749 I've fixed the errors when you visit that URL so you now get the warning. Due to the running process there is no way to output a comment after that URL is printed, but just before that there is:
```
==============================================================
To run the GUI you must point it at a running Juju controller.
The accepted way of doing this is via the GUIProxy project: https://github.com/frankban/guiproxy
==============================================================
```
Hopefully that's enough.

PR #2749, opened 2017-04-11
Fix errors when loading the GUI URL directly

huwshimi:
Fixes: #2710.


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7118/

jujugui/static/gui/src/app/app.js
```diff
@@ -491,7 +491,7 @@ YUI.add('juju-gui', function(Y) {
       let password = null;
       let macaroons = null;
       let credentials = this.user.controller;
-      if (credentials.areAvailable) {
+      if (credentials && credentials.areAvailable) {
```
frankban: As controller is a property of user, this check should never be required, so I am a bit confused.
See https://github.com/juju/juju-gui/blob/develop/jujugui/static/gui/src/app/user/user.js#L44
The areAvailable attribute should always be defined, even if we don;t have credentials.

jujugui/static/gui/src/app/store/env/base.js
```diff
@@ -282,7 +282,7 @@ YUI.add('juju-env-base', function(Y) {
       // Consider the user unauthenticated until proven otherwise.
       this.userIsAuthenticated = false;
       this.failedAuthentication = false;
-      const credentials = this.get('user').controller;
+      const credentials = this.get('user').controller || {};
```
frankban: Ditto. The controller property should always be a proper object or the empty object.

jujugui/static/gui/src/app/app.js
huwshimi: This is the case where someone loads the gui server without using guiproxy so they're not connected to a controller. We just need these sections to not error so that they get give the notice about using guiproxy.

frankban: I understand that, but I am still confused about why in that case user.controller would be null or undefined. It should still be a property right? What I actually encounter when just visiting the python server set up by "make run" is another error: Cannot read property 'bakeryFactory' of undefined.

frankban: Now I also see the areAvailable error in the api object. So there it's ok to check if user is defined I guess. But in the app initializer we have `this.user = this.get('user') || new window.jujugui.User();` so I thing we always have the user object available.

huwshimi: Ah, we do have a user, it's just that it's a string: `admin`

huwshimi: @frankban As fars as the further errors you get, I think we'll continue to get errors as more stuff is not set up correctly, I guess we just need to get it working to the point where that notification can be displayed...

huwshimi:
It looks like the real reason we were getting  these issues was due to some sandbox code still remaining. I've removed all the sandbox code I could find so this needs a new review and QA.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7153/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7154/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7155/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7156/

jujugui/static/gui/src/app/utils/bundle-importer.js
```diff
@@ -165,31 +165,10 @@ YUI.add('bundle-importer', function(Y) {
...
     */
     _fileReaderOnload: function(file, e) {
-      var data;
       var notifications = this.db.notifications;
-      // We support dropping a bundle YAML file and a changeSet JSON file onto
-      // the canvas. The JSON file support should only ever be used
-      // for when there is no guiserver is available like in sandbox mode.
       var extension = file.name.split('.').pop();
       var result = e.target.result;
-      if (extension === 'json') {
```
hatched: Do we really want to remove this functionality?

huwshimi: I was going by this comment: "The JSON file support should only ever be used for when there is no guiserver is available like in sandbox mode."

I can leave it in if we use it at other times...

Lukewh:
:shipit:

jujugui:

PR #2777, opened 2017-04-16
Do not fail when rendering cloud provider logo if none is available

### LINKED ISSUE #2776: juju-gui broken on arm64?
I have a RPi3B cluster and my own custom-built xenial arm64 boot/root disk.  PI 64-bit kernel support isn't complete--some HW accelerations and wifi still don't work--but for my purposes it's functional enough.  

I've installed juju 2.1.2, added a manual cloud, and boostrapped a controller onto the head of the cluster without any problems.  I can deploy charms through the juju client and that appears to work.

The GUI breaks.  I can log in and it displays the application canvas and decorations, but none of the UI elements work--I can't change models, can't inspect the machines or applications list, can't add charms, and can't search the store.  Nothing shows up in the controller logs; breakage is in the browser.  I reproduced this with current Chrome, Safari, and Firefox.  

Attached is a browser log from Chrome 57.0.2987.146.  The GUI logs into the controller fine.  There's an early 404 error for a charm API fetch for a bundle named after my model, but if that's normal behavior then 404 would be expected since the model is empty.  After that the GUI logs into the model fine, but then it throws a TypeError trying to read svgHeight off an undefined object.

I know I'm out on the ragged edge with my system, and I don't expect every charm to work, but I would have expected the GUI to work--or at least, for errors to be server-side, not browser side.

-- T

[192.168.86.50-1492262739764.txt](https://github.com/juju/juju-gui/files/923628/192.168.86.50-1492262739764.txt)

hatched:
If we do not have cloud provider details for the cloud the GUI is loaded on, do not fail when rendering the logo.

Fixes #2776

### DIFF
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -1527,9 +1527,7 @@ YUI.add('juju-gui', function(Y) {
     _renderProviderLogo: function() {
       const container = document.getElementById('provider-logo-container');
       const cloudProvider = this.env.get('providerType');
-      const providerDetails = cloudProvider ?
-        views.utils.getCloudProviderDetails(cloudProvider) :
-        {};
+      let providerDetails = views.utils.getCloudProviderDetails(cloudProvider);
       const currentState = this.state.current || {};
       const isDisabled = (
         // There is no provider.
@@ -1548,6 +1546,11 @@ YUI.add('juju-gui', function(Y) {
         }
       );
       const scale = 0.65;
+      if (!providerDetails) {
+        // It's possible that the GUI is being run on a provider that we have
+        // not yet setup in the cloud provider details.
+        providerDetails = {};
+      }
       ReactDOM.render(
         <div className={classes}>
           <window.juju.components.SvgIcon


hatched:
#### To QA

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7146/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7147/

jujugui/static/gui/src/app/app.js
```diff
@@ -1527,9 +1527,7 @@ YUI.add('juju-gui', function(Y) {
     _renderProviderLogo: function() {
       const container = document.getElementById('provider-logo-container');
       const cloudProvider = this.env.get('providerType');
-      const providerDetails = cloudProvider ?
-        views.utils.getCloudProviderDetails(cloudProvider) :
-        {};
+      let providerDetails = views.utils.getCloudProviderDetails(cloudProvider);
```
mitechie: So the old code only called getCloudProviderDetails if there was a cloudProvider value and now it's calling it all the time. Is this safe? Is this wasted cycles?

hatched: Yeah it's safe, `getCloudProviderDetails` returns undefined when there is no provider found. We need this component to get re-rendered when values change so that the css animations work as expected so we can't simply `unmount` as I was hoping to do if none was provided. This is why the `providerDetails` value gets set to an empty object below.

jaycee:
Per discussion in IRC a follow up branch is going to break out the render methods and add tests for them.

jaycee:
QA OK.

hatched:
Thanks all! :shipit:

jujugui:

PR #2778, opened 2017-04-17
Creates a separate store on user for model creds

jaycee:


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7148/

jaycee:
QA:

On GISF and GIJOE

Ensure login works--loading from '/', logging in from the canvas, and logging in during deployment flow from '/new'.
Ensure that switching between models works.
Ensure that you can create a new model and deploy into it.

jujugui/static/gui/src/app/app.js
```diff
@@ -515,15 +515,7 @@ YUI.add('juju-gui', function(Y) {
...
-      let user = null;
-      let password = null;
-      let macaroons = null;
-      let credentials = this.user.controller;
-      if (credentials.areAvailable) {
-        user = credentials.user;
-        password = credentials.password;
-        macaroons = credentials.macaroons;
-      }
+      const controller = this.user.controller;
```
hatched: This name is a little confusing. What about `controllerUser`?

jujugui/static/gui/src/app/app.js
```diff
@@ -536,7 +528,11 @@ YUI.add('juju-gui', function(Y) {
         entityPromise = this._fetchEntityFromUserState(pathState.state.user);
       }
       this.controllerAPI = this.setUpControllerAPI(
-        controllerAPI, user, password, macaroons, entityPromise);
+        controllerAPI,
+        controller.user,
```
hatched: Instead of this, what do you think about passing the user instance as an argument and having the `setupControllerAPI` method pull out the necessary values? If you do that, then you don't need the above declaration.

jujugui/static/gui/src/app/app.js
```diff
@@ -872,6 +868,11 @@ YUI.add('juju-gui', function(Y) {
       credentials, useMacaroons, apis=[this.controllerAPI, this.env]) {
       if (useMacaroons) {
         apis.forEach(api => {
+          if (credentials && api.name === 'model-api') {
```
hatched: This is done twice in this method can you extract it out into a function? It can even be local to this function.

Why is this done twice now instead of only done where it was before? Can you add a comment as to why this is necessary? If I recall this was done so that the u/p entered in the `Login` component would get added to the user class so this block of it would never evaluate true?

jujugui/static/gui/src/app/user/user.js
```diff
@@ -98,8 +100,39 @@ const User = class User {
...
   set controller(credentials) {
-    this.sessionStorage.setItem('credentials', JSON.stringify(credentials));
+    this._setCredentials('controller', credentials);
+  }
+
+  get model() {
+    // There are situations where we have no model creds but can fall back to
+    // the controller credentials. So we only return empty credentials if both
+    // the model creds and the controller creds are empty.
+    let credentials = this._getCredentials('model');
```
hatched: Nice approach in this class 👍

jujugui/static/gui/src/app/app.js
jaycee: That would be misleading. I'll switch it to controllerCreds, since that's what it is.

jujugui/static/gui/src/app/app.js
jaycee: That seems like a good idea, I'll do it.

jujugui/static/gui/src/app/app.js
jaycee: So it only needed to be in the original place, as it turns out.

jujugui/static/gui/src/app/app.js
jaycee: I can't suss out why, but making this change is really throwing off the anonymousMode tests; given we want to release today and this is just ugly rather than wrong, I'm going to punt on it--we can clean this up more as we continue to clean up app.js and do the user class refactor branches.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7149/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7150/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7151/

hatched:
:shipit: 
Thanks @jaycee

jujugui:

Issue #2147, opened 2016-10-20
When selecting an application, don't move the canvas if the app circle is visible already


bac:
It is nice to be able to click through the apps, but moving the canvas to center them makes it feel jumpy. Let's not move the canvas if the selected app is already fully visible, and if it's partially visible let's just move it enough to bring it onto the canvas. If it's way off, then let's pan the canvas just enough to bring it fully into view.  

Original: https://bugs.launchpad.net/juju-gui/+bug/1634007

spencerbygraves:
Thanks @bac we were talking about this just last week. I think your solution sounds good.

huwshimi:
The first part (not moving the canvas if the app is on screen) has been done in #2782.

PR #2782, opened 2017-04-18
Don't pan to an app if it is already on the screen

huwshimi:


huwshimi:
QA:
- add a number of apps, or deploy a bundle
- move the screen around and drag some apps around the screen and change to different zoom levels
- use the inspector to click on different charms
- the canvas should only pan to a charm if it is off screen

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7159/

jujugui/static/gui/src/app/views/topology/panzoom.js
```diff
@@ -208,16 +208,30 @@ YUI.add('juju-topology-panzoom', function(Y) {
...
+      const pointY = (point[1] * scale) + offset[1];
+      // Find out if the points we have are within the screen rectangle.
+      const withinScreen = pointX > 0 && pointX < screenWidth && pointY > 0 &&
+        pointY < screenHeight;
+      // If the object is not within the screen the move the screen so it is
+      // visible.
+      if (!withinScreen) {
+        this.rescale({
+          scale: scale,
+          translate: point.map(function(d, i) {
```
hatched: could be an arrow fn :)

```JavaScript
translate: point.map((d, i) => ((0 - d) * scale) + (size[i] / 2)))
```

makyo: Mixing new and old (arrowfuncs, let/const vs var) drives me up the wall.

That said, you're right 👍

huwshimi:
Looks like this still needs a QA.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7168/

Lukewh:
## QA-OK(ish)

This PR fixes the "Let's not move the canvas if the selected app is already fully visible" - which is so much better - but not "if it's partially visible let's just move it enough to bring it onto the canvas" or "If it's way off, then let's pan the canvas just enough to bring it fully into view". As when the charm you navigate to is "way off" it does a big'ol jumpathon.

@huwshimi can you add a couple of follow ups to implement the other two cases if they're too much of an undertaking to do as part of this PR, as I still think they're valid? Cheers

huwshimi:
@Lukewh good call, I've left a comment in #2147. We can issue that issue to track the remaining work I think.

huwshimi:
Thanks all :shipit:

jujugui:

PR #2795, opened 2017-04-21
Add Direct Deploy functionality

hatched:
If visiting the GUI with the url `{host}/new/?dd={entityId}` the GUI will now immediately open the deployment flow with the appropriate entity added to the canvas. This is the first step towards a much quicker deploy experience.

### DIFF
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -1367,7 +1367,6 @@ YUI.add('juju-gui', function(Y) {
       const controllerIsAvailable = () =>
         this.controllerAPI &&
         this.controllerAPI.get('connected') &&
-        !this.controllerAPI.pendingLoginResponse &&
         this.controllerAPI.userIsAuthenticated;
       const loginToController =
         controllerAPI.loginWithMacaroon.bind(
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -1360,6 +1360,10 @@ YUI.add('juju-gui', function(Y) {
         const credentials = this.user.controller;
         return credentials ? credentials.user : undefined;
       };
+      const controllerIsAvailable = () =>
+        this.controllerAPI &&
+        this.controllerAPI.get('connected') &&
+        !this.controllerAPI.pendingLoginResponse;
       const loginToController =
         controllerAPI.loginWithMacaroon.bind(
           controllerAPI, this.bakeryFactory.get('juju'));
@@ -1379,6 +1383,7 @@ YUI.add('juju-gui', function(Y) {
             changesUtils.filterByParent.bind(changesUtils, currentChangeSet)}
           changeState={this.state.changeState.bind(this.state)}
           cloud={cloud}
+          controllerIsAvailable={controllerIsAvailable}
           createToken={this.stripe && this.stripe.createToken.bind(this.stripe)}
           createCardElement={
             this.stripe && this.stripe.createCardElement.bind(this.stripe)}
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -731,7 +731,7 @@ YUI.add('juju-gui', function(Y) {
 
         if (state.current.root === 'login') {
           if (dd) {
-            console.log('Initiating Direct Deploy.');
+            console.log('initiating direct deploy');
             this.maskVisibility(false);
             this._directDeploy(dd);
             return;
@@ -2284,15 +2284,15 @@ YUI.add('juju-gui', function(Y) {
       Calls the necessary methods to setup the GUI and put the user in the
       Deployment Flow when they have used Direct Deploy.
 
-      @param {Object} dd - The Direct Deploy data from state.
+      @param {Object} ddData - The Direct Deploy data from state.
     */
     _directDeploy: function(ddData) {
       this.state.changeState({
         special: {dd: null}
       });
-      // Reusing deployTarget state method so faking request signature
+      // The deployTarget method is called directly by state. We're reusing it
+      // here so the state call signature needs to be replicated.
       this._deployTarget({special: {deployTarget: ddData.id}}, ()=>{});
-      // The user must log in first
       this._renderDeployment(this.state, ()=>{}, ddData);
     },
 
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -727,25 +727,34 @@ YUI.add('juju-gui', function(Y) {
         // If state has a `next` property then that overrides all defaults.
         const specialState = current.special;
         const next = specialState && specialState.next;
-        // There should never be a `next` state if we aren't on login but
-        // adding the state login check here just to be sure.
-        if (state.current.root === 'login' && next) {
-          console.log('redirecting to "next" state', next);
-          const {error, state: newState} = state.generateState(next, false);
-          if (error === null) {
-            // The root at this point will be 'login' and because the `next`
-            // url may not explicitly define a new root path we have to set it
-            // to null to clear 'login' from the url.
-            if (!newState.root) {
-              newState.root = null;
-            }
-            newState.special = null;
+        const dd = specialState && specialState.dd;
+
+        if (state.current.root === 'login') {
+          if (dd) {
+            console.log('Initiating Direct Deploy.');
             this.maskVisibility(false);
-            state.changeState(newState);
+            this._directDeploy(dd);
+            return;
+          } else if (next) {
+            // There should never be a `next` state if we aren't on login but
+            // adding the state login check here just to be sure.
+            console.log('redirecting to "next" state', next);
+            const {error, state: newState} = state.generateState(next, false);
+            if (error === null) {
+              // The root at this point will be 'login' and because the `next`
+              // url may not explicitly define a new root path we have to set it
+              // to null to clear 'login' from the url.
+              if (!newState.root) {
+                newState.root = null;
+              }
+              newState.special = null;
+              this.maskVisibility(false);
+              state.changeState(newState);
+              return;
+            }
+            console.error('error redirecting to previous state', error);
             return;
           }
-          console.error('error redirecting to previous state', error);
-          return;
         }
 
         // If the user is connected to a model then the modelList will be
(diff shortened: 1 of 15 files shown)


hatched:
#### To QA

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7182/

jujugui/static/gui/src/app/app.js
```diff
@@ -2255,6 +2278,22 @@ YUI.add('juju-gui', function(Y) {
     },
 
     /**
+      Calls the necessary methods to setup the GUI and put the user in the
+      Deployment Flow when they have used Direct Deploy.
+
+      @param {Object} dd - The Direct Deploy data from state.
```
huwshimi: `ddData`

Also, needs a `@method`?

jujugui/static/gui/src/app/components/deployment-flow/changes/changes.js
```diff
@@ -21,38 +21,49 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+        changes: this.props.getCurrentChangeSet()
+      };
     },
 
-    /**
-      Generate the list of change items.
+    componentDidMount: function() {
+      // poll the changeset every second for updates as there is no way that
+      // we can be sure when a bundle import will be complete.
+      this.interval = setInterval(() => {
```
huwshimi: In the case of a Direct Deploy, we know we need to wait for a bundle import to complete, so is there a way we could fire an event when the bundle import completes that we can listen to here?

jujugui/static/gui/src/app/components/deployment-flow/deployment-flow.js
```diff
@@ -960,6 +967,34 @@ YUI.add('deployment-flow', function() {
...
+      Generates the Direct Deploy component if necessary
+      @returns {Object} The React elements.
+    */
+    _generateDirectDeploy: function() {
+      const props = this.props;
+      const ddEntityId = props.ddData && props.ddData.id;
+      if (!ddEntityId) {
+        return;
+      };
+      let diagram = undefined;
```
huwshimi: Don't we use `null` for this now?

huwshimi:
Probably for an update, but we could probably hide some of the sections such as "Model changes", just strip it back to the info they need to provide for the deploy.

jujugui/static/gui/src/app/app.js
hatched: The new jsdoc doesn't require the `@method` any longer.

jujugui/static/gui/src/app/components/deployment-flow/changes/changes.js
hatched: I didn't like this fix much either :)  but dd can deploy either a bundle or an individual charm and I had found instances where it would take time to load the charm as well and at that point the changes were still incomplete. We need a better way to determine when a set of user actions have been completed but I don't think we're there yet.

jujugui/static/gui/src/app/components/deployment-flow/deployment-flow.js
hatched: Yes, my bad.

hatched:
I was also thinking of hiding the model changes as that would remove that `setTimeout` however because we're creating actions which may incur charges on their cloud provider I felt it best to still have a way to show them the changes we're making - and this also has the side effect of showing the magic of Juju ;)

hatched:
The QA issue you outlined is caused by the DF being rendered and it trying to fetch your credentials before the controller is connected. I'll modify the credentials code to be more resilient to this issue.

jujugui/static/gui/src/app/components/deployment-flow/changes/changes.js
huwshimi: I actually think we should just not show the changes section for a direct deploy... get rid of anything the user doesn't HAVE to see.

jujugui/static/gui/src/app/state/test-state.js
```diff
@@ -29,6 +29,10 @@ describe('State', () => {
     state: {special: {deployTarget: 'cs:trusty/kibana-15'}},
     error: null
   }, {
+    path: 'http://abc.com:123/new/?dd=trusty/kibana-15',
```
frankban: What happens if we have both dd and deployTarget?

hatched: You'll end up in a state where you will get both and the app will choose whichever one it wants to render first. That is an invalid path so I didn't feel it necessary to test it.

jujugui/static/gui/src/app/components/deployment-flow/deployment-flow.js
```diff
@@ -960,6 +967,34 @@ YUI.add('deployment-flow', function() {
     },
 
     /**
+      Generates the Direct Deploy component if necessary
```
hatched: done

jujugui/static/gui/src/app/components/deployment-flow/deployment-flow.js
hatched: done

jujugui/static/gui/src/app/components/deployment-flow/cloud/cloud.js
```diff
@@ -37,28 +40,37 @@ YUI.add('deployment-cloud', function() {
...
+        this.props.listClouds((error, clouds) => {
+          if (error) {
+            console.error('Unable to list clouds', error);
+            return;
+          }
+          let cloudList = [];
+          if (clouds) {
+            cloudList = Object.keys(clouds).map(name => {
+              const cloud = clouds[name];
+              cloud.name = name;
```
hatched: I didn't make any changes here, it was just indented from the debounce change.

jujugui/static/gui/src/app/components/deployment-flow/cloud/cloud.js
```diff
@@ -37,28 +40,37 @@ YUI.add('deployment-cloud', function() {
...
         }
-        let cloudList = [];
-        if (clouds) {
-          cloudList = Object.keys(clouds).map(name => {
-            const cloud = clouds[name];
-            cloud.name = name;
-            return cloud;
+        this.props.listClouds((error, clouds) => {
+          if (error) {
+            console.error('Unable to list clouds', error);
```
hatched: done

jujugui/static/gui/src/app/components/deployment-flow/changes/changes.js
hatched: Ok I'll remove this and then add another indicator as the number of machines this will add in the future to give the user an indication that there might be charges.

jujugui/static/gui/src/app/app.js
```diff
@@ -2255,6 +2278,22 @@ YUI.add('juju-gui', function(Y) {
...
+    */
+    _directDeploy: function(ddData) {
+      this.state.changeState({
+        special: {dd: null}
+      });
+      // Reusing deployTarget state method so faking request signature
+      this._deployTarget({special: {deployTarget: ddData.id}}, ()=>{});
+      // The user must log in first
```
hatched: It was actually just a note to myself that I forgot to remove because I also forgot to put my 'this is a note' tag ;)

jujugui/static/gui/src/app/app.js
```diff
@@ -2255,6 +2278,22 @@ YUI.add('juju-gui', function(Y) {
...
+      Calls the necessary methods to setup the GUI and put the user in the
+      Deployment Flow when they have used Direct Deploy.
+
+      @param {Object} dd - The Direct Deploy data from state.
+    */
+    _directDeploy: function(ddData) {
+      this.state.changeState({
+        special: {dd: null}
+      });
+      // Reusing deployTarget state method so faking request signature
```
hatched: I updated it to be: 

      // The deployTarget method is called directly by state. We're reusing it
      // here so the state call signature needs to be replicated.

jujugui/static/gui/src/app/app.js
hatched: fixed

jujugui/static/gui/src/app/app.js
```diff
@@ -746,25 +746,34 @@ YUI.add('juju-gui', function(Y) {
...
-            // to null to clear 'login' from the url.
-            if (!newState.root) {
-              newState.root = null;
-            }
-            newState.special = null;
+        const dd = specialState && specialState.dd;
+
+        if (state.current.root === 'login') {
+          if (dd) {
+            console.log('Initiating Direct Deploy.');
```

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7230/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7231/

huwshimi:
Two QA things:
- we can probably hide the "Applications to be deployed" section as well, it doesn't really give us anything over the diagram and the list of machines
- when I get the diagram for elk-stack I get:
diagram

jujugui/static/gui/src/app/components/deployment-flow/cloud/cloud.js
```diff
@@ -37,28 +40,37 @@ YUI.add('deployment-cloud', function() {
...
 
     componentWillMount: function() {
-      this.props.listClouds((error, clouds) => {
-        if (error) {
-          console.error('Unable to list clouds', error);
+      const listClouds = () => {
+        // It is possible that the controller hasn't yet connected so
+        // bounce until it's connected then fetch the clouds list.
+        if (!this.props.controllerIsAvailable()) {
+          setTimeout(listClouds, 500);
```
huwshimi: This will fire once, so we assume that the controller will be available after 500ms?

hatched: if the controller is not available it will call this function `listClouds` again, which will check if the controller is available and repeat.

huwshimi: Ah sweet.

hatched:
For follow-up: https://github.com/juju/juju-gui/issues/2828

hatched:
Thanks for all the reviews! :shipit:

jujugui:

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7245/

PR #2803, opened 2017-04-24
DO NOT LAND - for notes only, see new-bakery branch for ongoing changes

hatched:


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7193/

jujugui/static/gui/src/app/store/env/bakery.js
```diff
@@ -18,721 +18,535 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+      @param {String} method The HTTP method, like "get" or "POST".
+      @param {Object} headers Headers that must be included in the request.
+        Note that bakery specific headers are automatically added internally.
+      @param {String} body The request body if it applies, or null.
+      @param {Function} callback A function called when the response is
+        received from the remote URL. It receives a tuple (error, response).
+        If the request succeeds the error is null.
+      @return {Object} the XHR instance.
+    */
+    sendRequest(url, method, headers, body, callback) {
```
hatched: There are a number of positional arguments here, and I see that you're assigning the values to variables in methods below so that the call signatures make sense (yay!). BUT maybe this should instead take a single args object and callback so that we don't have to do that?

jujugui/static/gui/src/app/store/env/bakery.js
```diff
@@ -18,721 +18,535 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+    /**
+      Try to parse the given JSON decoded response in order to retrieve a
+      human friendly error.
+
+      @param {Object} jsonResponse The JSON decoded response text.
+      @return {String} The error message.
+    */
+    _getErrorMessage(jsonResponse) {
+      return (
+        jsonResponse.Message ||
```
hatched: Above you've removed these 'message' responses but kept them here. Was this an oversight?

jujugui/static/gui/src/app/store/env/bakery.js
```diff
@@ -18,721 +18,535 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+    }
+
+  };
+
+  /**
+    A storage for the macaroon bakery.
+
+    The storage is used to persist macaroons.
+  */
+  const BakeryStorage = class BakeryStorage {
```
hatched: I'd prefer if this was in its own file, even if it is small and required by the bakery...single class/file old habit and all that ;)

jujugui/static/gui/src/app/user/user.js
```diff
@@ -160,6 +148,15 @@ const User = class User {
...
     }
   }
+
+  clearMacaroons() {
+    this.localStorage.clear();
+  }
+
+};
+
+const cookieName = service => {
```
hatched: This should be moved up in the file as the above methods are using something defined after their use

jujugui/static/gui/src/app/app.js
```diff
@@ -992,25 +961,26 @@ YUI.add('juju-gui', function(Y) {
...
-          gisf={window.juju_config.gisf || false}
-          charmstoreLogoutUrl={charmstore.getLogoutUrl()}
-          getUser={this.getUser.bind(this, 'charmstore')}
-          clearUser={this.clearUser.bind(this, 'charmstore')}
-          // If the charmbrowser is open then don't show the logout link.
-          visible={!this.state.current.store}
-          locationAssign={window.location.assign.bind(window.location)}
-        />);
+        logout={this.logout.bind(this)}
+        clearCookie={bakery.storage.clear.bind(bakery.storage, () => {})}
```
frankban: The bakery storage contract does not accept the callback anymore, and this is a leftover from my original impl.
So now this should be just `bakery.storage.clear.bind(bakery.storage)`.

jujugui/static/gui/src/app/components/user-profile/user-profile.js
```diff
@@ -64,23 +64,23 @@ YUI.add('user-profile', function() {
...
-    _fetchMacaroonCallback: function(error, macaroon) {
-      if (error) {
-        console.log(error);
+      const props = this.props;
+      const handler = err => {
+        if (err) {
+          console.log('cannot retrieve charm store macaroon:', err);
+          return;
+        }
+        props.storeUser('charmstore', true);
```
frankban: +1 fora less ambiguous naming here.

jujugui/static/gui/src/app/jujulib/index.js
```diff
@@ -41,65 +41,62 @@ var module = module;
...
-       Failure callback that attempts to parse any error messages out of the
-       JSON response, before invoking the user-specified callback.
-
-       @param xhrOrMessage {Object} the XHR response object; can also be a
-                                    plain error string.
-    */
-    var failure = function(xhrOrMessage) {
-      var data;
+      const text = response.target.responseText;
+      if (!text) {
```
frankban: When the server does not behave correctly? As this is a response from an external service it's not a bad idea to be defensive.

jujugui/static/gui/src/app/jujulib/index.js
```diff
@@ -41,65 +41,62 @@ var module = module;
...
+        callback(null, null);
+        return;
+      }
+      // TODO frankban: we should really avoid defaulting optional
+      // parameters to true.
```
frankban: This comment will go away as we get rid of makeResponse. The meaning is that parse is used as an optional parameter (even if in JS all parameters are "optional", in this case we often use makeRequest without passing the parse value), and, if not provided, we assume the response must be parsed. In essence, we assume parse=true. The comment is that we should really try to avoid this behavior, as it is confusing.

jujugui/static/gui/src/app/jujulib/index.js
```diff
@@ -41,65 +41,62 @@ var module = module;
...
-      } catch (e) {
-        data = { error: xhrOrMessage };
+        jsonResponse = JSON.parse(text);
+      } catch(err) {
+        callback(err, null);
+        return;
       }
-      var error = data.Message || data.message || data.Error || data.error;
-      callback(error, data);
+      err = jsonResponse.error || jsonResponse.Error;
```
frankban: We don't know the contract used for errors by the service we are sending requests to. So we try some common patterns, and if we fail, we just use the whole response.

jujugui/static/gui/src/app/store/env/bakery.js
```diff
@@ -18,721 +18,535 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+
+      @param {Object} httpClient The client used to send HTTP requests. It must
+        implement a "send{Method}Request" method for every HTTP method. An
+        instance of WebHandler is usually provided.
+      @param {Object} storage The storage used to persist macaroons. It must
+        implement the following interface:
+          - get(key) -> value;
+          - set(key, value, callback): the callback is called without arguments
+            when the set operation has been performed;
+          - clear(): clear all stored data.
```
frankban: We can remove clear from the storage contract as it is never used by the bakery itself, only by code external to the bakery. So the bakery storage can implement only set and get.

jujugui/static/gui/src/app/store/env/bakery.js
frankban: That's an idea, 5 args is pretty at the limit, let's consider this for a follow up so that for now we reduce the number of changes to the current bakery before ensuring it works?

jujugui/static/gui/src/app/store/env/bakery.js
frankban: I think this function is ok.

jujugui/static/gui/src/app/store/env/bakery.js
frankban: Is "one class for each file" a good old habit, or is it a bad Java limitation instead? ;-)

PR #2821, opened 2017-05-01
Account charges list

huwshimi:


huwshimi:
QA:
- ask @fabricematrat

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7229/

fabricematrat:
QA:

The show receipt on Firefox doesn't work.
The price is wrong on the charge list the number you get from the payment service is in cents on in dollar. So it show a price which is 100 times higher :)

Would be nice to put the line items under the charge.

Fabrice.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7232/

jujugui/static/gui/src/app/jujulib/payment.js
```diff
@@ -456,12 +575,11 @@ var module = module;
...
       @public _parsePaymentMethod
       @param paymentMethod {Object} A payment method reponse from the API.
-      @param parseCardholder {Boolean} Whether to include the card-holder.
       @returns {Object} A parsed payment method object.
     */
-    _parsePaymentMethod: function(paymentMethod, parseCardholder=false) {
+    _parsePaymentMethod: function(paymentMethod) {
       const address = paymentMethod.address;
-      const parsed = {
+      let parsed = {
```
Lukewh: const?

jujugui/static/gui/src/app/components/account/payment/charges/charges.js
```diff
@@ -0,0 +1,262 @@
...
+YUI.add('account-payment-charges', function() {
+
+  juju.components.AccountPaymentCharges = React.createClass({
+    displayName: 'AccountPaymentCharges',
+
+    propTypes: {
+      acl: React.PropTypes.object.isRequired,
+      addNotification: React.PropTypes.func.isRequired,
+      getCharges: React.PropTypes.func.isRequired,
+      getReceipt: React.PropTypes.func.isRequired,
```
Lukewh: `getCharges` and `getReceipt` here are required, but account/account.js doesn't have them marked as required - should they be?

fabricematrat:
Code OK.

We will probably need another PR but we shall discuss as we return from the API all charges, some might be in error state. So we have to see what we display in that case.

jujugui/static/gui/src/app/components/account/payment/charges/charges.js
huwshimi: In the account the payment sections are only displayed under a feature flag under which those methods don't exist, when the sections are displayed the methods do exist (same flag).

jujugui/static/gui/src/app/components/account/payment/charges/charges.js
```diff
@@ -0,0 +1,262 @@
...
+    /**
+      Get the payment details for the user.
+
+      @method _getCharges
+    */
+    _getCharges: function() {
+      this.setState({loading: true}, () => {
+        const username = this.props.username;
+        const xhr = this.props.getCharges(username, (error, response) => {
+          if (error && error !== 'not found') {
```
huwshimi: Oops, that's actually a copy and paste error, but where that if statement is correct I've added a comment (if the user is not found we don't want to display the error, but rather display a message about creating a user.).

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7254/

huwshimi:
@fabricematrat it might be nice if the sorting could be done by the API, but if not I can do it in JS. 

Thanks everyone. :shipit:

jujugui:

jujugui/static/gui/src/app/components/account/payment/charges/charges.js
Lukewh: 👍  thanks for the context

PR #2832, opened 2017-05-02
Add KPI metrics charts 

makyo:


makyo:
This PR supercedes #2832 2796.

QA:
1. Run the gui, and use guiproxy against staging's controller
2. Follow the instructions in `components/user-profile/entity/entity.js:303` to enable using dummy data
3. Visit http://localhost:8042/u/hatch in an incognito window
4. Log into the controller
5. Expand one of the charms, click the KPI metrics button, and behold the chart.
6. Expanding a charm should do nothing.
7. Reverse step 2 and repeat step 5 - there should be no button (and thus no chart)

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7252/

jujugui/static/gui/src/app/components/user-profile/entity/entity.js
```diff
@@ -319,6 +333,102 @@ YUI.add('user-profile-entity', function() {
...
+              console.error(error);
+              return;
+            }
+            this.setState({
+              hasMetrics: false,
+              metrics: {},
+              metricTypes: []
+            });
+            if (charmMetrics.length > 0) {
+              let metrics = [];
```
hatched: These could be a `const` as you're only modifying their internal values, but 🤷‍♂️

jujugui/static/gui/src/app/components/user-profile/entity/entity.js
```diff
@@ -319,6 +333,102 @@ YUI.add('user-profile-entity', function() {
...
+                metrics: metrics,
+                metricTypes: metricTypes
+              });
+            }
+          });
+    },
+
+    /**
+      Show the metrics component, gated on state.
+    */
```
hatched: `@return`

jujugui/static/gui/src/app/components/user-profile/entity/entity.js
```diff
@@ -319,6 +333,102 @@ YUI.add('user-profile-entity', function() {
...
+    _toggleKpiVisibility: function() {
+      this.setState({
+        kpiVisible: !this.state.kpiVisible
+      });
+    },
+
+    /**
+      For charms, generate a button for showing/hiding the metrics component.
+    */
```
hatched: `@return`

jujugui/static/gui/src/app/components/user-profile/entity/entity.js
```diff
@@ -358,6 +468,9 @@ YUI.add('user-profile-entity', function() {
         });
         destroyButton = this._generateDestroyButton(entity);
       } else {
+        if (isCharm) {
+          this._getMetrics();
```
hatched: In a previous review we questioned polling for the metrics every time the component is rendered, which could be often. I think this should go in the `componentDidMount` section.

jujugui/static/gui/src/app/components/user-profile/entity/test-entity.js
```diff
@@ -391,13 +393,68 @@ describe('UserProfileEntity', () => {
...
+    expect(kpiDiv).toEqualJSX(expectedButton);
+    instance.setState({kpiVisible: true});
+    kpiDiv = renderer.getRenderOutput().props.children[1]
+			.props.children[1].props.children[7];
+    const expectedChart = (
+      <div>
+        <juju.components.GenericButton
+          action={function noRefCheck() {}}
+          title="Show KPI Metrics" />
+				<juju.components.UserProfileEntityKPI
```
hatched: Looks like there are a number of tabs around here, spaces only plz.

jujugui/static/gui/src/app/components/user-profile/entity/entity.js
```diff
@@ -319,6 +333,102 @@ YUI.add('user-profile-entity', function() {
...
+          // 'cs:~canonical/jimm-0',
+          this.props.entity.id,
+          filters,
+          (error, charmMetrics) => {
+            if (error) {
+              // TODO When there are designs for showing errors for metrics,
+              // we'll be able to implement them here.
+              // Makyo - 2017-04-13
+              console.error(error);
+              return;
```
hatched: So if there is an error we don't set the state as on `367`?

jujugui/static/gui/src/app/components/user-profile/entity/entity.js
```diff
@@ -319,6 +333,102 @@ YUI.add('user-profile-entity', function() {
...
+            }
+            this.setState({
```
hatched: Setting state causes a re-render which will cause this method to be called again (see comment below).

I think this setState should probably be moved to an else for the conditional on `372`

jujugui/static/gui/src/app/components/user-profile/kpi/kpi.js
```diff
@@ -0,0 +1,88 @@
...
+You should have received a copy of the GNU Affero General Public License along
+with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+'use strict';
+
+YUI.add('user-profile-entity-kpi', function() {
+
+  juju.components.UserProfileEntityKPI = React.createClass({
+    displayName: 'Profile KPIs',
```
hatched: we'll want to use the module name here to make it easier to find: `UserProfileEntityKPI`

jujugui/static/gui/src/app/components/user-profile/kpi/metric.js
```diff
@@ -0,0 +1,108 @@
...
+You should have received a copy of the GNU Affero General Public License along
+with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+'use strict';
+
+YUI.add('user-profile-entity-metric', function() {
+
+  juju.components.UserProfileEntityMetric = React.createClass({
+    displayName: 'Profile entity metrics',
```
hatched: see above about `displayName`

jujugui/static/gui/src/app/components/user-profile/kpi/metric.js
```diff
@@ -0,0 +1,108 @@
...
+        .attr('x', 250)
+        .attr('y', 20);
+      yg.call(yAxis);
+      xg.call(xAxis);
+      chart.append('path')
+        .classed('mean', true)
+        .attr('d', line);
+    },
+
+    componentDidUpdate: function() {
```
hatched: `componentDidUpdate` isn't called on initial render so you'll probably also want to call `_renderChart` in `componentDidMount`

https://facebook.github.io/react/docs/react-component.html#componentdidupdate

jujugui/static/gui/src/app/jujulib/index.js
```diff
@@ -103,6 +103,17 @@ var module = module;
     }
   };
 
+  /**
+    Serializes an object into a query string.
+
+    @param obj {Object} the object to serialize
+    @return a query string serialized from the object.
+  */
+  var serializeObject = function(obj) {
```
hatched: `const`

jujugui/static/gui/src/app/components/user-profile/entity/entity.js
makyo: Deferring.

jujugui/static/gui/src/app/components/user-profile/entity/test-entity.js
makyo: ARgh, come on, vim...

jujugui/static/gui/src/app/components/user-profile/kpi/kpi.js
makyo: Bleh, okay. Still confused as to what this is, but by now, I'm just doing whatever you tell me.

jujugui/static/gui/src/app/components/user-profile/kpi/metric.js
makyo: I understand, and yet here we are. It renders on load, and I'm unsure of what *also* calling it there will brook us.

jujugui/static/gui/src/app/jujulib/index.js
makyo: I hate these piecemeal changes, but as you wish.

jujugui/static/gui/src/app/components/user-profile/entity/entity.js
makyo: Correct. We have the ability to set an error down the line once we know more about the API.

hatched: Should we not set the state to the 'falsy' state so that it can be rendered accordingly on error?

jujugui/static/gui/src/app/components/user-profile/entity/entity.js
hatched: Because the app re-renders on every db update this will cause an xhr request every 5s or so when it's open so I'm not sure it's a good idea to defer.

jujugui/static/gui/src/app/components/user-profile/kpi/kpi.js
hatched: https://facebook.github.io/react/docs/react-component.html#displayname

jujugui/static/gui/src/app/components/user-profile/kpi/metric.js
hatched: it's probably being shown on the second render, if you put a log in the render method you'll probably see it being rendered multiple times.

jujugui/static/gui/src/app/components/user-profile/kpi/kpi.js
jaycee: For my own edification: it says in the link it's set automatically by JSX; does JSX not set it to basically the component name in code?

hatched: Because we're using the 'old' `createClass` we have to manually specify it. If this was an es6 class it would "just work(tm)"

jujugui/static/gui/src/app/components/user-profile/entity/entity.js
makyo: I misread, can do.

makyo: (I thought you meant moving the functionality, not the function call)

jujugui/static/gui/src/app/components/user-profile/kpi/metric.js
makyo: So...both? I need to be able to change chart when the metric is changed, which would be DidUpdate, right? Unless the entire component is torn down and rebuilt?

jujugui/static/gui/src/app/components/user-profile/entity/entity.js
makyo: Only set state when we have no metrics? That doesn't seem right.

makyo: This would mean that we only set state when we have no metrics, which feels wrong.

makyo: nvm, understood.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7262/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7263/

jaycee:
QA notes:

Running this against jujugui via guiproxy, I see the following results on /u/yellow's charm list:
2017-05-03-130841_1031x868_scrot

Logstash has no metrics and nothing is supplied. uptime, provided by @cmars, does have metrics, and they are displayed.

QA OK.

makyo:
:shipit:

jujugui:

makyo:
(oh god please work)

PR #2838, opened 2017-05-03
Implement bakery.withoutDischarge.

frankban:
Also fix bakery tests and lint.

### DIFF
--- a/jujugui/static/gui/src/app/jujulib/bakery.js
+++ b/jujugui/static/gui/src/app/jujulib/bakery.js
@@ -62,8 +62,6 @@ var module = module;
           - set(key, value, callback): the callback is called without arguments
             when the set operation has been performed.
       @param {Object} params Optional parameters including:
-        - nonInteractive: whether to switch to non-interactive mode while
-          acquiring macaroons. If set the visitPage parameter below is ignored;
         - onSuccess: a function to be called when the request completes
           properly. It defaults to a no-op function.
         - visitPage: the function used to visit the identity provider page when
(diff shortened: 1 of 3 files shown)


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7260/

jujugui/static/gui/src/app/jujulib/test-bakery.js
```diff
@@ -248,7 +262,7 @@ describe('Bakery', () => {
 
     it('handles discharge if needed', () => {
       const dischargeStub = sinon.stub(bakery, 'discharge');
-      const cb = sinon.stub();
```
jaycee: Hm. I guess our js linter isn't picking up trailing white space anymore?

hatched: Hmm nope, doesn't look like it - I can add it into develop

jujugui/static/gui/src/app/user/user.js
```diff
@@ -139,8 +139,4 @@ const User = class User {
 
 };
 
-const cookieName = service => {
```
hatched: Heh, I also removed this in my local branch

jujugui/static/gui/src/app/jujulib/bakery.js
```diff
@@ -73,14 +71,28 @@ var module = module;
...
-        this._visitPage = this._nonInteractiveVisitPage;
-      } else {
-        this._visitPage = params.visitPage || (visitURL => {
-          window.open(visitURL, 'Login');
-        });
-      }
+      this._visitPage = params.visitPage || (visitURL => {
+        window.open(visitURL, 'Login');
+      });
+      this._dischargeDisabled = false;
```
hatched: Why not have this configurable in the constructor, then in `withoutDischarge` below you can set it as a value there.

frankban: Because I didn't want users to be able to instantiate a bakery without discharge directly. As this is a really specific need that could even go away in the future, I think it's preferable to only have one way of creating such a bakery.

frankban:
Thanks for the reviews!

PR #2878, opened 2017-05-09
Remove YUI from login events

huwshimi:


huwshimi:
QA:
- log in in as many different ways as you can think of
- you should get logged in

jujugui/static/gui/src/test/test_app.js
```diff
@@ -638,7 +639,7 @@ describe('App', function() {
       conn.msg({'request-id': 1, error: 'bad wolf'});
       assert.equal(1, conn.messages.length);
       assertIsLogin(conn.last_message());
-      assert.equal(loginStub.callCount, 1);
+      assert.equal(loginStub.callCount >= 1, true);
```
hatched: Isn't this the one we only want to have fired once?

jujugui/static/gui/src/app/app.js
```diff
@@ -2632,8 +2639,13 @@ YUI.add('juju-gui', function(Y) {
       @private
     */
     onLogin: function(evt) {
-      if (evt.err) {
-        this._renderLogin(evt.err);
+      if (this.boundOnLogin) {
+        // We want this event only fire once, so if it exists then remove it so
```
hatched: "...this event to only fire..."

jujugui/static/gui/src/test/test_env_api.js
```diff
@@ -420,15 +422,17 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
       it('fires a login event on failed login', function() {
         let fired = false;
         let err;
-        env.on('login', evt => {
+        const listener = evt => {
           fired = true;
```
hatched: This is a little scary as it's requiring an async event process to execute synchronously...these tests should all be upgraded to use `done()` but since this seams to work maybe we just do that as a future follow-up (with an issue created)

jujugui/static/gui/src/test/test_app.js
huwshimi: I think this is OK, the app init sets up two listeners, one a generic on login event and one for the controller api, both of which call `_renderLogin` if there is an error. I'm not sure why this test didn't catch both events before, maybe something to do with their use of `after(` which we no longer have.

jujugui/static/gui/src/test/test_env_api.js
huwshimi: I've updated these.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7329/

jaycee:
QA:

_GISF_:
Login worked normally

_Bootstrapped lxd w/ identity provider_:
Logging in with admin/password: Worked, but the gui didn't update once the model loaded:
2017-05-11-101639_1354x407_scrot

Switching to the profile view then kept the model connecting message:
2017-05-11-101651_291x165_scrot

Logging in via usso failed; this is also true in `develop`, so not this branch's fault.

_Bootstrapped lxd w/o identity provider_:
Logging in via user/pass worked, but again the gui failed to update after the model connected.

huwshimi:
Thanks @jaycee I tracked down the issue, would you giving it another go?

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7349/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7350/

Lukewh:
Hesitant to add QA-OK as I saw the error @jaycee mentioned once, but then couldn't reproduce. Would be good for @jaycee to QA again to confirm whether the error is truly fixed.

jaycee:
I'm redoing QA now.

jaycee:
I could not reproduce the earlier issue; everything looked good.

QA OK.

Lukewh:
Thanks @jaycee

huwshimi:
Thanks everyone :shipit:

jujugui:

Issue #1311, opened 2016-02-01
Deploying the same bundle twice causes issues


anthonydillon:
To replicate:
- Go to http://0.0.0.0:6543/services/?search&type=bundle
- Click on Mediawiki Single
- Add to canvas
- Then do the same again and add another Mediawiki Single bundle
- You should see `Uncaught TypeError: Cannot read property 'get' of undefined` in the console
- The second bundle has no relations
- The inspector does not should the new charms as uncommitted

PR #2899, opened 2017-05-12
Fix adding multiple bundles

huwshimi:
Fixes: #1311.

### DIFF
--- a/jujugui/static/gui/src/app/utils/bundle-importer.js
+++ b/jujugui/static/gui/src/app/utils/bundle-importer.js
@@ -681,7 +681,7 @@ YUI.add('bundle-importer', function(Y) {
         endpoints[index][0] = record[ep[0].replace(/^\$/, '')].get('id');
       }, this);
 
-      const relationId = 'pending-' + record.args[0] + record.args[1];
+      const relationId = 'pending-' + endpoints[0][0] + endpoints[1][0];
       const pendingRelation = this.db.relations.add({
         relation_id: relationId,
         'interface': endpoints[0][1].name,
--- a/jujugui/static/gui/src/app/utils/bundle-importer.js
+++ b/jujugui/static/gui/src/app/utils/bundle-importer.js
@@ -681,8 +681,10 @@ YUI.add('bundle-importer', function(Y) {
         endpoints[index][0] = record[ep[0].replace(/^\$/, '')].get('id');
       }, this);
 
-      // Use the updated app names here to avoid conflicts if apps with the
-      // same name already exist.
+      // Create the pending relation id with a combination of the interfaces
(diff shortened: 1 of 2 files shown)


huwshimi:
QA:
- add a bundle
- add it again, it should finish adding, with no errors in your console and the app circles should not appear on top of each other

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7342/

hatched:
In the CLI deploying the same bundle again upgrades the existing applications....Maybe we want to mirror that functionality in the GUI instead of adding everything again?

huwshimi:
@hatched that's fine, if we want to do an upgrade I'll close this and open a new issue. How does the CLI know if a bundle has been deployed before? I don't think we have a way to know that in the GUI.

frankban:
There is no way to know that in the CLI as well. There is no concept of bundle internally in Juju.
The logic in the CLI is like:
- try to deploy an app;
- if the app is already there, with that name
  - is the bundle charm is compatible with the charm of the already deployed app?
    - if not, error
    - if so, upgrade the charm if required and continue

Later, the CLI check if the number of units declared in the bundle are already deployed for all applications. Be aware that the logic in the CLI is not straightforward, and will have to be basically duped in the GUI.

huwshimi:
So, I'll close this and open a new issue about doing the upgrade? I do wonder if the GUI case is a bit different to the CLI as we have uncommitted state, but happy either way.

huwshimi:
The other option would be to land this so we fix the breakage and in the future think about adding an "add as new or upgrade existing bundle" confirmation.

frankban:
I agree with you Huw. The proper fix can be done separately. In the meanwhile, if console errors are fixed by this branch, I'd say let's land it. Code looks good.

jujugui/static/gui/src/app/utils/bundle-importer.js
```diff
@@ -681,7 +681,10 @@ YUI.add('bundle-importer', function(Y) {
         endpoints[index][0] = record[ep[0].replace(/^\$/, '')].get('id');
       }, this);
 
-      const relationId = 'pending-' + record.args[0] + record.args[1];
+      // Use the updated app names here to avoid conflicts if apps with the
+      // same name already exist.
```
hatched: I don't follow this comment....updated app names from where? Maybe expand on this comment a bit?

hatched:
QA OK, I was able to deploy the elastic search bundle, then add it again to the canvas and deploy again.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7369/

huwshimi:
Thanks @frankban and @hatched :shipit:

jujugui:

Issue #1982, opened 2016-09-08
Constraints on uncommitted machines should be visible (and editable)


johnsca:
When an uncommitted machine is pending, any constraints specified are not visible.  It would also be nice if the pending constraints could be modified in place without removing the pending machine entirely and re-assigning the unit to a new machine (having to specify all of the constraints from scratch, rather than making just the changes that are needed).

This is most useful when tweaking a bundle just prior to deployment.

hatched:
Hi @johnsca thanks for bringing this up, this would be a great feature to add! I'll bring it up to the rest of the team and see what we can come up with.

Related #1379

kasiag:
adding @cassiocassio

Issue #1379, opened 2016-02-23
The constraints specified when adding a new machine should be reflected on the machine once it is created


elvirafigueras:

hatched:
What you're specifying are indeed constraints, this means that Juju will fetch you the 'best' machine available that matches what you have specified. This means that when you create a machine with constraints we don't really know what the actual machine will be. We could of course display these values in the same place as the actual machine values but this may cause some confusion, maybe we should have some indicator that they are constraints and not the 'actual' machine specs.

elvirafigueras:
Ok. This means that the machine you have added will have those constraints even though it wasn't created with those exact ones, right? 
It would be good to show the constraints that users have specified on the machine as otherwise they can't refere back to them. We might need design for this? @Lucapaulina

Lucapaulina:
Do constraints have any ramification on the machine?

There has been some talk about providing users with sizes instead of constraints, maybe we should look into that.

PR #2911, opened 2017-05-17
Machine constraints

huwshimi:
Fixes: #1379. Fixes: #1982.

### DIFF
--- a/jujugui/static/gui/src/app/components/machine-view/machine/machine.js
+++ b/jujugui/static/gui/src/app/components/machine-view/machine/machine.js
@@ -172,40 +172,25 @@ YUI.add('machine-view-machine', function() {
     */
     _getConstraints: function() {
       const constraints = this.props.machine.constraints || '';
-      let cpu;
-      let disk;
-      let mem;
-      let cpuCores;
-      let arch;
+      let types = {
+        'cpu-power': null,
+        cores: null,
+        mem: null,
+        'root-disk': null,
+        arch: null
+      };
       // The constraints are in the format:
       // cpu-power=w cores=x mem=y root-disk=z
-      const parts = constraints.split(' ');
-      parts.forEach(part => {
+      constraints.split(' ').forEach(part => {
         const keyVal = part.split('=');
-        switch (keyVal[0]) {
-          case 'cpu-power':
-            cpu = keyVal[1];
-            break;
-          case 'cores':
-            cpuCores = keyVal[1];
-            break;
-          case 'mem':
-            mem = keyVal[1];
-            break;
-          case 'root-disk':
-            disk = keyVal[1];
-            break;
-          case 'arch':
-            arch = keyVal[1];
-            break;
-        }
+        types[keyVal[0]] = keyVal[1];
       });
       return {
-        arch: arch || null,
-        cpuCores: cpuCores || null,
-        cpuPower: cpu || null,
-        disk: disk || null,
-        mem: mem || null
+        arch: types.arch,
+        cpuCores: types.cores,
+        cpuPower: types['cpu-power'],
+        disk: types['root-disk'],
+        mem: types.mem
       };
     },
 
--- a/jujugui/static/gui/src/app/components/machine-view/machine/machine.js
+++ b/jujugui/static/gui/src/app/components/machine-view/machine/machine.js
@@ -91,27 +91,70 @@ YUI.add('machine-view-machine', function() {
       if (this.props.type === 'container' || !this.props.showConstraints) {
         return;
       }
-      var machine = this.props.machine;
-      var hardware = machine.hardware;
-      var hardwareDetails;
+      let machine = this.props.machine;
+      let constraints = machine.constraints;
+      let hardware = machine.hardware;
+      let hardwareDetails;
+      let cpu;
+      let disk;
+      let mem;
+      let cpuCores;
       if (hardware) {
-        var cpu = hardware.cpuPower;
-        var disk = hardware.disk;
-        var mem = hardware.mem;
-        var cpuCores = hardware.cpuCores;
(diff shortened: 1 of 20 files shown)


huwshimi:
QA:
- start a new model
- open the machine view and add a machine with some constraints
- the constraints should be visible on the added machine
- use the context menu on the machine to open the update constraints form
- change some constraints
- the new constraints should be visible on the machine
- open the deployment flow
- scroll down to the machines section
- the machine constraints should be the same as the updated constraints

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7357/

jujugui/static/gui/src/app/components/machine-view/machine/machine.js
```diff
@@ -71,13 +73,140 @@ YUI.add('machine-view-machine', function() {
...
+            break;
+        }
+      });
+      return {
+        arch: arch || null,
+        cpuCores: cpuCores || null,
+        cpuPower: cpu || null,
+        disk: disk || null,
+        mem: mem || null
+      };
```
Lukewh: Might be cleaner to do something like:
```
let types = {
  'cpu-power': null,
  cores: null,
  mem: null,
  'root-disk': null,
  arch: null;
};
```
So you can remove the switch and do
```
const keyVal = part.split('=');
types[keyVal[0]] = keyVal[1];
```
Then assign to what you want in the return
```
return {
  arch: types.arch,
  cpuCores: types.cores,
  cpuPower: types['cpu-power'],
  disk: types['root-disk'],
  mem: mem
};
```
Saves a few lines.
The ideal in this situation is to use a consistent naming convention, but that's not always possible.

Lukewh: Worth noting, I don't like switch statements unless it's for numbers.

huwshimi: So much better, thankyou!

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7362/

jujugui/static/gui/src/app/components/machine-view/machine/machine.js
```diff
@@ -71,13 +73,125 @@ YUI.add('machine-view-machine', function() {
...
+      this.setState({constraints: constraints});
+    },
+
+    /**
+      Set the new constraints on the machine.
+    */
+    _setConstraints: function() {
+      const constraints = this.state.constraints;
+      const series = constraints.series || null;
+      delete constraints.series;
```
hatched: why? Can you add a code comment plz

jujugui/static/gui/src/app/components/machine-view/machine/machine.js
```diff
@@ -88,30 +202,47 @@ YUI.add('machine-view-machine', function() {
...
+      const machine = this.props.machine;
+      const hardware = machine.hardware || this._getConstraints() || {};
+      let hardwareDetails;
+      let cpu = hardware.cpuPower;
+      let disk = hardware.disk;
+      let mem = hardware.mem;
+      let cpuCores = hardware.cpuCores;
+      if (cpuCores || cpu || disk || mem) {
+        let details = [];
+        if (cpuCores) {
```
hatched: I wonder if these checks to generate this string shouldn't be moved to a util as you've duplicated it twice now.

jujugui/static/gui/src/app/components/machine-view/machine/machine.js
```diff
@@ -71,13 +73,125 @@ YUI.add('machine-view-machine', function() {
...
+    _getConstraints: function() {
+      const constraints = this.props.machine.constraints || '';
+      let types = {
+        'cpu-power': null,
+        cores: null,
+        mem: null,
+        'root-disk': null,
+        arch: null
+      };
+      // The constraints are in the format:
```
hatched: Why are the constraints a string here?

jujugui/static/gui/src/app/components/machine-view/machine/machine.js
```diff
@@ -71,13 +73,125 @@ YUI.add('machine-view-machine', function() {
...
+    },
+
+    /**
+      Get the constraints for the machine.
+
+      @returns {Object} the machine constraints.
+    */
+    _getConstraints: function() {
+      const constraints = this.props.machine.constraints || '';
+      let types = {
```
hatched: Do we not have a master list of these types somewhere? I imagine that would go well in jujulib

hatched: There are probably other places in the code base that use these, maybe a follow-up to centralize?

huwshimi: You're right we have a list in api.js. I'll update this to use that.

jujugui/static/gui/src/app/components/machine-view/machine/machine.js
huwshimi: The machine stores the constraints as a string. I've updated the comment to clarify that.

As to why that's the case, I'm not entirely sure.

jujugui/static/gui/src/app/components/machine-view/machine/machine.js
huwshimi: Done.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7370/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7371/

huwshimi:
Thanks @Lukewh and @hatched :shipit:

jujugui:

PR #2939, opened 2017-05-24
Updates serialization to properly convert to and from full macaroons

jaycee:


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7716/

hatched:
The changes look good but should there not have been test updates as the output formats have changed?

jujugui/static/gui/src/app/jujulib/test-bakery.js
```diff
@@ -408,6 +408,16 @@ describe('Bakery', () => {
...
+    let generateStub;
+
+    beforeEach(() => {
+      generateStub = sinon.stub(macaroonlib, 'generateMacaroons');
+      generateStub.returnsArg(0);
+    });
+
+    afterEach(() => {
+      generateStub.restore();
+    });
```
hatched: \n for some breathing room :)

jaycee:
QA:

Login should work. The profile page should load.

jaycee:
Per a conversation in IRC, the format changes don't need further test updates b/c tests start falling over if we're inconsistent in how we serialize/deserialize.

frankban:
The set-auth-cookie call to the charm store fails: `{"Message":"cannot unmarshal macaroons: signature has unexpected length 0","Code":""}`. So I guess we need to update how macaroons are serialized when doing that call?

jaycee:
This is no longer needed; a simpler fix is available.

PR #2980, opened 2017-06-05
Update login events to prevent model login loop

### LINKED ISSUE #2968: GUI gets into infinite loop after deploying to a new model
GUI gets into infinite loop after deploying to a new model.

To reproduce:
- Use prod/staging jimm
- open the devtools 
- connect to a model
- click to create new
- add a charm to the model and deploy
- once deployed, switch models via the model switcher
it'll now loop

hatched:
Fixes #2968

We were relying on non-namespaced events when the controller and model logged in which was causing a race condition when logging in. This would mean that sometimes the login state would be re-instated because it would get the login event from the controller but then check the model and it would still be connecting which would cause it to loop back to logging in.

### DIFF
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -2536,9 +2536,6 @@ YUI.add('juju-gui', function(Y) {
       document.removeEventListener(
         'ecs.currentCommitFinished', this.renderDeploymentBarListener);
       document.removeEventListener('login', this.onLoginHandler);
-      if (this.boundOnLogin) {
-        document.removeEventListener('login', this.boundOnLogin);
-      }
       document.removeEventListener('login', this.controllerLoginHandler);
       document.removeEventListener('delta', this.onDeltaBound);
     },
@@ -2686,11 +2683,6 @@ YUI.add('juju-gui', function(Y) {
       @private
     */
     onLogin: function(evt) {
-      if (this.boundOnLoginFired) {
-        // We want this event to only fire once, so if it exists then remove it
-        // so it can't get fired again.
-        document.removeEventListener('login', this.boundOnLogin);
-      }
       if (evt.detail && evt.detail.err) {
         this._renderLogin(evt.detail.err);
         return;
@@ -2781,7 +2773,6 @@ YUI.add('juju-gui', function(Y) {
       };
       const credentials = this.user.model;
       const onLogin = callback => {
-        this.boundOnLoginFired = true;
         this.env.loading = false;
         if (callback) {
           callback(this.env);
@@ -2789,9 +2780,8 @@ YUI.add('juju-gui', function(Y) {
       };
       // Delay the callback until after the env login as everything should be
       // set up by then.
-      this.boundOnLoginFired = false;
-      this.boundOnLogin = onLogin.bind(this, callback);
-      document.addEventListener('login', this.boundOnLogin);
+      document.addEventListener(
+        'model.login', onLogin.bind(this, callback), {once: true});
       if (clearDB) {
         // Clear uncommitted state.
         this.env.get('ecs').clear();
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -2664,32 +2664,6 @@ YUI.add('juju-gui', function(Y) {
     },
 
     /**
-      Hide the login mask and redispatch the router.
-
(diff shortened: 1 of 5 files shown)


hatched:
#### To QA

1) Reproduce the steps in the bug, it should not loop.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7764/

jujugui/static/gui/src/app/app.js
```diff
@@ -466,10 +466,6 @@ YUI.add('juju-gui', function(Y) {
           this.onEnvironmentNameChange, this);
       this.env.after('defaultSeriesChange', this.onDefaultSeriesChange, this);
 
-      // Once the user logs in, we need to redraw.
-      this.onLoginHandler = this.onLogin.bind(this);
-      document.addEventListener('login', this.onLoginHandler);
```
hatched: We now rely on the callbacks and state modifications for the UI updates from login

jujugui/static/gui/src/app/app.js
```diff
@@ -2651,18 +2643,15 @@ YUI.add('juju-gui', function(Y) {
...
         // If the api is connecting then we can't know if they are properly
         // logged in yet.
         if (api.get('connecting')) {
-          return true;
-        }
-        if (!api || !api.get('connected')) {
-          // If we do not have an api instance or if we are not connected with
-          // it then we don't need to concern ourselves with being
-          // authenticated to it.
           return false;
```
hatched: If we're still connecting then do not show the login screen as we're often logged into the controller but not the model causing the app to dispatch incorrectly. It's almost impossible to be logged into a controller but then not be able to log into a model you have access to as per the controller.

jujugui/static/gui/src/app/app.js
```diff
@@ -2675,37 +2664,6 @@ YUI.add('juju-gui', function(Y) {
...
-
-      When the environment gets a response from a login attempt,
-      it fires a login event, to which this responds.
-
-      @method onLogin
-      @param {Object} evt An event object that includes an "err" attribute
-        with an error if the authentication failed.
-      @private
-    */
-    onLogin: function(evt) {
```
hatched: No longer used, state updates now handle this cleanup.

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -569,7 +569,7 @@ YUI.add('juju-env-api', function(Y) {
       // Only fire login if this is not a redirect error as we will come back
       // here once the redirection is made.
       if (!utils.isRedirectError(data.error)) {
-        document.dispatchEvent(new CustomEvent('login', {
+        document.dispatchEvent(new CustomEvent('model.login', {
```
hatched: namespaced events so that they don't conflict with the controller 'login' event.

jujugui/static/gui/src/app/app.js
jaycee: Was that added before? I don't see changes in this diff for that.

hatched: Yeah the state changes were done quite some time ago

jujugui/static/gui/src/test/test_app.js
```diff
@@ -596,45 +596,6 @@ describe('App', function() {
...
-        controllerAPI: controller,
-        env: env,
-        controllerSocketTemplate: '/api',
-        viewContainer: container,
-        jujuCoreVersion: '2.0.0'
-      });
-      env.connect();
-      localApp.env.userIsAuthenticated = true;
-      localApp.env.login();
-    });
```
hatched: Yeah because the `onLogin` method is removed and duplicated functionality the tests which tested that functionality are no longer needed.

jujugui/static/gui/src/test/test_app.js
```diff
@@ -1512,12 +1473,12 @@ describe('App', function() {
       checkNext({root: 'store'});
     });
 
-    it('displays login if one of the apis is still connecting', () => {
```
hatched: I'm pretty confident that it was an incorrect assumption based on the idea that you could be unable to log into a model the controller said you could - which isn't actually possible.

makyo:
QA OK here, as well. Wasn't able to produce the loop.

hatched:
Thanks peeps! :shipit:

jujugui:

PR #2990, opened 2017-06-12
Created a new initialization script for the GUI

hatched:
In our continued drive to remove YUI from the GUI codebase this branch implements a new initialization class to replace the old app.js. By making this change this will allow us much easier testing of init.js and renderers compared to the old monolithic app.js and once the transformation is complete it'll also give us the ability to split and defer loading of modules to dramatically speed up the loading of the GUI.

Going forward, any modifications to app.js should also be evaluated in the init.js and required modules so we don't inadvertently introduce a regression when we switch over.

- A new init.js file has been created and all of the functionalities have been copied over, rewritten, and modified where necessary to make the execution easier. 
- Some code has been grouped and moved into mixins or libraries which are consumed by the new JujuGUI class. Additional groupings are possible and will be done in the future.
- This init.js code can exist along side the existing app.js until it's fully tested and reaches feature partiy
- To build a GUI with the new init.js follow instructions at the top of the init.js file. Known todo and issues are also to be included here.
- Two new dependencies, browserify and babelify, have been added to allow us to build a bundle and require modules.


hatched:
#### To QA
QA the original GUI as usual, this code should not be loaded and as such the GUI should work as expected.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7784/

jujugui/static/gui/src/app/components/header-breadcrumb/_header-breadcrumb.scss
```diff
@@ -61,6 +61,7 @@
     line-height: $breadcrumb-line-height;
     margin-left: 20px;
     position: absolute;
+    min-width: 300px;
```
makyo: Driveby?

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7785/

jujugui/static/gui/src/app/components/header-breadcrumb/_header-breadcrumb.scss
hatched: Driveby ish....I couldn't reproduce this in the original app.js but in the new one, when rendering it would be like:

Loading
Model...

Then switch to:

Loading Model...

So my guess is that the explicit rendering order is causing the container to not be full width, whereas before we were relying on some implicit rendering to make the container full width.

makyo: Aha, sounds good. +1

Lukewh: I agree with your analysis @hatched. I've come across this multiple times. As I've been going I've been making things explicit and therefore more predictable so +1 from me.

Lukewh:
Do we get sourcemaps with this?

jujugui/static/gui/src/app/init.js
```diff
@@ -0,0 +1,1590 @@
...
+    */
+    this.user = config.user || new window.jujugui.User({
+      externalAuth: config.auth,
+      expiration: window.sessionStorage.getItem('expirationDatetime')
+    });
+    /**
+      Stores the user object for the charmstore.
+      @type {Object}
+    */
+    this.users = csUser.create();
```
frankban: I think this should be called charmStoreUsers or csUsers for clarity. Also, I believe this should be completely handled by the cs mixin at this point, and treated as an internal state there, rather than exposed as an app instance attribute. The mxiin already has methods for mutating this value, adding users and removing users, so we should use those methods when required. This would also be a nice exercise for checking if mixin constructors are well handled when using this pattern in javascript.

jujugui/static/gui/src/app/init/charmstore-user.js
```diff
@@ -0,0 +1,58 @@
...
+    @param {Boolean} rerenderBreadcrumb Rerender the breadcrumb.
+   */
+  storeUser(service, rerenderProfile, rerenderBreadcrumb) {
+    const callback = (error, auth) => {
+      if (error) {
+        console.error('Unable to query user information', error);
+        return;
+      }
+      if (auth) {
+        this.users[service] = auth;
```
frankban: As mentioned, since this relies on this.users, then this.users (or better, this._csUsers) should be defined here.
Otherwise mixins can easily become a nightmare in which you jump here and there trying to find where the state comes from.

jujugui/static/gui/src/app/init/charmstore-user.js
```diff
@@ -0,0 +1,58 @@
+/* Copyright (C) 2017 Canonical Ltd. */
+'use strict';
+
+const CharmstoreUserMixin = (superclass) => class extends superclass {
```
frankban: Please add a docstring for the mixin, describing its usage and goals.

jujugui/static/gui/src/app/init/component-renderers-mixin.js
```diff
@@ -0,0 +1,1080 @@
+/* Copyright (C) 2017 Canonical Ltd. */
+'use strict';
+
+const yui = window.yui;
+
+const ComponentRenderersMixin = (superclass) => class extends superclass {
```
frankban: Please add a docstring for the mixin, describing its usage and goals.

frankban: Also, please mention the set of attributes and methods expected to be present in the provided superclass. This way we establish a contract that must be respected by the given superclass. For instance, in this case the superclass must have the db and topology attributes, etc. To be honest, I like self-sufficient mixins more, like, mixins for adding a specific functionality rather than mixins for putting some kinda related methods together in a separate file. But if the mixin relies on something existing in the provided superclass then we must absolutely document it in the factory.

frankban: A possible alternative idea is, instead of using a mixin for the renderer, defining a renderer class that gets instantiated passing the required objects (like the db, the topology etc). This class has methods for rendering stuff, like userProfile etc. This way the required attributes are explicit because they must be passed to the constructor. The JujuApp can then just do something like:
```javascript
this._render = new Renderer(this.db, this.topology...);
this._render.userProfile(...);
this._render.components();
```
What do you think?

jujugui/static/gui/src/app/init/deployer-mixin.js
```diff
@@ -0,0 +1,237 @@
+/* Copyright (C) 2017 Canonical Ltd. */
+'use strict';
+
+const DeployerMixin = (superclass) => class extends superclass {
```
frankban: Guess what I'd ask here? ;-)

jujugui/static/gui/src/app/views/utils.js
```diff
@@ -1314,23 +1316,26 @@ YUI.add('juju-view-utils', function(Y) {
     Switch model, displaying a confirmation if there are uncommitted changes.
 
     @method switchModel
-    @param {Object} env Reference to the app env.
+    @param {Object} modelAPI Reference to the app modelAPI.
```
frankban: Thank you!

jujugui/templates/index.html.mako
```diff
@@ -306,12 +306,14 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
     % if raw:
     <script src="${convoy_url}?app/assets/javascripts/version.js"></script>
     <!-- data-manual tells the Prism syntax highlighting lib to not auto-highlight -->
+    <!-- <script src="${convoy_url}?app/init-pkg.js"></script> -->
```
frankban: I know this is temporary, but I'd still add comments here to prevent deletion.

Makefile
```diff
@@ -209,6 +209,7 @@ images: $(STATIC_IMAGES) $(SVG_SPRITE_FILE) $(FAVICON)
 
 .PHONY: gui
 gui: $(JUJUGUI) $(MODULESMIN) $(BUILT_JS_ASSETS) $(BUILT_YUI) $(CSS_FILE) $(STATIC_CSS_FILES) $(STATIC_IMAGES) $(SVG_SPRITE_FILE) $(FAVICON) $(REACT_ASSETS) $(STATIC_FONT_FILES)
+	# $(NODE_MODULES)/.bin/browserify --no-builtins -r ./$(GUISRC)/app/init.js:init -o ./$(GUIBUILD)/app/init-pkg.js -t [ babelify --plugins [ transform-react-jsx ] ]
```
hatched: done

jujugui/static/gui/src/app/init.js
```diff
@@ -0,0 +1,1590 @@
...
+const csUser = require('./init/charmstore-user');
+
+const ComponentRenderersMixin = require('./init/component-renderers-mixin');
+const DeployerMixin = require('./init/deployer-mixin');
+
+const yui = window.yui;
+
+class GUIApp {
+  constructor(config) {
+    /**
```
hatched: \o/ Thx

jujugui/static/gui/src/app/init.js
```diff
@@ -0,0 +1,1590 @@
...
+      The supplied application configuration object on instantiation. Stored
+      here to avoid having to pass it around.
+      @type {Object}
+    */
+    this.applicationConfig = config;
```
hatched: I wanted something obvious for this switchover, I'll leave it for now and we can revisit later.

jujugui/static/gui/src/app/init.js
```diff
@@ -0,0 +1,1590 @@
...
+      @type {String}
+    */
+    this.defaultPageTitle = 'Juju GUI';
+    /**
+      The object used for storing a mapping of previously visited user paths
+      to the type of entity (model, store). e.g. /u/spinach/ghost would map to
+      store.
+      @type {Map}
+    */
+    this.userPaths = new Map();
```
hatched: done

jujugui/static/gui/src/app/init.js
```diff
@@ -0,0 +1,1590 @@
...
+      @type {juju.Cookies}
+    */
+    this.cookieHandler = null;
+    /**
+      The keydown event listener from the hotkey activation.
+      @type {Object}
+    */
+    this._hotkeyListener = hotkeys.activate(this);
+    /**
+      The application database
```
hatched: done

jujugui/templates/index.html.mako
hatched: done

jujugui/static/gui/src/app/init.js
```diff
@@ -0,0 +1,1590 @@
...
+      navigation is allowed, in essence when a GISF anonymous user is being
+      modeling on a new canvas.
+      @type {Boolean}
+    */
+    this.anonymousMode = false;
+    /**
+      Reference to the changesUtils utilities.
+      @type {Object}
+    */
+    this.changesUtils = window.juju.utils.ChangesUtils;
```
hatched: Definitely can and should, however, because it's still a YUI module and I can't require it, I didn't want to include it in multiple places. I'll make a note to do just that when it's de-yui'd

jujugui/static/gui/src/app/init.js
```diff
@@ -0,0 +1,1590 @@
...
+    /**
+      Timer for the drag over handler.
+      @type {Integer}
+    */
+    this._dragLeaveTimer = null;
+    /**
+      Reference to the juju.Cookies instance.
+      @type {juju.Cookies}
+    */
+    this.cookieHandler = null;
```
hatched: done

jujugui/static/gui/src/app/init.js
```diff
@@ -0,0 +1,1590 @@
...
+    /**
+      The application database
+      @type {Object}
+    */
+    this.db = new yui.juju.models.Database();
+    /**
+      Timer for debouncing database change events.
+      @type {Integer}
+    */
+    this.dbChangedTimer = null;
```
hatched: done

jujugui/static/gui/src/app/init.js
hatched: I agree on all accounts and have actually already discussed doing just that with JC. However (I'm sure saying that a lot) I feel that it's out of scope of this branch as this was just the first step to move the code over (and then some...) So I'll make a note to do this in follow-ups

jujugui/static/gui/src/app/init.js
```diff
@@ -0,0 +1,1590 @@
...
+    */
+    this.users = csUser.create();
+
+    const webHandler = new yui.juju.environments.web.WebHandler();
+    const stateGetter = () => this.state.current;
+    const cookieSetter = (value, callback) => {
+      this.charmstore.setAuthCookie(value, callback);
+    };
+    /**
+      Application instance of the bakery
```
hatched: agreed, done done done done

jujugui/static/gui/src/app/init.js
```diff
@@ -0,0 +1,1590 @@
...
+      ecs: this.ecs,
+      conn: config.conn,
+      jujuCoreVersion: config.jujuCoreVersion,
+      bundleService: this.bundleService
+    };
+    const controllerOptions = Object.assign({}, modelOptions);
+    const environments = yui.juju.environments;
+    modelOptions.webHandler = new environments.web.WebHandler();
+    /**
+      Application instance of the model API.
```
hatched: done

jujugui/static/gui/src/app/init.js
```diff
@@ -0,0 +1,1590 @@
...
+      window.dataLayer);
+
+    let baseURL = config.baseUrl;
+    if (baseURL.indexOf('://') < 0) {
+      // If there is no protocol in the baseURL then prefix the origin when
+      // creating state.
+      baseURL = `${window.location.origin}${baseURL}`;
+    }
+    /**
+      Application instance of State.
```
hatched: done

jujugui/static/gui/src/app/init.js
```diff
@@ -0,0 +1,1590 @@
...
+      charmstore: this.charmstore,
+      hideDragOverNotification: this._hideDragOverNotification.bind(this)
+    });
+
+    if (config.gisf) {
+      document.body.classList.add('u-is-beta');
+    }
+
+    /**
+      Application instance of the ACL.
```
hatched: done

jujugui/static/gui/src/app/init.js
```diff
@@ -0,0 +1,1590 @@
...
+    /**
+      Reference to the rendered topology.
+      @type {Object}
+    */
+    this.topology = this._renderTopology();
+
+    this.state.bootstrap();
+  }
+  /**
+    Creates a new instance of the charm store API. This method is idempotent.
```
hatched: I wanted to keep as much logic out of the constructor and as much in methods as possible to make it easier to stub in tests if necessary. But yes I can make those changes to it.

jujugui/static/gui/src/app/init.js
```diff
@@ -0,0 +1,1590 @@
...
+    remove('delta', handlers['onDeltaBound']);
+    remove('update', handlers['boundOnDatabaseChanged']);
+    remove('initiateDeploy', handlers['onInitiateDeploy']);
+    ['dragenter', 'dragover', 'dragleave'].forEach(eventName => {
+      remove(eventName, handlers['boundAppDragOverHandler']);
+    });
+  }
+}
+
+class JujuGUI extends mixwith.mix(GUIApp)
```
hatched: done

jujugui/static/gui/src/app/init/charmstore-user.js
hatched: I will but I fully expect this mixin to disappear and this become part of the `user` class instead as this is excessive just for a charmstore user.

jujugui/static/gui/src/app/init/charmstore-user.js
hatched: see above

jujugui/static/gui/src/app/init/charmstore-user.js
hatched: done

jujugui/static/gui/src/app/init/component-renderers-mixin.js
hatched: I've added the doc block.

I totally agree that splitting the code out like this does little to clean up those types of things however it's a little out of scope for this work. Something for a follow-up

jujugui/static/gui/src/app/init/deployer-mixin.js
hatched: :D done

hatched:
@Lukewh currently no, but we will have them eventually :)

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7798/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7799/

huwshimi:
This branch appears to run the old code correctly without issues.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7801/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7802/

jujugui/static/gui/src/app/init/component-renderers-mixin.js
frankban: I agree that this can be done in a follow up, but I am not sure my suggestion is out of scope. The mixin has been introduced in this branch, and I am suggesting an alternative approach using delegation. Given how big this change is I am just worried that these suggestions will be lost in desolate land of old github reviews ;-)

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7805/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7806/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7807/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7808/

hatched:
Thanks everyone for the reviews and qa'ing. I've created an issue https://github.com/juju/juju-gui/issues/3009 which we'll update to keep track of the remaining tasks and any issues that may come up during the transition.

hatched:
:shipit:

jujugui:

PR #2997, opened 2017-06-13
Fixes badly created logout url

jaycee:

PR #3001, opened 2017-06-14
Fix logout URL when base URL is provided.

frankban:


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7794/

hatched:
Addressed in #2997

PR #3018, opened 2017-06-21
Create added keys list and the inputs for new keys

anthonydillon:


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7816/

PR #3037, opened 2017-06-28
Fix QA errors found while testing the new ES6 class based components.

frankban:
Also update the inspect-components script so that it handles class based components.


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7833/

jaycee:
QA notes:

Using a dist uploaded into LXD with `juju upgrade-gui`:

* Loaded my active model (previously created with the CLI) correctly
* Switched between default and the new model
* Created a new model, and deploy cs:ubuntu from search results using the left hand button
* Switched to the profile
* Created another new model, deploying kibana from its details page
* Went back to the canvas from DF on the kibana model
* Actually deployed kibana
* Switched between all models
* Destroyed the ubuntu application
* Destroyed the ubuntu model
* Destroyed the kibana model with still active kibana application

I saw no errors.

QA OK for GIJoe.

frankban:
Thank you both for the reviews!
:shipit:

jujugui:

PR #3038, opened 2017-06-30
Decouple components so they can be shared between profiles

huwshimi:


huwshimi:
QA:
- load your profile and another user's profile
- the profiles should appear as before

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7834/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7835/

jujugui/static/gui/src/app/components/user-profile/model-list/model-list.js
```diff
@@ -83,10 +86,8 @@ class UserProfileModelList extends React.Component {
           return model.owner === extUser;
         });
       }
-      if (modelList.length) {
-        broadcastStatus('ok');
-      } else {
-        broadcastStatus('empty');
+      if (modelList && modelList.length > 0) {
+        this.props.setHasEntities(true);
```
frankban: Do we ever set this to false? I mean, even if we do, I am not sure about one single component setting has entities to falso for all the others. So maybe simplify this call like `this.props.notifyHasEntities()` ?

jujugui/static/gui/src/app/components/user-profile/user-profile.js
```diff
@@ -28,6 +35,15 @@ class UserProfile extends React.Component {
   }
 
   /**
+    Set whether the profile has entities.
```
frankban: Please consider the change above, and also describe better how and where this is set.

jujugui/static/gui/src/app/components/user-profile/user-profile.js
```diff
@@ -53,20 +69,31 @@ class UserProfile extends React.Component {
   }
 
   /**
+    Generate the empty state.
+
+    @method _generateContent
```
frankban: This is wrong and you can just remove this line.

frankban:
QA is suboptimal. On a not amazingly fast browser, every time you click to go to a profile page the empty profile is displayed, and it disappears when the calls to get charms, bundle etc return and there are contents. But initially it's just confusing. So I'd suggest that the components collaborate to sigal that the profile is ready, and only at that point we can decide whether to display the empty profile.

hatched:
After discussing with @frankban we've determined that we will still need a scaleable watcher even in the new profile view for the summary page. And for situations where we are rendering these components without the watcher, they can simply be updated to call a noop broadcast method if none is provided.

huwshimi:
As discussed with @hatched we'll need to do this work eventually so I've fixed the loading issue. The sections now show the spinner and if all sections are empty then the empty profile state is displayed.

huwshimi:
Continued in #3054.

PR #3039, opened 2017-06-30
Fetch additional model info: agent version, SLA and status info.

frankban:


frankban:
QA:
It must be done in gijoe and gisf scenarios (for gisf, given the changes in this branch, guiproxy should be sufficient).

- open the Javascript console and run the following:
  `b = function() {console.log(arguments)}; app.controllerAPI.listModelsWithInfo(cb);`;
- inspect the results and check that the agent version, SLA and status informations are properly reported.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7836/

jujugui/static/gui/src/app/store/env/controller-api.js
```diff
@@ -463,7 +463,6 @@ YUI.add('juju-controller-api', function(Y) {
       Return information about Juju models, such as their names, series, and
       provider types, by performing a ModelManager.ModelInfo Juju API request.
 
-      @method modelInfo
```
jaycee: I don't think it matters, but why are we removing the '@method' comment statement?

frankban: It's not required anymore, and that's a good thing, as the name of the method/function is obvious without the annotation, and removing it reduces noise and means one less thing that can go out of sync (we have many examples of methods changing name but not the docstring). So I decided to remove them when I encounter them.

jaycee:
QA

GIjoe:

Available models were 2.0.0, which do not provide the necessary information, however the GUI properly degraded without errors.

I was unable to QA gisf b/c of errors in guiproxy.

frankban:
Thanks for the reviews!
:shipit:

jujugui:

PR #3051, opened 2017-07-05
Initial work on the new Profile UI

hatched:
Adds a new Profile, Navigation, and Header components skeleton for the new profile. It is hooked up to be functional with the state system.

It's behind the 'profile' feature flag.

### DIFF
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -1045,7 +1045,7 @@ YUI.add('juju-gui', function(Y) {
         profile =
           <window.juju.components.Profile
             changeState={this.bounded.changeState}
-            activeSection={state.hash}/>;
+            activeSection={decodeURIComponent(state.hash)}/>;
       }
 
       ReactDOM.render(profile, document.getElementById('top-page-container'));
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -558,6 +558,13 @@ YUI.add('juju-gui', function(Y) {
         document.addEventListener(
           eventName, this._boundAppDragOverHandler);
       });
+      // As a minor performance boost and to avoid potential rerenderings
+      // because of rebinding functions in the render methods. Any method that
+      // requires binding and is passed into components should be bound here
+      // and then used across components.
+      this.bounded = {
+        changeState: this.state.changeState.bind(this.state)
+      };
       // In Juju >= 2 we connect to the controller and then to the model.
       this.state.bootstrap();
     },
@@ -1035,7 +1042,10 @@ YUI.add('juju-gui', function(Y) {
       />;
 
       if (window.juju_config.flags.profile) {
-        profile = <window.juju.components.Profile />;
+        profile =
+          <window.juju.components.Profile
+            changeState={this.bounded.changeState}
+            activeSection={state.hash}/>;
       }
 
       ReactDOM.render(profile, document.getElementById('top-page-container'));
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -562,7 +562,7 @@ YUI.add('juju-gui', function(Y) {
       // because of rebinding functions in the render methods. Any method that
       // requires binding and is passed into components should be bound here
       // and then used across components.
-      this.bounded = {
+      this._bound = {
         changeState: this.state.changeState.bind(this.state)
       };
       // In Juju >= 2 we connect to the controller and then to the model.
@@ -1044,8 +1044,8 @@ YUI.add('juju-gui', function(Y) {
       if (window.juju_config.flags.profile) {
         profile =
           <window.juju.components.Profile
-            changeState={this.bounded.changeState}
-            activeSection={decodeURIComponent(state.hash)}/>;
+            changeState={this._bound.changeState}
+            activeSection={state.hash}/>;
       }
 
       ReactDOM.render(profile, document.getElementById('top-page-container'));
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -1008,33 +1008,37 @@ YUI.add('juju-gui', function(Y) {
       // NOTE: we need to clone this.get('users') below; passing in without
       // cloning breaks React's ability to distinguish between this.props and
       // nextProps on the lifecycle methods.
-      ReactDOM.render(
-        <window.juju.components.UserProfile
-          acl={this.acl}
-          addNotification=
-            {this.db.notifications.add.bind(this.db.notifications)}
-          charmstore={charmstore}
-          currentModel={currentModel}
-          d3={d3}
-          facadesExist={facadesExist}
-          listBudgets={this.plans.listBudgets.bind(this.plans)}
-          listModelsWithInfo={
-            this.controllerAPI.listModelsWithInfo.bind(this.controllerAPI)}
-          getKpiMetrics={this.plans.getKpiMetrics.bind(this.plans)}
-          changeState={this.state.changeState.bind(this.state)}
-          destroyModels={
-            this.controllerAPI.destroyModels.bind(this.controllerAPI)}
-          getAgreements={this.terms.getAgreements.bind(this.terms)}
-          getDiagramURL={charmstore.getDiagramURL.bind(charmstore)}
-          interactiveLogin={this.get('interactiveLogin')}
-          pluralize={utils.pluralize.bind(this)}
-          setPageTitle={this.setPageTitle}
-          staticURL={window.juju_config.staticURL}
-          storeUser={this.storeUser.bind(this)}
-          switchModel={utils.switchModel.bind(this, this.env)}
-          userInfo={this._getUserInfo(state)}
-        />,
-        document.getElementById('top-page-container'));
+      let profile = <window.juju.components.UserProfile
+        acl={this.acl}
+        addNotification=
+          {this.db.notifications.add.bind(this.db.notifications)}
+        charmstore={charmstore}
+        currentModel={currentModel}
+        d3={d3}
+        facadesExist={facadesExist}
+        listBudgets={this.plans.listBudgets.bind(this.plans)}
+        listModelsWithInfo={
+          this.controllerAPI.listModelsWithInfo.bind(this.controllerAPI)}
+        getKpiMetrics={this.plans.getKpiMetrics.bind(this.plans)}
+        changeState={this.state.changeState.bind(this.state)}
+        destroyModels={
+          this.controllerAPI.destroyModels.bind(this.controllerAPI)}
+        getAgreements={this.terms.getAgreements.bind(this.terms)}
+        getDiagramURL={charmstore.getDiagramURL.bind(charmstore)}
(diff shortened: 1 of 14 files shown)


hatched:
#### To QA

As this is very much a work in progress a QA isn't really necessary but if you'd like to see it functional you can view it by loading the GUI with the 'profile' feature flag and visiting your profile.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7843/

jujugui/static/gui/src/app/components/profile/header/header.js
```diff
@@ -0,0 +1,22 @@
+/* Copyright (C) 2017 Canonical Ltd. */
+
+'use strict';
+
+class ProfileHeader extends React.Component {
+
+  render() {
+    return (
+      <div className="profile-header">
+        <div className="content"></div>
```
anthonydillon: It might be better to BEM your class names here. Something like `profile-header__content`. Reduces the risk of styles leaking.

Then your Sass should be something like this:
``` scss
.profile-header {
  ...

  &__content {
    ...
  }
}
```

jujugui/static/gui/src/app/components/profile/navigation/navigation.js
```diff
@@ -0,0 +1,62 @@
...
+      // case on top of model and entity names.
+      hash: encodeURIComponent(label.toLowerCase())
+    });
+  }
+
+  render() {
+    const links = ['Models', 'Bundles', 'Charms', 'Cloud Credentials']
+      .map(label =>
+        <li
+          className={this.state.activeSection === label.toLowerCase() ? 'active' : ''} // eslint-disable-line max-len
```
anthonydillon: I think `is-active` is a better state classname.

jujugui/static/gui/src/app/components/profile/navigation/_navigation.scss
```diff
@@ -0,0 +1,12 @@
+.profile-navigation {
+
+  li {
```
anthonydillon: You should target class names in components. So the underlying markup can change or be different in other places of the app without breaking the styling.

For example: `&__list-item`.

jujugui/static/gui/src/app/components/profile/navigation/navigation.js
```diff
@@ -0,0 +1,62 @@
...
+      // of the ambiguous paths can not easily support yet another collision
+      // case on top of model and entity names.
+      hash: encodeURIComponent(label.toLowerCase())
+    });
+  }
+
+  render() {
+    const links = ['Models', 'Bundles', 'Charms', 'Cloud Credentials']
+      .map(label =>
+        <li
```
anthonydillon: Please add `role="button"`

jujugui/static/gui/src/app/components/profile/header/_header.scss
```diff
@@ -0,0 +1,12 @@
+.profile-header {
+
```
anthonydillon: No need for whitespace here.

jujugui/static/gui/src/app/components/profile/navigation/navigation.js
```diff
@@ -0,0 +1,62 @@
...
+          className={this.state.activeSection === label.toLowerCase() ? 'active' : ''} // eslint-disable-line max-len
+          key={label}
```
frankban: I am not sure I like using labels as keys/hashes. Why not building a map? Like
```javascript
const sections = new Map();
sections.set('models', 'Models');
sections.set('charms', 'Charms');
sections.set('bundles', 'Bundles');
sections.set('credentials', 'Cloud Credentials');

const activeSection = this.state.activeSection;
const links = [];
sections.forEach((value, key) => {
  links.push(<li 
    className={key === activeSection ? 'active' : ''}
    key={key}
    onClick={this._changeState.bind(this, key)}>
    {value}
  </li>);
});
```
This way the order is maintained, the map can be expanded later if we need more state, and we can also avoid the arbitrary toLowerCase modification in _changeState.
What do you think?

jujugui/static/gui/src/app/components/profile/navigation/test-navigation.js
```diff
@@ -0,0 +1,109 @@
...
+describe('Profile Navigation', function() {
+
+  beforeAll(function(done) {
+    // By loading this file it adds the component to the juju components.
+    YUI().use('profile-navigation', function() { done(); });
+  });
+
+  it('can render', () => {
+    const changeState = sinon.stub();
+    const output = jsTestUtils.shallowRender(
```
frankban: I'd start putting rendering code in its own method here so that all tests can use that without too much repetition. We have too many cases of testing code never abstracted out, so maybe by doing that earlier even when it's not really required could help.

jujugui/static/gui/src/app/components/profile/navigation/navigation.js
```diff
@@ -0,0 +1,62 @@
+/* Copyright (C) 2017 Canonical Ltd. */
+
+'use strict';
+
+class ProfileNavigation extends React.Component {
```
frankban: Docstring for the component, how is it used, for what purpose, what does it provide?

jujugui/static/gui/src/app/components/profile/header/header.js
```diff
@@ -0,0 +1,22 @@
+/* Copyright (C) 2017 Canonical Ltd. */
+
+'use strict';
+
+class ProfileHeader extends React.Component {
```
frankban: Docstring for the component, how is it used, for what purpose, what does it provide?

jujugui/static/gui/src/app/components/profile/profile.js
```diff
@@ -0,0 +1,80 @@
+/* Copyright (C) 2017 Canonical Ltd. */
+
+'use strict';
+
+class Profile extends React.Component {
```
frankban: Try to guess?

jujugui/static/gui/src/app/components/profile/navigation/navigation.js
```diff
@@ -0,0 +1,62 @@
...
+      activeSection: this.props.activeSection || 'models'
+    };
+  }
+
+  componentWillReceiveProps(nextProps) {
+    this.setState({activeSection: nextProps.activeSection});
+  }
+
+  shouldComponentUpdate(nextProps, nextState) {
+    if (this.state.activeSection !== nextProps.activeSection) {
```
frankban: This function could be rewritten as just `return this.state.activeSection !== nextProps.activeSection;`

jujugui/static/gui/src/app/components/profile/profile.js
```diff
@@ -0,0 +1,80 @@
...
+    const activeSection = this.props.activeSection || 'models';
+    this.state = {activeSection};
+  }
+
+  componentWillReceiveProps(nextProps) {
+    this.setState({activeSection: nextProps.activeSection});
+  }
+
+  shouldComponentUpdate(nextProps, nextState) {
+    if (nextProps.activeSection !== this.state.activeSection) {
```
frankban: This function could be rewritten as just `return this.state.activeSection !== nextProps.activeSection;`

jujugui/static/gui/src/app/components/profile/profile.js
```diff
@@ -0,0 +1,80 @@
...
+      return true;
+    }
+    return false;
+  }
+
+  _generateContent() {
```
frankban: I appears that we can reuse the map I suggested above again and again, and avoid all the repetition we have currently, even when calculating the default section if not provided above (shuld be the first entry in the map, rather than 'models'). So maybe define the map here and then propagate it to Navigation? I'd like to avoid having to write the same logic multiple time when adding a section. I think I also like the navigation to be as agnostic as possible about the number and labels of sections.

jujugui/static/gui/src/app/app.js
```diff
@@ -558,6 +558,13 @@ YUI.add('juju-gui', function(Y) {
         document.addEventListener(
           eventName, this._boundAppDragOverHandler);
       });
+      // As a minor performance boost and to avoid potential rerenderings
+      // because of rebinding functions in the render methods. Any method that
+      // requires binding and is passed into components should be bound here
+      // and then used across components.
+      this.bounded = {
```
hatched: changed to `_bound`

jujugui/static/gui/src/app/components/profile/header/_header.scss
hatched: removed

jujugui/static/gui/src/app/components/profile/header/header.js
hatched: agreed, updated.

jujugui/static/gui/src/app/components/profile/header/header.js
hatched: done

jujugui/static/gui/src/app/components/profile/navigation/navigation.js
hatched: done

jujugui/static/gui/src/app/components/profile/navigation/navigation.js
hatched: Yeah good call, I had originally planned on expanding this, but I can do that at a later date if necessary

jujugui/static/gui/src/app/components/profile/navigation/_navigation.scss
hatched: done

jujugui/static/gui/src/app/components/profile/navigation/navigation.js
hatched: done

jujugui/static/gui/src/app/components/profile/navigation/navigation.js
hatched: done

jujugui/static/gui/src/app/components/profile/navigation/navigation.js
hatched: I've moved the `Map` into the `Profile` component and passed it in.

jujugui/static/gui/src/app/components/profile/profile.js
hatched: hah, done

jujugui/static/gui/src/app/components/profile/navigation/test-navigation.js
hatched: done

jujugui/static/gui/src/app/components/profile/profile.js
hatched: done

jujugui/static/gui/src/app/components/profile/profile.js
hatched: Done, the profile page now defines a static property which is the sections that gets used in the tests and is passed through the components.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7851/

hatched:
Thanks all! :shipit:

jujugui:

PR #3061, opened 2017-07-07
Create new model-list component 

hatched:
This creates a new model-list component that's more appropriate for the new profile designs. This simply renders the two tables of models, My Models, and Shared With Me. Other features will be added in follow-up branches.

### DIFF
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -563,7 +563,11 @@ YUI.add('juju-gui', function(Y) {
       // requires binding and is passed into components should be bound here
       // and then used across components.
       this._bound = {
-        changeState: this.state.changeState.bind(this.state)
+        changeState: this.state.changeState.bind(this.state),
+        addNotification: this.db.notifications.add.bind(this.db.notifications),
+        listModelsWithInfo: this.controllerAPI.listModelsWithInfo.bind(this.controllerAPI), // eslint-disable-line max-len
+        destroyModels: this.controllerAPI.destroyModels.bind(this.controllerAPI), // eslint-disable-line max-len
+        switchModel: views.utils.switchModel.bind(this, this.env)
       };
       // In Juju >= 2 we connect to the controller and then to the model.
       this.state.bootstrap();
@@ -1032,8 +1036,15 @@ YUI.add('juju-gui', function(Y) {
       if (window.juju_config.flags.profile) {
         profile =
           <window.juju.components.Profile
+            acl={this.acl}
+            activeSection={state.hash}
+            addNotification={this._bound.addNotification}
             changeState={this._bound.changeState}
-            activeSection={state.hash}/>;
+            facadesExist={facadesExist}
+            listModelsWithInfo={this._bound.listModelsWithInfo}
+            destroyModels={this._bound.destroyModels}
+            switchModel={this._bound.switchModel}
+            userInfo={this._getUserInfo(state)} />;
       }
 
       ReactDOM.render(profile, document.getElementById('top-page-container'));
--- a/jujugui/static/gui/src/app/app.js
+++ b/jujugui/static/gui/src/app/app.js
@@ -563,10 +563,10 @@ YUI.add('juju-gui', function(Y) {
       // requires binding and is passed into components should be bound here
       // and then used across components.
       this._bound = {
-        changeState: this.state.changeState.bind(this.state),
         addNotification: this.db.notifications.add.bind(this.db.notifications),
-        listModelsWithInfo: this.controllerAPI.listModelsWithInfo.bind(this.controllerAPI), // eslint-disable-line max-len
+        changeState: this.state.changeState.bind(this.state),
         destroyModels: this.controllerAPI.destroyModels.bind(this.controllerAPI), // eslint-disable-line max-len
+        listModelsWithInfo: this.controllerAPI.listModelsWithInfo.bind(this.controllerAPI), // eslint-disable-line max-len
         switchModel: views.utils.switchModel.bind(this, this.env)
       };
       // In Juju >= 2 we connect to the controller and then to the model.
--- a/jujugui/static/gui/src/app/components/profile/model-list/model-list.js
+++ b/jujugui/static/gui/src/app/components/profile/model-list/model-list.js
@@ -44,7 +44,7 @@ class ProfileModelList extends React.Component {
   }
 
   /**
-    Callback for the controller listModels call.
+    Callback for the controller listModelsWithInfo call.
     @param {String} err The error from the request, or null.
     @param {Array} modelList The list of models.
   */
--- a/jujugui/static/gui/src/app/components/profile/model-list/model-list.js
+++ b/jujugui/static/gui/src/app/components/profile/model-list/model-list.js
@@ -83,13 +83,14 @@ class ProfileModelList extends React.Component {
   */
   _generateMyModels() {
     const key = 'mymodels';
+    const headerLabel = 'My models';
     const labels = [
       'Name', 'Machines, Cloud/region', 'last accessed', 'action'];
-    const header = this._generateHeader(labels, key);
+    const tableHeader = this._generateTableHeader(labels, key);
     const state = this.state.models;
     const modelList = state && state.owned;
     if (!modelList || modelList.length === 0) {
-      return header;
+      return [this._generateHeader(headerLabel, 0, true), tableHeader];
     }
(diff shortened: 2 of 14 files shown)


hatched:
#### To QA

- Use the `profile` feature flag
- Visit your profile and you should see two tables, one with your own models and one with models shared with you

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7862/

jujugui/static/gui/src/app/app.js
```diff
@@ -563,7 +563,11 @@ YUI.add('juju-gui', function(Y) {
       // requires binding and is passed into components should be bound here
       // and then used across components.
       this._bound = {
-        changeState: this.state.changeState.bind(this.state)
+        changeState: this.state.changeState.bind(this.state),
```
frankban: Keep these in alphabetical order?

jujugui/static/gui/src/app/components/profile/model-list/model-list.js
```diff
@@ -0,0 +1,256 @@
...
+  constructor() {
+    super();
+    this.state = {
+      loadingModels: false,
+      models: null
+    };
+  }
+
+  componentWillMount() {
+    this._fetchModels(this.props.facadesExist);
```
frankban: Should facadesExist be called controllerConnected or just apiConnected? I guess this component is interested in whether the API is connected, not caring too much about the facade concept.

jujugui/static/gui/src/app/components/profile/model-list/model-list.js
```diff
@@ -0,0 +1,256 @@
...
+    }
+    // Delay the call until after the state change to prevent race
+    // conditions.
+    this.setState({loadingModels: true}, () => {
+      this.props.listModelsWithInfo(this._fetchModelsCallback.bind(this));
+    });
+  }
+
+  /**
+    Callback for the controller listModels call.
```
frankban: listModelsWithInfo

jujugui/static/gui/src/app/components/profile/model-list/model-list.js
```diff
@@ -0,0 +1,256 @@
...
+        return;
+      }
+      // Split the models into ones owned by the active user and ones
+      // shared to the active user.
+      const models = {
+        owned: [],
+        shared: []
+      };
+      modelList.forEach(model => {
+        if (model.owner === `${this.props.userInfo.profile}@external`) {
```
frankban: Does this work in gijoe and local users?
I guess that rather than just adding `@external` at the end we should be more precise about the domain fo the currently logged in user.

jujugui/static/gui/src/app/components/profile/model-list/model-list.js
```diff
@@ -0,0 +1,256 @@
...
+    const headerLabel = 'My models';
+    const labels = [
+      'Name', 'Machines, Cloud/region', 'Last accessed', 'Action'];
+    const tableHeader = this._generateTableHeader(labels, key);
+    const state = this.state.models;
+    const modelList = state && state.owned;
+    if (!modelList || modelList.length === 0) {
+      return [this._generateHeader(headerLabel, 0, true), tableHeader];
+    }
+    const rowData = modelList.reduce((models, model) => {
```
frankban: Currently models are sorted by last connection date, most recent first. I think we should keep that sorting.

jujugui/static/gui/src/app/components/profile/model-list/model-list.js
```diff
@@ -0,0 +1,256 @@
...
+      }
+      models.push([
+        model.name,
+        `${model.numMachines} ${model.provider.toUpperCase()}/${model.region.toUpperCase()}`, // eslint-disable-line max-len
+        model.lastConnection,
+        '-'
+      ]);
+      return models;
+    }, []);
+    const rows = this._generateRows(rowData, key);
```
frankban: What happens when there are no rows (no models) in both here and the shared models case?
Those two cases also need to be tested below.

jujugui/static/gui/src/app/components/profile/model-list/test-model-list.js
```diff
@@ -0,0 +1,231 @@
...
+  }]`;
+
+  beforeAll(function(done) {
+    // By loading this file it adds the component to the juju components.
+    YUI().use('profile-model-list', function() {
+      done();
+    });
+  });
+
+  function renderComponent(options={}) {
```
frankban: Maybe add a modelResponse parameter, defaulting to rawModelData?
This way we can easily test the cases in which we have no owned models, no shared models, and no models at all.

jujugui/static/gui/src/app/components/profile/profile.js
```diff
@@ -43,25 +39,51 @@ class Profile extends React.Component {
...
   ['models', {
     label: 'Models',
-    component: 'Models'
+    getComponent: context => {
+      return (
+        <juju.components.ProfileModelList
+          acl={context.props.acl}
+          addNotification={context.props.addNotification}
+          changeState={context.props.changeState}
+          facadesExist={context.props.facadesExist}
```
frankban: As mentioned, apiConnected would be less confusing.

jujugui/static/gui/src/app/components/profile/profile.js
```diff
@@ -43,25 +39,51 @@ class Profile extends React.Component {
...
+          destroyModels={context.props.destroyModels}
+          listModelsWithInfo={context.props.listModelsWithInfo}
+          switchModel={context.props.switchModel}
+          userInfo={context.props.userInfo} />);
+    }
   }],
   ['charms', {
     label: 'Charms',
-    component: 'Charms'
+    getComponent: context => 'Charms'
```
frankban: Rather than context (very generic/ it could be anuthing) isn't this the actual component instance? So why not comp, or compInstance?

jujugui/static/gui/src/app/init.js
```diff
@@ -319,7 +319,11 @@ class GUIApp {
       @type {Object}
     */
     this._bound = {
-      changeState: this.state.changeState.bind(this.state)
+      changeState: this.state.changeState.bind(this.state),
```
frankban: Alphabetical order.

jujugui/static/gui/src/app/components/profile/model-list/model-list.js
hatched: Not at the moment, none of this works for gijoe or users which aren't necessarily your logged in user. That will be coming in a follow-up

jujugui/static/gui/src/app/app.js

jujugui/static/gui/src/app/components/profile/model-list/model-list.js
hatched: done

jujugui/static/gui/src/app/components/profile/model-list/model-list.js
hatched: Yep that order is kept.

jujugui/static/gui/src/app/components/profile/model-list/model-list.js
hatched: Right now there is just a header with no content - this will be updated and tested for once we have a better UI for that eventuality.

jujugui/static/gui/src/app/components/profile/model-list/model-list.js
hatched: In this case we not only need to know that the controller is connected and logged in, but also that the facades have been returned from Juju. If we don't have that then the RPC calls will fail as 'not implemented' because it won't have matching facades.

jujugui/static/gui/src/app/components/profile/model-list/test-model-list.js
hatched: Sounds like a good idea

jujugui/static/gui/src/app/components/profile/profile.js
hatched: updated

jujugui/static/gui/src/app/init.js
hatched: done

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7864/

hatched:
I'm marking it as QA OK because I'll add it in a follow-up PR :)

THANKS!

:shipit:

jujugui:

PR #3065, opened 2017-07-10
Add tag support to statsd client

jaycee:


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7866/

jaycee:
QA:

* Start up the gui w/ guiproxy pointing at prod (make sure to update guiproxy to use the -env option); also pass `-config 'flags: {"test.foo": true, "bar": true}'`
* In the gui, open the debugger; insert a breakpoint in the _send function in the statsd.js file.
* In the console, run:

```
const stats = new window.jujugui.StatsClient('http://example.com');
stats.increase('foo');
```

The breakpoint should trigger. Inspecting data should show `foo,test.foo=true:1|c`.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7868/

jujugui/static/gui/src/app/utils/statsd.js
```diff
@@ -29,10 +29,29 @@ const StatsClient = class StatsClient {
...
   increase(name, count=1) {
+    name = this._addFlags(name);
     this._send(`${name}:${count}|c`);
   }
 
   /**
+   Adds the active ab test flags to the stat name.
+
+   @param {Array} flags The flags to check for test tags. Used for testing.
+   @param {String} name The stats name.
```
hatched: Missing return docstring

jujugui/static/gui/src/app/utils/statsd.js
```diff
@@ -29,10 +29,29 @@ const StatsClient = class StatsClient {
...
+  */
+  _addFlags(name, flags=[]) {
+    if (flags.length === 0 && window.juju_config && window.juju_config.flags) {
```
hatched: the config flags is now always defined so you can just check `juju_config.flags` https://github.com/juju/juju-gui/blob/develop/jujugui/templates/index.html.mako#L350

jujugui/static/gui/src/app/utils/statsd.js
```diff
@@ -29,10 +29,29 @@ const StatsClient = class StatsClient {
...
+      flags = Object.keys(window.juju_config.flags);
+    }
+    flags = flags.filter(key => key.indexOf('test') === 0);
+    flags = flags.map(val => `${val}=true`).join(',');
+    if (flags !== '') {
```
hatched: `flags` will be an array at this point, so I don't think this can ever be true

jujugui/static/gui/src/app/utils/statsd.js
```diff
@@ -29,10 +29,29 @@ const StatsClient = class StatsClient {
...
+   Adds the active ab test flags to the stat name.
+
+   @param {Array} flags The flags to check for test tags. Used for testing.
+   @param {String} name The stats name.
+  */
+  _addFlags(name, flags=[]) {
+    if (flags.length === 0 && window.juju_config && window.juju_config.flags) {
+      flags = Object.keys(window.juju_config.flags);
+    }
+    flags = flags.filter(key => key.indexOf('test') === 0);
```
hatched:
```JavaScript
flags = flags.filter(key => key.indexOf('test') === 0)
             .map(val => `${val}=true`).join(',');
```

or you can use `reduce` so it only takes one loop (not really a problem since there will probably be....2 items :D

(untested)
```JavaScript
flags = flags.reduce((flags, flag) => {
  if (key.indexOf('test') !== 0) {
    return;
  }
  flags.push(flag)
  return flags;
});
```

jujugui/static/gui/src/app/utils/statsd.js
hatched: There also isn't a test for this conditional

jujugui/static/gui/src/app/utils/statsd.js
jaycee: Look above, flags is a string as a result of that last 'join' call.

jaycee: And if the array that join was called on was [], it'll be ''.

hatched: Woops, see that's what happens when you rush :D

jujugui/static/gui/src/app/utils/statsd.js
jaycee: Per discussion in IRC we're now passing flags into the constructor so there no longer is a conditional.

jujugui/static/gui/src/app/utils/statsd.js
jaycee: I took the non reduce option; as you said, minimizing loops isn't a big deal here and the other option is more clearly understood when reading through.

jujugui/static/gui/src/app/utils/statsd.js
jaycee: Added.

jaycee:
:shipit:

jujugui:

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7869/

PR #3066, opened 2017-07-11
Rely on the newly added support for tags when sending stats.

frankban:
Also slightly improve support for multiple tags.

### DIFF
--- a/jujugui/static/gui/src/app/utils/statsd.js
+++ b/jujugui/static/gui/src/app/utils/statsd.js
@@ -19,7 +19,7 @@ const StatsClient = class StatsClient {
   */
   constructor(url, prefix='', flags={}) {
     this.url = url.replace(/\/?$/, '/');
-    this.flags = flags;
+    this._flags = flags || {};
     this._prefix = prefix;
   }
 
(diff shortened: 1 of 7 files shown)


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7870/

jujugui/static/gui/src/app/utils/statsd.js
```diff
@@ -19,24 +19,25 @@ const StatsClient = class StatsClient {
   */
   constructor(url, prefix='', flags={}) {
     this.url = url.replace(/\/?$/, '/');
-    this.flags = flags;
+    this._flags = flags || {};
```
jaycee: So flags default to {} in the declaration, and according to hatch juju_config.flags is always defined. I don't think we need the `|| {}` bit. Not that there's any harm in having it.

jujugui/static/gui/src/app/utils/test-statsd.js
```diff
@@ -71,11 +71,24 @@ describe('StatsClient', () => {
...
     const name = client._addFlags('foo');
     assert.equal(name, 'foo,test.bar=true');
   });
+
+  it('adds multiple flags as statsd tags', () => {
+    const url = 'https://example.com/stats/';
+    const flags = {
+      'test.bar': true,
+      baz: true,
+      'test.bad': false,
```
jaycee: We never actually pass along inactive flags; these are always going to be true. We only put a value in so that we can create flags as an object when we're setting up the gui config on the blues browser side.

jujugui/static/gui/src/app/utils/statsd.js
frankban: Yes indeed, but now flags are provided as an argument, and some sanity check could maybe help?

jujugui/static/gui/src/app/utils/test-statsd.js
frankban: Yes you are right, but given we established the contract that flags are a map, and given  that checking the value truthiness is very easy, I added the check so that we don't rely on a convention established by the client.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7871/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7872/

frankban:
Thanks for the review and for fixing lint on Jenkins!
:shipit:

jujugui:

Issue #3073, opened 2017-07-11
"Get Support" Link not shown in GUI in Juju


hatched:
I don't see why we'd only want to show the link to the documentation only in JAAS. This restriction should be removed.

PR #3076, opened 2017-07-13
Always show the support link

huwshimi:
Fixes: #3073. Fixes: #3072. Fixes: #3078.

### DIFF
--- a/jujugui/static/gui/src/app/components/header-help/header-help.js
+++ b/jujugui/static/gui/src/app/components/header-help/header-help.js
@@ -60,7 +60,8 @@ class HeaderHelp extends React.Component {
     let link = 'https://github.com/juju/juju-gui/issues';
     if (this.props.user) {
       label = 'Get Support';
-      link = 'https://jujucharms.com/support';
+      link = this.props.gisf ? 'https://jujucharms.com/support' :
+        'https://jujucharms.com/docs/stable/about-juju';
     }
     return (
       <li className="header-menu__menu-list-item
@@ -84,7 +85,6 @@ class HeaderHelp extends React.Component {
           View Documentation</a>
       </li>);
     }
-    return;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


huwshimi:
QA:
- in gisf and gij click the help icon in the header
- the "Get support" link should be visible
- hover the applications/machines buttons
- the cursor should be the hand pointer thingy

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7882/

jujugui/static/gui/src/app/components/header-help/header-help.js
```diff
@@ -58,7 +58,7 @@ class HeaderHelp extends React.Component {
   _generateIssuesLink() {
     let label = 'File Issue';
     let link = 'https://github.com/juju/juju-gui/issues';
-    if (this.props.gisf && this.props.user) {
+    if (this.props.user) {
```
frankban: I am not sure about pointing gijoe users to the JAAS support page. I'd like them not to assume that we provide support for their own controllers.

hatched: @frankban what if we created another landing page that was identical but with some of the wordings/links changed? But in the mean time updating the link to point to https://jujucharms.com/docs/stable/about-juju in the gisf case?

frankban: Yes, let's add a link to that page in the gijoe case.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7906/

jujugui/static/gui/src/app/components/header-help/header-help.js
huwshimi: I've added the alternate link for gijoe.

jujugui/static/gui/src/app/components/header-help/header-help.js
```diff
@@ -58,9 +58,10 @@ class HeaderHelp extends React.Component {
...
       label = 'Get Support';
```
hatched: I'm not sure that this label makes sense in the gijoe case. Maybe "Documentation" is better in that case?

hatched:
Thanks @huwshimi 

Everything works great, just one small request about the label then this is good to land!

huwshimi:
@hatched there is a separate "View documentation" link, so instead of having a different gijoe link for support it should be a different link for documentation?

hatched:
:shipit:

jujugui:

Issue #3080, opened 2017-07-13
Visiting dd while logged in still shows as logged out


hatched:
Log into jujucharms then close the tab, visit a url which direct deploys a bundle. It'll now show you as logged out. Clicking the login, logs you in immediately.

This is likely a race condition because by the time the df is rendered the login hasn't yet completed.

PR #3081, opened 2017-07-14
Fix deployment flow authentication state updates

huwshimi:
Fixes: #3080.

### DIFF
--- a/jujugui/static/gui/src/app/components/deployment-flow/deployment-flow.js
+++ b/jujugui/static/gui/src/app/components/deployment-flow/deployment-flow.js
@@ -33,7 +33,6 @@ class DeploymentFlow extends React.Component {
       deploying: false,
       credential: this.props.credential,
       loadingTerms: false,
-      loggedIn: this.props.isLoggedIn(),
       modelName: this.props.modelName,
       newTerms: [],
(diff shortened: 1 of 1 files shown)


jujugui/static/gui/src/app/components/deployment-flow/deployment-flow.js
```diff
@@ -33,7 +33,6 @@ class DeploymentFlow extends React.Component {
       deploying: false,
       credential: this.props.credential,
       loadingTerms: false,
-      loggedIn: this.props.isLoggedIn(),
```
huwshimi: If the logged in state is set here when the component is created then the `isLoggedIn()` method never gets called again to update the state.

huwshimi:
QA:
- load the direct deploy flow when logged in (e.g. /new/?dd=cs:bundle/mediawiki-single-9)
- you should see the logged in version of the direct deploy
- log out and load the direct deploy, you should the logged out version
- click the login button and then you should get the logged in version

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7891/

frankban:
QA is ok.
:shipit:

jujugui:

PR #3097, opened 2017-07-25
Gracefully fail when charmstore requests return no data

### LINKED ISSUE #3086: model/entity disambiguation needs to reject on timeout.
If the GUI does not have access to the charmstore the disambiguation call to the charmstore times out. This appears to be causing it to execute the success path of `_fetchEntityFromUserState` https://github.com/juju/juju-gui/blob/develop/jujugui/static/gui/src/app/app.js#L1997

Related: https://github.com/juju/juju-gui/issues/3083

hatched:
Fixes #3086

If the API is functioning properly this only happens when the host is blocked by a VPN. This also allows the GUI to display model information even if the user cannot access the charmstore.

### DIFF
--- a/jujugui/static/gui/src/app/components/entity-details/entity-details.js
+++ b/jujugui/static/gui/src/app/components/entity-details/entity-details.js
@@ -150,7 +150,7 @@ class EntityDetails extends React.Component {
     @param {Array} models A list of the entity models found.
   */
   _fetchCallback(error, data) {
-    if (error) {
+    if (error || !data) {
       this._changeActiveComponent('error');
       console.error('cannot fetch the entity:' + error);
       return;
--- a/jujugui/static/gui/src/app/jujulib/charmstore.js
+++ b/jujugui/static/gui/src/app/jujulib/charmstore.js
@@ -83,6 +83,10 @@ var module = module;
     _transformQueryResults: function(callback, error, data) {
       if (error !== null) {
         callback(error, data);
+      } else if (!data) {
+        // The only time there should be no data and no error is if the
+        // request to the charmstore times out, or is blocked by a vpn.
+        callback('no entity data returned, can you access the charmstore?');
       } else {
         // If there is a single charm or bundle being requested then we need
         // to wrap it in an array so we can use the same map code.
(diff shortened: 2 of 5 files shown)


hatched:
#### To QA

- Use the GUI via guiproxy on jaas/gijoe to make sure it works as usual.
- Bootstrap a new controller
- Open your `/etc/hosts/` and add `0.0.0.0         api.jujucharms.com`
- Start guiproxy and point it to this new controller.
- Visit `<host>/gui/u/admin/default`
- Log in - you should see a failed charmstore request trying to diambiguate the url but then the model will load anyways.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7911/

jujugui/static/gui/src/app/components/entity-details/entity-details.js
```diff
@@ -150,7 +150,7 @@ class EntityDetails extends React.Component {
     @param {Array} models A list of the entity models found.
   */
   _fetchCallback(error, data) {
```
frankban: Pre-existing: please remove the `@method` entry in this docstring: it's not required and it's wrong.

jujugui/static/gui/src/app/jujulib/charmstore.js
```diff
@@ -83,6 +83,10 @@ var module = module;
     _transformQueryResults: function(callback, error, data) {
       if (error !== null) {
         callback(error, data);
+      } else if (!data) {
+        // The only time there should be no data and no error is if the
```
frankban: As this comment explains, when retrieving an entity from the charm store we can always assume that the response contains data or it is a 404. An empty response is never a legitimate response from that API endpoint, therefore maybe we should turn that case into an error at the API client level. Take it or leave it.

jujugui/static/gui/src/app/components/entity-details/entity-details.js
hatched: removed

jujugui/static/gui/src/app/jujulib/charmstore.js
hatched: I agree.

The `getEntity` call response callback is first wrapped with the `jujulib._wrap` and `charmstore._transformQueryResults` both of which are used by multiple functions. I'm not able to add it to _wrap as it's used by every call so I'd have to create another wrapper. I think that keeping it in _transformQueryResults makes sense because it can't transform data that doesn't exist.

hatched:
Thanks for the reviews/qa :shipit:

jujugui:

PR #3103, opened 2017-07-27
Implement the Client.FullStatus API call.

frankban:
We can consider using this call for the new status view.

### DIFF
--- a/jujugui/static/gui/src/app/store/env/api.js
+++ b/jujugui/static/gui/src/app/store/env/api.js
@@ -2972,7 +2972,6 @@ YUI.add('juju-env-api', function(Y) {
     fullStatus: function(callback) {
       // Define the callback handler.
       const handler = data => {
-        console.log(JSON.stringify(data));
         if (!callback) {
           console.log('data returned by Client.FullStatus API call:', data);
           return;
--- a/jujugui/static/gui/src/app/store/env/api.js
+++ b/jujugui/static/gui/src/app/store/env/api.js
@@ -2705,6 +2705,397 @@ YUI.add('juju-env-api', function(Y) {
     },
 
     /**
+      Return information about all entities in this model.
+
+      @param {Function} callback A callable that must be called once the
+        operation is performed. It will receive two arguments: an error string
+        and an empty object if an error occurred, null and the status result
+        otherwise. The status result is an object similar to the following:
+
+          {
+            // Information about this model.
+            model: {
+              // The model name.
+              name: "default",
+              // The cloud in which the model lives.
+              cloud: "aws",
+              // The region for model machines.
+              region: "eu-central-1",
+              // The current version of the agents running on the model.
+              version: "2.3-alpha1.1",
+              // The optional version available for a model upgrade.
+              availableVersion: "",
+              // The level of the SLA, "unsupported", "essential", standard"
+              // or "advanced";
+              sla: "unsupported",
+              // Information about model status.
+              status: {
+                // The agent status, like "available" or "error".
+                status: "available",
+                // Additional status info if relevant.
+                info: "",
+                // A date object with the datetime of latest status change.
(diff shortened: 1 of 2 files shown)


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7923/

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -2705,6 +2705,397 @@ YUI.add('juju-env-api', function(Y) {
...
 
     /**
+      Return information about all entities in this model.
+
+      @param {Function} callback A callable that must be called once the
+        operation is performed. It will receive two arguments: an error string
+        and an empty object if an error occurred, null and the status result
+        otherwise. The status result is an object similar to the following:
+
+          {
```
hatched: This can be a great resource for anyone implementing this but will this ever change? Is it missing some things or is this everything possible from status? When it gets out of date will it still be helpful or just filling up lines in the file? Maybe a better thing would be a command on how to generate this output?

hatched: This information is basically a duplication of the return object in the function. Why not just add the comments inline there if it's ambiguous?

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -2705,6 +2705,397 @@ YUI.add('juju-env-api', function(Y) {
...
+          }, {});
+        };
+        const makeStatus = data => ({
+          status: data.status,
+          info: data.info,
+          life: data.life,
+          since: new Date(data.since)
+        });
+        const makeMachine = data => ({
+          id: data.id,
```
hatched: Lots of repeated properties here, maybe we should investigate creating a `cloneButModifyTheseProperties` function 🤷‍♂️

jujugui/static/gui/src/test/test_env_api.js
```diff
@@ -3551,6 +3551,66 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
       });
     });
 
+    describe('fullStatus', function() {
+      it('succeeds', function(done) {
+        // Snapshots could help here!
```
hatched: :D indeed

jujugui/static/gui/src/app/store/env/api.js
```diff
@@ -2705,6 +2705,397 @@ YUI.add('juju-env-api', function(Y) {
...
+                  subordinate: false
+                }]
+              }
+            ]
+          }
+    */
+    fullStatus: function(callback) {
+      // Define the callback handler.
+      const handler = data => {
+        console.log(JSON.stringify(data));
```
frankban: No, this is a leftover, good catch!

jujugui/static/gui/src/app/store/env/api.js
frankban: In theory, it should not change. It could be extended, but not modified. There is no command to generate that output, except, of course, the env.fullStatus itself. I think that documenting the value passed to callbacks is always valuable. Inline comments are not required in this case, as the logic is pretty much trivial.

jujugui/static/gui/src/app/store/env/api.js
frankban: I am not sure. I think that the value of doing all of this is making the object structure explicit, which would not be if we implemented a "pass everything through except this and this" logic. You already know I am never keen on injecting data from the network directly into the core application.

jujugui/static/gui/src/app/store/env/api.js
frankban: Removed.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7931/

frankban:
Thanks for the reviews!
:shipit:

jujugui:

Issue #3106, opened 2017-07-28
Issue removing SSH keys


huwshimi:
When I removed an SSH key in the deployment flow it had an error: `Uncaught TypeError: Cannot read property 'length' of null`.

PR #3107, opened 2017-07-28
Fix removing SSH keys.

frankban:
Additionally, do not pass the same key used when creating the model to the addSSHKey ECS: while it's true that it's a noop, an error is still returned and we can easily avoid that.

Also fix some component errors (props required but not provided) and make lint-components block CI in the case of errors so that this won't happen again.

Fixes https://github.com/juju/juju-gui/issues/3106

### DIFF
--- a/jujugui/static/gui/src/app/components/deployment-flow/sshkey/sshkey.js
+++ b/jujugui/static/gui/src/app/components/deployment-flow/sshkey/sshkey.js
@@ -164,13 +164,10 @@ class DeploymentSSHKey extends React.Component {
     const newSSHkeyList = this.state.SSHkeys.filter(key => {
       return key.id !== keyId;
     });
-
-    this.setState({
-      SSHkeys: newSSHkeyList
-    });
-
+    this.setState({SSHkeys: newSSHkeyList});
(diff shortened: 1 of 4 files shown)


jujugui/static/gui/src/app/components/deployment-flow/sshkey/sshkey.js
```diff
@@ -164,13 +164,10 @@ class DeploymentSSHKey extends React.Component {
...
-
-    this.setState({
-      SSHkeys: newSSHkeyList
-    });
-
+    this.setState({SSHkeys: newSSHkeyList});
+    // For now, we only support import all or none.
     if (!newSSHkeyList.length) {
-      this.props.setSSHKeys(null);
+      this.props.setSSHKeys([]);
```
frankban: This is the actual bug fix.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7932/

Lukewh:
QA OK

frankban:
Thanks Luke!
:shipit:

jujugui:

PR #3124, opened 2017-08-02
Initial status pane content implementation.

frankban:
This is just an example of what we can show there: we'll make it prettier, for the time being this is protected by the "status" feature flag.

In order to make required info available, some model API calls and handlers have been extended to include additional data, like workload version, app status and SLA. The SLA information should be available but it's not due to a bug in juju that I'll fix asap.

This branch also includes a fix for the inspect-component script to better handle shape fields.

### DIFF
--- a/jujugui/static/gui/src/app/components/status/status.js
+++ b/jujugui/static/gui/src/app/components/status/status.js
@@ -5,23 +5,363 @@
 /** Status React component used to display Juju status. */
 class Status extends React.Component {
 
+  /**
+    Render the component.
+  */
   render() {
     return (
       <juju.components.Panel
         instanceName="status-view"
         visible={true}>
         <div className="status-view__content">
-          Status
+          {this.renderStatus()}
         </div>
       </juju.components.Panel>
     );
   }
 
+  /**
+    Render the current model status.
+  */
+  renderStatus() {
+    const elements = [];
+    const db = this.props.db;
+    const model = this.props.model;
+    if (!model.environmentName) {
+      // No need to go further: we are not connected to a model.
+      return 'Cannot show the status: the GUI is not connected to a model.';
+    }
+    elements.push(this._renderModel(model));
+    if (db.remoteServices.size()) {
+      elements.push(this._renderRemoteApplications(db.remoteServices));
+    }
+    if (db.services.size()) {
+      elements.push(
+        this._renderApplications(db.services),
+        this._renderUnits(db.services)
+      );
+    }
+    if (db.machines.size()) {
+      elements.push(this._renderMachines(db.machines));
+    }
+    if (db.relations.size()) {
+      elements.push(this._renderRelations(db.relations));
+    }
+    return elements;
+  }
+
+  /**
+    Render the model fragment of the status.
+    @params {Object} model The model attributes.
+  */
+  _renderModel(model) {
+    return (
+      <table key="model">
+        <thead>
+          <tr>
+            <th>Model</th>
+            <th>Cloud/Region</th>
+            <th>Version</th>
+            <th>SLA</th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr key={model.environmentName}>
+            <td>{model.environmentName}</td>
+            <td>{model.cloud}/{model.region}</td>
+            <td>{model.version}</td>
+            <td>{model.sla}</td>
+          </tr>
+        </tbody>
+      </table>
+    );
+  }
+
+  /**
+    Render the remote applications fragment of the status.
+    @params {Object} remoteApplications The remote applications as included in
+      the GUI db.
+  */
+  _renderRemoteApplications(remoteApplications) {
+    const rows = remoteApplications.map(application => {
+      const app = application.getAttrs();
+      const urlParts = app.url.split(':');
+      return (
+        <tr key={app.url}>
+          <td>{app.service}</td>
+          <td>{app.status.current}</td>
+          <td>{urlParts[0]}</td>
+          <td>{urlParts[1]}</td>
+        </tr>
+      );
+    });
+    return (
+      <table key="remote-applications">
+        <thead>
+          <tr>
+            <th>SAAS</th>
+            <th>Status</th>
+            <th>Store</th>
+            <th>URL</th>
+          </tr>
+        </thead>
+        <tbody>
+          {rows.sort(byKey)}
+        </tbody>
+      </table>
+    );
+  }
+
+  /**
+    Render the applications fragment of the status.
+    @params {Object} applications The applications as included in the GUI db.
+  */
+  _renderApplications(applications) {
+    const urllib = this.props.urllib;
+    const rows = applications.map(application => {
+      const app = application.getAttrs();
+      const charm = urllib.fromLegacyString(app.charm);
+      const store = charm.schema === 'cs' ? 'jujucharms' : 'local';
+      const revision = charm.revision;
+      charm.revision = null;
+      return (
+        <tr key={app.name}>
+          <td>{app.name}</td>
+          <td>{app.workloadVersion}</td>
+          <td className={getClass(app.status.current)}>
+            {app.status.current}
+          </td>
+          <td>{app.units.size()}</td>
+          <td>{charm.path()}</td>
+          <td>{store}</td>
+          <td>{revision}</td>
+        </tr>
+      );
+    });
+    return (
+      <table key="applications">
+        <thead>
+          <tr>
+            <th>Application</th>
+            <th>Version</th>
+            <th>Status</th>
+            <th>Scale</th>
+            <th>Charm</th>
+            <th>Store</th>
+            <th>Rev</th>
+          </tr>
+        </thead>
+        <tbody>
+          {rows.sort(byKey)}
+        </tbody>
+      </table>
+    );
+  }
+
+  /**
+    Render the units fragment of the status.
+    @params {Object} applications The applications as included in the GUI db.
+  */
+  _renderUnits(applications) {
+    const formatPorts = ranges => {
+      if (!ranges) {
+        return '';
+      }
+      return ranges.map(range => {
+        if (range.from === range.to) {
+          return `${range.from}/${range.protocol}`;
+        }
+        return `${range.from}-${range.to}/${range.protocol}`;
+      }).join(', ');
+    };
+    const rows = [];
+    applications.each(application => {
+      application.get('units').each(unit => {
+        rows.push(
+          <tr key={unit.id}>
+            <td>{unit.displayName}</td>
+            <td className={getClass(unit.workloadStatus)}>
+              {unit.workloadStatus}
+            </td>
+            <td className={getClass(unit.agentStatus)}>
+              {unit.agentStatus}
+            </td>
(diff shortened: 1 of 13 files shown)


frankban:
QA:
Enable the status flag (i.e. `guiproxy -config 'flags: {"status": true}'`).
Check that the status pane makes sense on multiple models, switch to other models/new model while looking at the status.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7949/

jujugui/static/gui/src/app/components/status/status.js
```diff
@@ -5,23 +5,363 @@
...
+    }
+    if (db.relations.size()) {
+      elements.push(this._renderRelations(db.relations));
+    }
+    return elements;
+  }
+
+  /**
+    Render the model fragment of the status.
+    @params {Object} model The model attributes.
```
huwshimi: Add `@returns`?

jujugui/static/gui/src/app/components/status/status.js
```diff
@@ -5,23 +5,363 @@
...
+          </tr>
+        </tbody>
+      </table>
+    );
+  }
+
+  /**
+    Render the remote applications fragment of the status.
+    @params {Object} remoteApplications The remote applications as included in
+      the GUI db.
```
huwshimi: Add `@returns`?

jujugui/static/gui/src/app/components/status/status.js
```diff
@@ -5,23 +5,363 @@
...
+        <tbody>
+          {rows.sort(byKey)}
+        </tbody>
+      </table>
+    );
+  }
+
+  /**
+    Render the applications fragment of the status.
+    @params {Object} applications The applications as included in the GUI db.
```
huwshimi: Add `@returns`?

jujugui/static/gui/src/app/components/status/status.js
```diff
@@ -5,23 +5,363 @@
...
+        <tbody>
+          {rows.sort(byKey)}
+        </tbody>
+      </table>
+    );
+  }
+
+  /**
+    Render the units fragment of the status.
+    @params {Object} applications The applications as included in the GUI db.
```
huwshimi: Add `@returns`?

jujugui/static/gui/src/app/components/status/status.js
```diff
@@ -5,23 +5,363 @@
...
+        <tbody>
+          {rows.sort(byKey)}
+        </tbody>
+      </table>
+    );
+  }
+
+  /**
+    Render the machines fragment of the status.
+    @params {Object} machines The machines as included in the GUI db.
```
huwshimi: Add `@returns`?

jujugui/static/gui/src/app/components/status/status.js
```diff
@@ -5,23 +5,363 @@
...
+        <tbody>
+          {rows.sort(byKey)}
+        </tbody>
+      </table>
+    );
+  }
+
+  /**
+    Render the relations fragment of the status.
+    @params {Array} relations The relations as included in the GUI db.
```
huwshimi: Add `@returns`?

jujugui/static/gui/src/app/components/status/status.js
huwshimi: Also, I think this should be `@param`

jujugui/static/gui/src/app/components/status/status.js
```diff
@@ -5,23 +5,363 @@
...
+            <td>{model.sla}</td>
+          </tr>
+        </tbody>
+      </table>
+    );
+  }
+
+  /**
+    Render the remote applications fragment of the status.
+    @params {Object} remoteApplications The remote applications as included in
```
huwshimi: `@param`?

jujugui/static/gui/src/app/components/status/status.js
huwshimi: `@param`?

jujugui/static/gui/src/app/components/status/status.js
huwshimi: `@param`?

jujugui/static/gui/src/app/components/status/status.js
huwshimi: `@param`?

jujugui/static/gui/src/app/components/status/status.js
huwshimi: `@param`?

jujugui/static/gui/src/app/components/status/status.js
```diff
@@ -5,23 +5,363 @@
...
+    region: PropTypes.string,
+    sla: PropTypes.string,
+    version: PropTypes.string
+  }).isRequired,
+  urllib: PropTypes.func.isRequired
+};
+
+/**
+  Return an element class name suitable for the given value.
+  @params {String} value The provided value.
```
huwshimi: `@param`?

jujugui/static/gui/src/app/components/status/status.js
```diff
@@ -5,23 +5,363 @@
...
+    case 'error':
+      return 'error';
+  }
+  return '';
+};
+
+/**
+  A compare function for sorting an array by the key property.
+  @params {Object} a The first value.
+  @params {Object} b The second value.
```
huwshimi: `@param`?

jujugui/static/gui/src/app/components/status/status.js
```diff
@@ -5,23 +5,363 @@
...
+  }).isRequired,
+  urllib: PropTypes.func.isRequired
+};
+
+/**
+  Return an element class name suitable for the given value.
+  @params {String} value The provided value.
+  @returns {String} The class name ('ok', 'error' or '').
+*/
+const getClass = value => {
```
huwshimi: Curious as to why this function and `byKey` below are outside the module and not either utils or module methods. Having them here could lead to variable name conflicts due to the way YUI modules are imported.

jujugui/static/gui/src/app/components/status/status.js
```diff
@@ -5,23 +5,363 @@
...
+    @params {Object} applications The applications as included in the GUI db.
+  */
+  _renderApplications(applications) {
+    const urllib = this.props.urllib;
+    const rows = applications.map(application => {
+      const app = application.getAttrs();
+      const charm = urllib.fromLegacyString(app.charm);
+      const store = charm.schema === 'cs' ? 'jujucharms' : 'local';
+      const revision = charm.revision;
+      charm.revision = null;
```
huwshimi: Might be worth adding a comment here as to why this is necessary.

jujugui/static/gui/src/app/components/status/status.js
```diff
@@ -5,23 +5,363 @@
...
         </div>
       </juju.components.Panel>
     );
   }
 
-};
+  /**
+    Render the current model status.
+  */
+  renderStatus() {
```
huwshimi: @hatched would say these methods don't render the element, but rather generate the objects that are used to render them so instead of `renderStatus` they could be `_generateStatus`, but up to you if you want to change them or not.

jujugui/static/gui/src/app/components/status/status.js
```diff
@@ -5,23 +5,363 @@
 /** Status React component used to display Juju status. */
 class Status extends React.Component {
 
+  /**
+    Render the component.
```
frankban: All other methods are called by render, so it seems to me good to have them right after render.
Also, why not putting all React specific methods on top and all user defined ones on the bottom?
Any specific reason to have render at the bottom?

jujugui/static/gui/src/app/components/status/status.js
frankban: Done.

jujugui/static/gui/src/app/components/status/status.js
frankban: I would reply to @hatched that all those methods are called by render in order to do the render job, which is rendering elements. The render prefix makes it clear that these methods are used in the rendering process.

jujugui/static/gui/src/app/components/status/status.js
frankban: Done.

jujugui/static/gui/src/app/components/status/status.js
frankban: Done.

jujugui/static/gui/src/app/components/status/status.js
frankban: Done.

jujugui/static/gui/src/app/components/status/status.js
frankban: Done.

jujugui/static/gui/src/app/components/status/status.js
frankban: Done.

jujugui/static/gui/src/app/components/status/status.js
frankban: Done.

jujugui/static/gui/src/app/components/status/status.js
frankban: Done.

jujugui/static/gui/src/app/components/status/status.js
frankban: Done.

jujugui/static/gui/src/app/components/status/status.js
frankban: They don't have to be methods, they don't need access to the class state. Is there a way to avoid possible name clashes while keeping them as simple functions?

jujugui/static/gui/src/app/components/status/status.js
frankban: Done.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7950/

mitechie:
Awesome! feedback:

  - left align
  - no lines in the content of the tables to reduce visual clutter
  - ubuntu-mono font

jaycee:
QA notes:

Connected on an lxd. It properly showed status for a pre-existing model with the ubuntu application.
I created a new model and deployed ghost into it; status properly showed it allocating. From the cli I then added mysql and made a relation--the status pane updated properly.
Switching between models properly showed the "not connected" msg for a moment, then an updated status for the model I switched to.

QA OK.

frankban:
@huwshimi see suggestions from Rick ^^^ when doing the follow-up branch to make this pretty!

frankban:
Thanks for the reviews!
:shipit:

jujugui:

jujugui/static/gui/src/app/components/status/status.js
huwshimi: I can't remember the exact reason, but I would argue at this point it would be about consistency.

jujugui/static/gui/src/app/components/status/status.js
huwshimi: Again, at this point I would argue it's about consistency, but I'm not too fussed.

PR #3125, opened 2017-08-07
Update the status screen with our styles

huwshimi:


huwshimi:
QA:
- connect to a model
- open the status view
- it should look like our tables

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7957/

jujugui/static/gui/src/app/components/status/status.js
```diff
@@ -4,172 +4,199 @@
 
 /** Status React component used to display Juju status. */
 class Status extends React.Component {
+  /**
```
frankban: I am not sure a class should start with the least interesting non-methods [shrug].
Why these can't be just functions? They really don't need the Status state.

jujugui/static/gui/src/app/components/status/status.js
```diff
@@ -4,172 +4,199 @@
...
+  _byIndex(index, a, b) {
+    if (a[index] < b[index])
+      return -1;
+    if (a[index] > b[index])
+      return 1;
+    return 0;
   }
 
   /**
-    Render the current model status.
```
frankban: Heh, thanks for making it consistent with all other components. In my defense, I am not sure I get this generate vs render semantics :-).

jujugui/static/gui/src/app/components/status/status.js
```diff
@@ -4,172 +4,199 @@
...
-            <td>{model.environmentName}</td>
-            <td>{model.cloud}/{model.region}</td>
-            <td>{model.version}</td>
-            <td>{model.sla}</td>
-          </tr>
-        </tbody>
-      </table>
-    );
+      <juju.components.BasicTable
+        columns={[{
```
frankban: Yes, I think headers and body describe better what's being provided here.

jujugui/static/gui/src/app/components/basic-table/basic-table.js
```diff
@@ -0,0 +1,112 @@
...
+    });
+    const classes = classNames(
+      'twelve-col',
+      {
+        'basic-table__header': header,
+        'basic-table__row': !header
+      });
+    return (
+      <li className={classes}
+        key={index || 0}>
```
frankban: Is there a way of passing more meaningful keys than just the indexes? My understanding is that we would help react if we pass something that really identify the row rather than just the row position. And we have that data in all cases (machine id, unit name, app name etc). We are using meaningful keys in the current version, so we could just allow providing them, like (BasicTable header=[...] body=[[...]] keys=[...])

jujugui/static/gui/src/app/components/basic-table/basic-table.js
```diff
@@ -0,0 +1,112 @@
...
+      }
+      const classes = classNames(conditionalClasses);
+      let content;
+      if (header) {
+        content = column.title;
+      } else {
+        content = columnValues[i] || '';
+      }
+      return (
+        <div className={classes}
```
huwshimi: Typically we use a grid for tabular data as it can be modified to be displayed on a mobile whereas an actual `<table>` can't really be changed.

jujugui/static/gui/src/app/components/basic-table/basic-table.js
```diff
@@ -0,0 +1,112 @@
...
+};
+
+BasicTable.propTypes = {
+  // The columns array should contain objects in the following format:
+  // [{
+  //  title: 'Column title',
+  //  size: 3, // The width of the column (between 1 and 12).
+  //  classes: ['class1', 'class2', ...]
+  // }, ...]
+  columns: PropTypes.array.isRequired,
```
huwshimi: Ah thanks, I wasn't sure how to do a shape inside an array, I'll update that.

jujugui/static/gui/src/app/components/status/status.js
huwshimi: The reason I was using columns and not something like headers is that the columns definition is also used to shape the content as well (it sets the size and classes for the row columns as well as the header). Does that help or would you like to see another name?

jujugui/static/gui/src/app/components/status/status.js
huwshimi: We really don't use tables for anything, it is rarely used at all in HTML these days (see my explanation above).

jujugui/static/gui/src/app/components/basic-table/test-basic-table.js
```diff
@@ -0,0 +1,83 @@
...
+      ['row 2 column 1'],
+      ['row 3 column 1', 'row 3 column 2', 'row 3 column 3']
+    ];
+    const renderer = jsTestUtils.shallowRender(
+      <juju.components.BasicTable
+        columns={columns}
+        rows={rows} />, true);
+    const output = renderer.getRenderOutput();
+    const expected = (
+      <ul className="basic-table twelve-col">
```
huwshimi: We really don't use tables for anything, it is rarely used at all in HTML these days (see my explanation above).

jujugui/static/gui/src/app/components/status/status.js
huwshimi: That's right, they don't need to access state, but they are necessary for the component and if they're a component level method they won't clash with other modules if they're scoped to the component.

jujugui/static/gui/src/app/components/basic-table/basic-table.js
huwshimi: You're right it can be faster for react if these are more unique, I'll try to figure something out, I'd like to not have to pass in extra data if we don't have to.

jujugui/static/gui/src/app/components/basic-table/basic-table.js
```diff
@@ -0,0 +1,112 @@
...
+      'ten-col',
+      'eleven-col',
+      'twelve-col'
+    ];
+  }
+
+  /**
+    Generate the a row.
+    @param header {Boolean} Whether this is a header row.
+    @param columnValues {Array} The list of column values.
```
huwshimi: Updated.

jujugui/static/gui/src/app/components/basic-table/basic-table.js
```diff
@@ -0,0 +1,112 @@
...
+      'nine-col',
+      'ten-col',
+      'eleven-col',
+      'twelve-col'
+    ];
+  }
+
+  /**
+    Generate the a row.
+    @param header {Boolean} Whether this is a header row.
```
huwshimi: Thanks.

jujugui/static/gui/src/app/components/basic-table/basic-table.js
```diff
@@ -0,0 +1,112 @@
...
+      'eight-col',
+      'nine-col',
+      'ten-col',
+      'eleven-col',
+      'twelve-col'
+    ];
+  }
+
+  /**
+    Generate the a row.
```
huwshimi: Thanks.

frankban:
QA is ok.
Some suggestions for a follow-up branch:
- we could make ip addresses clickable in the case 80/443 ports are open;
- we could make error cases use the red shade already used for errors in the inspector;
- the green message could be more readable, maybe make it bolder?
- we could allow clicking on app name to open the corresponding app in the inspector.

jujugui/static/gui/src/app/components/basic-table/basic-table.js
```diff
@@ -0,0 +1,112 @@
+/* Copyright (C) 2017 Canonical Ltd. */
+
+'use strict';
+
+/** Basic table React component used to display data in a table structure. */
+class BasicTable extends React.Component {
+  constructor() {
+    super();
+    this.columnClasses = [
```
hatched: I dislike us using a library when css grid will work just fine

huwshimi: I think that would be fine (without knowing much about CSS grids), but I think this is OK too, as we'd still have to have classes like that to set the different widths and this grid system is used across the company and maintained (and already has support for mobile etc.)

huwshimi:
This has been updated and is ready for another review.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7960/

huwshimi:
Thanks @frankban, @hatched :shipit:

jujugui:

jujugui/static/gui/src/app/components/basic-table/basic-table.js
anthonydillon: Vanilla will begin to support CSS grids soon and in by proxy so will all apps and sites across the company. It is very helpful to have a central place to develop a framework which is rolled out across all the products.

PR #3126, opened 2017-08-07
Do not allow switching models if ECS is committing

### LINKED ISSUE #2971: Block switching models while deploying
When you create a new model in the deployment flow it's possible to then use the model switcher before the new model has been connected to, and the commit successfully completed.

This causes the commit to fail and the model to be in a weird state.

makyo:
Fixes #2971

### DIFF
--- a/jujugui/static/gui/src/app/components/env-switcher/env-switcher.js
+++ b/jujugui/static/gui/src/app/components/env-switcher/env-switcher.js
@@ -116,16 +116,7 @@ class EnvSwitcher extends React.Component {
   handleModelClick(model) {
     const props = this.props;
     this.setState({showEnvList: false});
-    if (this.props.isCommitting()) {
-      const message = 'cannot switch models while deploying.';
-      this.props.addNotification({
-        title: message,
-        message: message,
-        level: 'error'
-      });
-    } else {
-      props.switchModel(model);
-    }
+    props.switchModel(model);
   }
 
   /**
@@ -200,7 +191,6 @@ EnvSwitcher.propTypes = {
   changeState: PropTypes.func.isRequired,
   environmentName: PropTypes.string,
   humanizeTimestamp: PropTypes.func.isRequired,
-  isCommitting: PropTypes.func.isRequired,
   listModelsWithInfo: PropTypes.func,
   switchModel: PropTypes.func.isRequired,
   user: PropTypes.object
--- a/jujugui/static/gui/src/app/components/env-switcher/env-switcher.js
+++ b/jujugui/static/gui/src/app/components/env-switcher/env-switcher.js
@@ -116,7 +116,16 @@ class EnvSwitcher extends React.Component {
   handleModelClick(model) {
     const props = this.props;
     this.setState({showEnvList: false});
-    props.switchModel(model);
+    if (this.props.isCommitting()) {
+      const message = 'cannot switch models while deploying.';
+      this.props.addNotification({
+        title: message,
+        message: message,
+        level: 'error'
+      });
+    } else {
+      props.switchModel(model);
+    }
   }
 
   /**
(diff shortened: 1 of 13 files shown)


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7958/

makyo:
# QA

Add a bundle/lot of services. Click deploy and immediately try to switch models; you should get a notification.

huwshimi:
@makyo Thanks for this!

Strangely, I clicked deploy and then immediately clicked "Start a new model", and I got the message to say "cannot switch models while deploying." but it did actually switch me to the new model!

jujugui/static/gui/src/app/utils/environment-change-set.js
```diff
@@ -434,10 +436,20 @@ YUI.add('environment-change-set', function(Y) {
       this.currentCommit = this._buildHierarchy(true);
       this.currentIndex += 1;
       this.currentLevel = -1;
+      this.committing = true;
       this._commitNext(env, this.currentIndex - 1);
     },
 
     /**
+      Check to see if the current  commit is in progress.
```
frankban: Double space between "current" and "commit".

jujugui/static/gui/src/app/utils/environment-change-set.js
```diff
@@ -374,6 +375,7 @@ YUI.add('environment-change-set', function(Y) {
           window.setTimeout(this._commitNext.bind(this, env, currentIndex), 0);
         } else {
           this.currentLevel = -1;
+          this.committing = false;
```
frankban: Nice, can we demonstrate that, sooner or later, in spite of errors, committing will be reset to false?
If so, could you please add a comment explaining why?

jujugui/static/gui/src/app/components/env-switcher/env-switcher.js
```diff
@@ -116,7 +116,16 @@ class EnvSwitcher extends React.Component {
   handleModelClick(model) {
     const props = this.props;
     this.setState({showEnvList: false});
-    props.switchModel(model);
+    if (this.props.isCommitting()) {
+      const message = 'cannot switch models while deploying.';
```
frankban: s/cannot/Cannot/ ?

jujugui/static/gui/src/app/components/env-switcher/env-switcher.js
```diff
@@ -116,7 +116,16 @@ class EnvSwitcher extends React.Component {
...
+      this.props.addNotification({
+        title: message,
+        message: message,
+        level: 'error'
+      });
```
frankban: Return here and avoid the else block? [shrug]

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7971/

hatched:
Because `switchModel` here doesn't take a callback it could be expected to return a value indicating the success or failure of the call. Up until now we've simply been returning undefined when this failure case hits, but for now we want to indicate to the user why we aren't switching.

I propose that instead of passing the addNotification into here, which will create a notification separate from the users action, we return a boolean value from this function which the caller can use to update the user within it's own UI.

This would allow us to do things like 'cannot switch to already connected model' inline in the profile. Or a notification in the model switcher itself saying that it cannot switch because of x. I believe this also simplifies the codebase a bit because the notification support doesn't need to be included, or tested in `switchModel`

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7972/

jujugui/static/gui/src/app/utils/environment-change-set.js
```diff
@@ -35,6 +35,7 @@ YUI.add('environment-change-set', function(Y) {
       this.changeSet = {};
       this.currentCommit = [];
       this.currentIndex = 0;
+      this.committing = false;
```
hatched: Can you add a test for this functionality plz

hatched:
:shipit:

jujugui:

PR #3137, opened 2017-08-17
Select machines using state

huwshimi:
This allows the machines to be linked to from elsewhere in the GUI.

### DIFF
--- a/jujugui/static/gui/src/app/components/machine-view/machine-view.js
+++ b/jujugui/static/gui/src/app/components/machine-view/machine-view.js
@@ -369,10 +369,12 @@ class MachineView extends React.Component {
           dbAPI={props.dbAPI.reshape(propTypes.dbAPI)}
           dropUnit={this._dropUnit.bind(this)}
           key={container.id}
-          machine={container}
           machineAPI={{
             machine: container,
-            removeUnit: this._removeUnit.bind(this)
+            removeUnit: this._removeUnit.bind(this),
+            selected: selected.container ? selected.container === container.id :
+              selected.machine === container.id,
+            selectMachine: this.selectMachine.bind(this)
           }}
           modelAPI={props.modelAPI.reshape(propTypes.modelAPI)}
           parseConstraints={props.parseConstraints}
@@ -472,7 +474,7 @@ class MachineView extends React.Component {
     return (
       <juju.components.MachineViewAddMachine
         acl={props.acl.reshape(propTypes.acl)}
-        close={this._closeAddMachine.bind(this)}
+        close={this._closeAddContainer.bind(this)}
         modelAPI={props.modelAPI.reshape(propTypes.modelAPI)}
         parentId={this._getSelected().machine}
         series={props.series}
--- a/jujugui/static/gui/src/app/components/machine-view/machine-view.js
+++ b/jujugui/static/gui/src/app/components/machine-view/machine-view.js
@@ -32,7 +32,7 @@ class MachineView extends React.Component {
   }
 
   componentDidMount() {
-    this.selectMachine(this._getFirstMachineId(this.props.machines));
+    this.selectMachine(this._getFirstMachineId(this.props.dbAPI.machines));
   }
 
   /**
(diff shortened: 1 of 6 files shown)


jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7982/

jujugui/static/gui/src/app/components/machine-view/machine-view.js
```diff
@@ -19,40 +19,42 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
 'use strict';
 
 class MachineView extends React.Component {
-  constructor(props) {
-    super(props);
+  constructor() {
+    super();
```
frankban: Shouldn't we always call super with props? If that's not the case, could you please ad a comment?

jujugui/static/gui/src/app/components/machine-view/machine-view.js
```diff
@@ -227,7 +229,32 @@ class MachineView extends React.Component {
...
+    @returns {Object} The selected machine and container.
+  */
+  _getSelected() {
+    const machine = this.props.machine;
+    let selected = {
+      machine: null,
+      container: null
+    };
+    if (machine) {
+      // The ID will be in the format "machine-id/container-type/container-id"
```
frankban: It is possible to parse the id using db.machines.parseMachineName(), rather than reimplementing the logic here.

frankban: Or we could pass a machine instance from the db alternatively, already parsed, rather than a string?

huwshimi:
QA:
- open the machine view and add a few machines and containers
- click on the machines and containers and check that the state updates

jujugui/static/gui/src/app/components/machine-view/machine-view.js
huwshimi: We actually rarely pass props into the constructor as they're only required if you need to access the props in the constructor e.g. setting up state.

jujugui/static/gui/src/app/components/machine-view/machine-view.js
huwshimi: In the machine view we need to know what both the selected machine and selected container are at the same time. Unfortunately the parsing in `parseMachineName` returns that info in an inconsistent way. I think what I'll do is implement an intermediate method that normalises the results in the status view.

jujugui/static/gui/src/app/models/models.js
```diff
@@ -1505,14 +1505,14 @@ YUI.add('juju-models', function(Y) {
         return {
           parentId: null,
           containerType: null,
-          number: parseInt(name, 10)
+          number: name
```
huwshimi: The machine/container ids now need to be strings so that they work with uncommitted ids e.g. 'new0'. I don't think this code has been used in a long while.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7987/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7988/

jujugui/static/gui/src/app/components/machine-view/machine-view.js
frankban: Doesn't the constructor also set up this.props? From https://facebook.github.io/react/docs/state-and-lifecycle.html:
"Class components should always call the base constructor with props". Surely it does not hurt, and we keep the same signature when overriding the constructor?

hatched:
Looks like this now has conflicts @huwshimi

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/7998/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/8000/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/8001/

hatched:
:shipit:

jujugui:

PR #3155, opened 2017-08-23
Import SSH keys from launchpad

### LINKED ISSUE #3142: JaaS public SSH key import list does not include Launchpad
As per https://bugs.launchpad.net/juju/+bug/1711305, in JAAS, SSH dropdown does not include launchpad import.

makyo:
fixes #3142

### DIFF
--- a/jujugui/static/gui/src/app/components/deployment-flow/deployment-flow.js
+++ b/jujugui/static/gui/src/app/components/deployment-flow/deployment-flow.js
@@ -40,6 +40,7 @@ class DeploymentFlow extends React.Component {
       paymentUser: null,
       region: this.props.region,
       sshKeys: [],
+      lpUsernames: [],
       // The list of term ids for the uncommitted applications.
       terms: this._getTerms() || [],
       // Whether the user has ticked the checked to agree to terms.
@@ -229,6 +230,16 @@ class DeploymentFlow extends React.Component {
   }
 
   /**
+    Store the provided Launchpad usernames in state.
+
+    @method _setLaunchpadUsernames
+    @param {Array} usernames The list of Launchpad usernames
+  */
+  _setLaunchpadUsernames(usernames) {
+    this.setState({lpUsernames: usernames});
+  }
+
+  /**
     Store the provided AWS virtual private cloud value in state.
     In the case the value is set, also set whether to force Juju to use the
     given value, even when it fails the minimum validation criteria.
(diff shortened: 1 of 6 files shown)


makyo:
# QA

Create a model. In the deployment flow, add a launchpad username, but no other keys. After deployment, SSH into a unit from a machine with private key matching a public key on Launchpad.

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/8013/

jujugui:
http://ci-cge.jujugui.org:8080//job/juju-gui/8014/

jujugui/static/gui/src/app/components/deployment-flow/deployment-flow.js
```diff
@@ -229,6 +230,16 @@ class DeploymentFlow extends React.Component {
   }
 
   /**
+    Store the provided Launchpad usernames in state.
+
+    @method _setLaunchpadUsernames
+    @param {Array} usernames The list of Launchpad usernames
```
frankban: Final dot.

urosj:
+1 that username name should be used  from the macaroon. This will work in every case. So, user should be able to just chane it. Not all users are LP users, import will just fail when they are invalid (in case of pure sso user). We should also track the fail vs success ratio (comulative metric).

This addition can be dobe in new PR.

urosj:
Potential danger of not using a name from the macaroon is that another person accidentally gets access.

urosj:
Adding users, what if we provide link to LP so that one can check it's correct user? Same with GH? (Again ne PR)

jujugui/static/gui/src/app/components/deployment-flow/deployment-flow.js
```diff
@@ -344,6 +355,22 @@ class DeploymentFlow extends React.Component {
         },
         ecsOptions);
     }
+    if (this.state.lpUsernames.length) {
+      this.props.importSSHKeys(
+        this.props.getUserName(),
+        this.state.lpUsernames.map(key => `lp:${key}`),
```
hatched: s/key/username 🤷‍♂️

jujugui/static/gui/src/app/components/deployment-flow/deployment-flow.js
```diff
@@ -344,6 +355,22 @@ class DeploymentFlow extends React.Component {
...
+        (error, data) => {
+          if (!error) {
+            return;
+          }
+          this.props.addNotification({
+            title: 'Cannot import SSH keys',
+            message: `Cannot import SSH keys: ${error}`,
+            level: 'error'
+          });
+        }
```
hatched: Can you add `console.log(error)` here as well plz

makyo:
:shipit: with a followup branch on the way.

jujugui:

PR #3184, opened 2017-08-31
Autofill launchpad name

makyo:


jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8039/

jujugui/static/gui/src/app/init/component-renderers-mixin.js
```diff
@@ -869,6 +869,7 @@ const ComponentRenderersMixin = (superclass) => class extends superclass {
         stats={this.stats}
         updateCloudCredential={
           controllerAPI.updateCloudCredential.bind(controllerAPI)}
+        user={this.user}
```
huwshimi: Thanks for remembering this!

jujugui/static/gui/src/app/components/deployment-flow/sshkey/sshkey.js
```diff
@@ -323,6 +323,13 @@ class DeploymentSSHKey extends React.Component {
         </div>
       );
     } else if (this.state.addSource === 'launchpad') {
+      let username = undefined;
```
huwshimi: `let username;` would set it to undefined as well, right?

jujugui/static/gui/src/app/components/deployment-flow/sshkey/sshkey.js
```diff
@@ -442,7 +450,8 @@ DeploymentSSHKey.propTypes = {
   cloud: PropTypes.object,
   getGithubSSHKeys: PropTypes.func.isRequired,
   setLaunchpadUsernames: PropTypes.func.isRequired,
-  setSSHKeys: PropTypes.func.isRequired
+  setSSHKeys: PropTypes.func.isRequired,
+  user: PropTypes.func
```
huwshimi: Should this be an object like in deployment-flow.js?

jujugui/static/gui/src/app/components/deployment-flow/sshkey/test-sshkey.js
```diff
@@ -44,11 +44,16 @@ describe('DeploymentSSHKey', function() {
...
   // Render the component and return the instance and the output.
-  const render = (cloudType, _getGithubSSHKeys) => {
+  const render = (cloudType, _getGithubSSHKeys, user) => {
     let cloud = null;
     if (cloudType) {
       cloud = {cloudType: cloudType};
     }
+    let userInstance = null;
+    if (user) {
+      userInstance = function() {};
```
huwshimi: No I really am confused about whether this should be a function or an object :).

huwshimi:
QA notes: selecting "Launchpad" from the list of sources auto-filled my username, but I got an error in my console:
`Warning: setState(...): Cannot update during an existing state transition (such as within `render` or another component's constructor). Render methods should be a pure function of props and state; constructor side-effects are an anti-pattern, but can be moved to `componentWillMount`.`

jujugui/static/gui/src/app/components/deployment-flow/sshkey/sshkey.js
makyo: I don't even remember now. At one point, PropTypes yelled at me that it was expecting a func, not an object. I'll dig into that again.

makyo:
@huwshimi Will look into that. I get that pretty regularly throughout, but understand that it should be avoided.

jujugui/static/gui/src/app/components/deployment-flow/sshkey/sshkey.js
jaycee: This should be an obj; you're using props.user.displayName above, indicating this is the user class object.

jaycee: That said, you're only using the displayName, and that doesn't change for a user once logged in, so you could redo this to pass in only the displayName and do props.String.

I don't think that actually *gets* you anything except that it might be easier for testing, but :: shrugs :: ?

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8046/

makyo:
Thanks all :shipit:

jujugui:

PR #3191, opened 2017-09-05
Convert remaining components to modules

hatched:
Includes passing test-js suite
Includes running app.


jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8047/

jujugui/templates/index.html.mako
```diff
@@ -347,7 +347,49 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+            'app-renderer-extension',
+            'cookie',
+            'querystring',
+            'event-key',
+            'event-touch',
+            'model-controller',
+            'FileSaver',
+            'ghost-deployer-extension',
+            'environment-change-set',
+            'relation-utils', 'yui-patches'], function(Y) {
```
huwshimi: What's all this for?

hatched: Ahh yes, it's ugly isn't it?

It will be cleaned up as we move more and more away from YUI modules to require modules. Basically these are all the dependencies that used to be included by the old app.js. We still need those dependencies added to the YUI instance for now.

huwshimi: Ah yes, I see, thanks for the clarification.

PR #3203, opened 2017-09-12
Convert bundle importer to module

hatched:


jujugui/static/gui/src/app/init.js
```diff
@@ -1619,7 +1620,7 @@ class GUIApp {
     this.db.destroy();
     this.endpointsController.destroy();
     // Detach event listeners.
-    const remove = document.removeEventListener;
+    const remove = document.removeEventListener.bind(document);
```
hatched: Without this, the `remove()` calls no longer have access to the proper `document` context so it never removes the events causing cascading test failures

jujugui/static/gui/src/app/init/bundle-importer.js
```diff
@@ -0,0 +1,683 @@
+/* Copyright (C) 2017 Canonical Ltd. */
```
hatched: No changes were made, just porting over and updating from var's to lets, etc.

jujugui/static/gui/src/app/init/test-bundle-importer.js
```diff
@@ -1,42 +1,26 @@
-/*
```
hatched: The tests were simply ported over as well.

jujugui/static/gui/src/app/utils/bakery-utils.js
```diff
@@ -1,28 +1,13 @@
...
+/* Copyright (C) 2017 Canonical Ltd. */
 'use strict';
 
 /**
   Helper functions for creating and working with the bakery.
 */
 
 const Notification = require('../components/notification/notification');
+const React = require('react');
+const ReactDOM = require('react-dom');
```
hatched: Just a driveby looks like they were missed in @Lukewh 's branch which I had to land to continue on this...nbd!

jujugui/static/gui/src/app/utils/bundle-importer.js
```diff
@@ -1,760 +0,0 @@
...
-SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Affero
-General Public License for more details.
-
-You should have received a copy of the GNU Affero General Public License along
-with this program.  If not, see <http://www.gnu.org/licenses/>.
-*/
-
-'use strict';
-
-YUI.add('bundle-importer', function(Y) {
```
hatched: bye bye more YUI

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8067/

jujugui/static/gui/src/test/factory.js
```diff
@@ -18,7 +18,7 @@ with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 'use strict';
 
-YUI(GlobalConfig).add('juju-tests-factory', function(Y) {
+YUI.add('juju-tests-factory', function(Y) {
```
hatched: This was wrong, has been wrong forever, not sure how it ever worked...

jujugui/static/gui/src/test/test_app.js
```diff
@@ -1,2226 +0,0 @@
-/*
```
hatched: We'll probably want to revisit this file to write the tests for the init, but it being here causes issues when trying to get a running suite.

jujugui/static/gui/src/test/test_autodeploy_extension.js
```diff
@@ -1,185 +0,0 @@
-/*
```
hatched: These tests and the module functionality has been moved over to the init.

jujugui/static/gui/src/test/test_app.js
huwshimi: I wonder if we could skip these tests instead of removing them, then we can remove the parts as we port them over to keep track of what has been done. Thoughts?

hatched: I figured that because we would be rewriting them and only making sure we cover the same functionality the code itself probably wouldn't be very helpful whereas having it in here could cause some issues with dependency loading that we don't need to deal with

PR #3209, opened 2017-09-14
Migrate tests to the new suite

huwshimi:


jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8075/

hatched:
It looks like the test failures are legit

huwshimi:
I've moved some of these tests over in other branches and the remaining ones will be moved over as we de-YUI them.

PR #3215, opened 2017-09-15
Add some init tests

huwshimi:


jujugui/static/gui/src/app/test-init.js
```diff
@@ -56,4 +66,356 @@ describe('init', () => {
     assert.equal(shortcuts.children.length > 0, true,
       'The shortcuts component did not render');
   });
+
+  describe('_listAndSwitchModel', function() {
```
huwshimi: The test have mostly been copied across with a few minor adjustments to make them work with init.js

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8080/

Issue #3230, opened 2017-09-19
new-init: loading the base url doesn't show the profile


huwshimi:
In the current GUI if you visit / it redirects you to your profile.

hatched:
As far as I can tell this works in the app.js as a side effect of a race with the login process.

huwshimi:
Fixed in #3232.

PR #3232, opened 2017-09-20
Fix the redirect to the profile when no url is provided

huwshimi:
Fixes: #3230.


huwshimi:
QA:
- load /
- you should be redirected to your profile.

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8118/

PR #3238, opened 2017-09-26
Allow custom ports in the test runner

hatched:


jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8142/

.gitignore
```diff
@@ -24,6 +24,7 @@ jujugui.egg-info/
 jujugui/static/build/*
 jujugui/static/gui/build/
 karma-mocha-old.conf.js
+karma.conf.js
```
frankban: Is this something we generate? Just to make sure, since it wasn't excluded before.

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8143/

scripts/test-js.sh
```diff
@@ -11,6 +11,12 @@ else
   PORT=8888
 fi
 
+# Modify the karma.conf.js to have the port for the proxy. This allows us to
+# use different ports in CI.
+# This needs to be placed before pserve starts or else the server will not
+# start properly.
+cp karma.conf.js.tmpl karma.conf.js
+sed -i -e 's/{TEST_PORT}/'$PORT'/' karma.conf.js
```
frankban: Does this work? Why do we need quotes in '$PORT'?

.gitignore
hatched: before we weren't generating it, it was just there. But now we only want to keep the template around.

scripts/test-js.sh
hatched: yes :)

PR #3278, opened 2017-10-13
Move store state out of root key

### LINKED ISSUE #2963: Navigating 'back' doesn't properly re-generate state
This issue is exposed by #2962 

Steps:
- Be connected to a model ie) /u/{username}/{model}
- Open the store
- Search for "ghost"
- View the charm
- Hit the browsers 'back' button until you reach /store
- Click another charm
The state will now be incorrect and so will the url. This is because when navigating 'back' the state that is generated see's that there is something in the 'root' portion of the url and stops generating state for the model component. This will also introduce another issue as the state will be parsed into the user component and not into the model component as it is coming from the url and will need to be disambiguated which may expose other complications.

hatched:
Having the 'store' state handled by root was causing a number of conflicts in app state. To resolve this, store has been moved to its own key.

Fixes #2963

### DIFF
--- a/jujugui/static/gui/src/app/components/deployment-flow/direct-deploy/direct-deploy.js
+++ b/jujugui/static/gui/src/app/components/deployment-flow/direct-deploy/direct-deploy.js
@@ -81,7 +81,7 @@ class DeploymentDirectDeploy extends React.Component {
     this.props.changeState({
       gui: {deploy: null},
       profile: null,
-      root: 'store',
+      store: '',
       special: {dd: null}
     });
   }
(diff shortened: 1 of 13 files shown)


hatched:
#### To QA

Use the store in every way you can think of, from many different places and keep an eye on the url, it should all still work as expected.

Note:
If you are connected to a model, then view your profile, then show the store, the URL will get out of sync because we don't actually disconnect from the model when you enter the profile. This is a separate existing issue that will be solved in a follow-up.

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8198/

hatched:
@Lukewh so there isn't really any convention for this anywhere but my rational was that we wanted the store to be shown, but not show a specific entity so `{store: ''}` seemed ok. I did consider `{store: true}` but because other values are `{store: 'u/hatch/ghost'}` having a boolean there seemed like mixing types.

hatched:
Thanks all :shipit:

jujugui:

PR #3302, opened 2017-11-06
Prototype loading hints

huwshimi:


huwshimi:
QA:
- load /new in a private browser
- see that you get hints of the elements that will load

jujugui:
Build finished.

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8236/

hatched:
Smart implementation idea @huwshimi and this works quite well with /new or when going straight into a model. However it's a little odd when you see the 'element hints' and then you're dropped right into the profile or entity details.

huwshimi:
@hatched yeah, I noticed that as well, maybe there's no way to do this properly...

cassiocassio:
screenshot from 2017-11-07 17-19-14

cassiocassio:
Ooo @huwshimi 
Interesting. Nice idea. I fear my connection at home and work is often too fast to feel the need for this, but that's not try for many,

I made a video and took some screengabs to see what Facebook do – their implementation is nice.

I like the way if makes good-enough guesses about what the content will be, that even when it's a bit wrong the illusion if complete.

You can see from the timer it takes perhaps 2 seconds for the four states:

screen shot 2017-11-07 at 17 22 28

screen shot 2017-11-07 at 17 22 36

screen shot 2017-11-07 at 17 22 40

screen shot 2017-11-07 at 17 22 48

PR #3355, opened 2017-11-21
Add bundle URL annotations to each application when deploying a bundle

hatched:
So that the UI can know which applications were deployed from bundles, this branch adds the `bundle-url` annotation to each application deployed that way.

### DIFF
--- a/jujugui/static/gui/src/app/init/bundle-importer.js
+++ b/jujugui/static/gui/src/app/init/bundle-importer.js
@@ -19,14 +19,15 @@ class BundleImporter {
   /**
     Import a bundle YAML into the current model.
     @param {String} bundleYAML The bundle YAML to deploy.
+    @param {String} bundleURL The bundle URL that the YAML represents.
   */
-  importBundleYAML(bundleYAML) {
+  importBundleYAML(bundleYAML, bundleURL) {
     this.db.notifications.add({
       title: 'Fetching bundle data',
       message: 'Fetching detailed bundle data, this may take some time',
       level: 'important'
     });
-    this.fetchDryRun(bundleYAML, null);
+    this.fetchDryRun(bundleYAML, bundleURL, null);
   }
 
   /**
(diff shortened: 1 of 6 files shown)


hatched:
#### To QA

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8297/

jujugui/static/gui/src/app/init/bundle-importer.js
```diff
@@ -106,9 +110,49 @@ class BundleImporter {
...
     this.importBundleDryRun(changes);
   }
 
+  /**
+    Adds the bundle-url annotation to the changeset.
+    @param {String} bundleURL THe bundle url to save to the annotations.
+    @param {Array} changes The bundle changeset.
+  */
+  _addBundleURLAnnotation(bundleURL, changes) {
+    // For each different application records and add an annotation for each one.
```
frankban: This sentence seems a bit weird.

jujugui/static/gui/src/app/init/bundle-importer.js
```diff
@@ -106,9 +110,49 @@ class BundleImporter {
...
       return;
     }
+    // Add the bundle id annotation record.
+    changes = this._addBundleURLAnnotation(bundleURL, changes);
     this.importBundleDryRun(changes);
   }
 
+  /**
+    Adds the bundle-url annotation to the changeset.
+    @param {String} bundleURL THe bundle url to save to the annotations.
```
frankban: URL

mitechie:
Can we go ahead and split/store two annotations in there. One that's just the bundle name and one that's the bundle-url for full reference? It'll mean the front end can do a better job of just grabbing the name field and grouping stuff later.

mitechie:
When using a local bundle we should add an incrementor to the bundle-url so that it's bundle-0, bundle-1, etc so that doing it with more than one bundle will not group things accidently. 

I also think using local: or something would be good to call out.

jujugui/static/gui/src/app/init/bundle-importer.js
hatched: done

jujugui/static/gui/src/app/init/bundle-importer.js
hatched: Yeah that is a lot weird, rewritten.

hatched:
After an offline discussion with @mitechie we've decided to leave the `bundle-url` annotation as the applications should be grouped by the url and not some inferred name to eliminate grouping issues.

The local bundle file name should be incremented because the GUI will duplicate all applications instead of replacing/upgrading existing ones.

hatched:
:shipit:

jujugui:

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8311/

PR #3383, opened 2017-12-08
Tweak font-size of headings within entity READMEs

### LINKED ISSUE #2977: h3 and h4 are confusing/backwards in appearance
In looking at the Canonical Kubernetes readme I had a hard time understanding this section and I realized it was because I had confused a lower heading (h4) with the main section vs the h3 above it. 

The styles should be adjusted so that h3 and h4 are able to be told as a higher/lower heading style. 
h3-h4-confusion

barrymcgee:
Fixes #2977

### DIFF
--- a/jujugui/static/gui/src/app/components/entity-details/content/_content.scss
+++ b/jujugui/static/gui/src/app/components/entity-details/content/_content.scss
@@ -100,24 +100,28 @@
     }
 
     h1 {
-      font-size: 1.4375rem;
+      font-size: 2.8125rem;
     }
 
     h2 {
-      font-size: 1.4375rem;
+      font-size: 2rem;
     }
 
     h3 {
-      font-size: 1.2rem;
+      font-size: 1.4375rem;
     }
 
     h4 {
-      font-size: 1.1rem;
+      font-size: 1.25rem;
     }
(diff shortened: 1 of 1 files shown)


barrymcgee:
This change updates the font-size of headings within READMEs to match the Ubuntu design spec:

image

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8359/

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8360/

huwshimi:
Thanks @barrymcgee, these look right according to the spec, but I wonder if the headings end up being a bit oversized e.g.
headings
These designs have them toned down a bit: https://github.com/CanonicalLtd/juju-design/issues/200#issuecomment-311612819 so I wonder if we can go somewhere in between? Or drop everything down a style (e.g. the markdown generates h1 tags, but we might want them to appear like h2s). Maybe @spencerbygraves has some thoughts.

barrymcgee:
@huwshimi I've had a meeting with Spencer and Martin about this and we've realised that this is no the correct approach, as there is a README embedded inception thing going on here. Thanks.

PR #3405, opened 2018-01-05
Add Get Started and Charm Details links to the inspector header

hatched:


hatched:
#### To QA

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8389/

makyo:
QA okay

jujugui/static/gui/src/app/components/inspector/header/header.js
```diff
@@ -39,6 +39,60 @@ class InspectorHeader extends React.Component {
...
+    });
+  }
+
+  /**
+    Generates and returns the JSX for the links in the inspector header.
+    @returns {Object} JSX links.
+  */
+  _generateLinks() {
+    if (!this.props.showLinks) {
+      return;
```
huwshimi: This could return `null` but no biggy.

hatched:
Thanks! :shipit:

jujugui:

cassiocassio:
Now I play with the GetStarted and GetStartedFree versions I'm unconvinced about the text being black and unclickable:

screen shot 2018-01-09 at 15 06 16

- it's right by the charm-name and looks pretty important in rgb(51, 51, 51)

- the feature is aimed at new-ish users who've come via "build a model" on jujucharms.com

- If a user hasn't yet seen getstarted panels with the majority of charms, it's going to -look- very helpful and almost always disappoint. Is it telling me something? Is it meant to be information - if it's not clickable? 

- I see we can use it to somewhat-shame/encourage charm-authors to add the feature, but they're a small audience compared to the many charm-deployers who will be left wondering what it means, or, even with a tooltip, why they're being teased and tantalised by an unimplemented feature.

I say remove it. 

Alternative routes to helping the world discover, ask for, and provide this feature:

1. What are our top-10 charm deployments, and who can write those getstarteds?
 
2. How can me make adding/noticing the omission of  getstarted super-easy to do 
 -  as part of the CLI charm push workflow? or
 -  as part of the profile/charm and profile/bundle screens?

hatched:
The problem with removing it is that we don't always know right when the inspector opens if we have one. So the links bounce around in those situations (slow connections for example). 

The colour of the text could be changed to a lighter grey.

barrymcgee:
I also think it should be removed if not available - to stop the "Charm details" link bouncing around, we could swap them around so the "Charm details" is the first item in the list as it's always shown?

hatched:
good call @barrymcgee we can do that.

cassiocassio:
+1

PR #3421, opened 2018-01-17
Remove basic table item hover bug

barrymcgee:


jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8421/

barrymcgee:
## QA

I found that this "white background on item hover" was never really working well to start with, see:

https://cl.ly/2A0d0V0k0E36

As this is not a traditional link, the hover behaviour is not activated by simply hovering the item. Only when clicking the item is the hover behaviour applied and it then persists, even when the cursor is no longer near it. 

By removing the hover styling as per this change, we get the desired behaviour.

https://cl.ly/3B2H2V0U041b

hatched:
I'm not sure that this is correct. This also removes the hover when the row is not expanded. Looking at the annotated design you'll see that the hovered row has the white background. So I think the proper fix would be when the row is hovered, but not expanded.

barrymcgee:
@hatched Do you get the same behaviour when you hover as I do (in the linked GIFs)? I found hover behaviour doesn't work as expected when working on Inspector tabs also and assumed it was due to D3.js - perhaps it's my machine?

hatched:
d3 is only used for the canvas (the circles and lines) and shouldn't have any effect on the React components. The second gif appears to be broken :)

barrymcgee:
@hatched I know it shouldn't but native browser hover behaviour for me doesn't work as expected in the Gui - it's like the Gui isn't aware of the mouse until I click the canvas, is this only my set-up or do you experience this too? See this video for illustration:

https://cl.ly/0y3L2O0a2f2G

hatched:
Closing this, as per our discussion, something is broken locally on your machine :)

PR #3425, opened 2018-01-18
Link credentials in profile

huwshimi:
Link the credentials in the model page to the credentials page.

### DIFF
--- a/jujugui/static/gui/src/app/components/profile/model-list/model-list.js
+++ b/jujugui/static/gui/src/app/components/profile/model-list/model-list.js
@@ -127,6 +127,16 @@ class ProfileModelList extends React.Component {
   }
 
   /**
+    Navigate to a credential.
+    @param credential {String} The credential name.
+  */
+  _handleCredentialClick(credential) {
+    this.props.changeState({
+      hash: `credentials/${credential}`
+    });
+  }
+
+  /**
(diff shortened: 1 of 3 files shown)


huwshimi:
QA:
- load the new profile
- expand a model row
- click on the credential name
- it should take you to the credentials section of the profile (in the future this should probably do something like highlight the correct row)

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8425/

jujugui/static/gui/src/app/components/profile/model-list/model-list.js
```diff
@@ -126,6 +126,16 @@ class ProfileModelList extends React.Component {
     this._showConfirmation(model);
   }
 
+  /**
+    Navigate to a credential.
+    @param credential {String} The credential name.
+  */
+  _handleCredentialClick(credential) {
+    this.props.changeState({
+      hash: `credentials/${credential}`
```
hatched: nice idea

hatched:
:shipit:

jujugui:

cassiocassio:
We don’t have an expanded style for creds because there’s not much useful
extra stuff to show.

It doesn’t feel right to open the creds in update/edit mode as that’s a
rare action.

What’s the right high lit style? @spencer?

PR #3437, opened 2018-01-24
Add tooltips to the icons in the profile bundle list

huwshimi:


huwshimi:
QA:
- load the new profile
- open the bundle list
- there should be a list of icons for each bundle
- hover over an icon
- a tooltip should appear
- close the profile and search for a bundle
- the bundles should have icons as normal

huwshimi:
[Display tooltips for bundle icons](https://trello.com/c/wgMFwkk5/230-display-tooltips-for-bundle-icons)

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8448/

jujugui/static/gui/src/app/components/icon-list/icon-list.js
```diff
@@ -0,0 +1,105 @@
...
+'use strict';
+
+const PropTypes = require('prop-types');
+const React = require('react');
+
+/**
+  Display a list of icons for an entity.
+*/
+class IconList extends React.Component {
+  constructor() {
```
frankban: The constructor usually receives props that must be propagated to super.
Why are we preventing that?

jujugui/static/gui/src/app/components/icon-list/icon-list.js
```diff
@@ -0,0 +1,105 @@
...
+    super();
+  }
+
+  /**
+    Generate the store state for the item.
+    @param {String} id The entity id.
+    @returns {Object} The store state.
+  */
+  _generateStoreState(id) {
+    // TODO frankban: it should be clear whether this id is legacy or not.
```
frankban: This is a new component so why not making it part of the contract already and handling ambiguity on the caller, rather then adding something to my TODO list? ;-)

jujugui/static/gui/src/app/components/icon-list/icon-list.js
```diff
@@ -0,0 +1,105 @@
...
+  Display a list of icons for an entity.
+*/
+class IconList extends React.Component {
+  constructor() {
+    super();
+  }
+
+  /**
+    Generate the store state for the item.
+    @param {String} id The entity id.
```
frankban: This should be URL object rather then a string.

jujugui/static/gui/src/app/components/icon-list/icon-list.js
```diff
@@ -0,0 +1,105 @@
...
+    return {
+      profile: null,
+      search: null,
+      store: url.path()
+    };
+  }
+
+  /**
+    Generate the store URL for an entity.
+    @param {String} id The entity id.
```
frankban: It'd be nice if internally we only pass URL objects, rather than strings. Strings are always ambiguous (is it a new style URL, and old one, does it have the schema, etc).

jujugui/static/gui/src/app/components/icon-list/icon-list.js
```diff
@@ -0,0 +1,105 @@
...
+  */
+  _generateStoreURL(id) {
+    return this.props.generatePath(this._generateStoreState(id));
+  }
+
+  /**
+    Show the entity details when clicked.
+
+    @method _handleItemClick
+    @param {String} id The entity id.
```
frankban: Ditto.

jujugui/static/gui/src/app/components/icon-list/icon-list.js
```diff
@@ -0,0 +1,105 @@
...
+    return (
+      <ul className="icon-list">
+        {this._generateIcons()}
+      </ul>);
+  }
+
+};
+
+IconList.propTypes = {
+  applications: PropTypes.arrayOf(PropTypes.shape({
```
frankban: Should we use shapeup.shape here?

jujugui/static/gui/src/app/components/icon-list/icon-list.js
hatched: Because nothing is being added in the constructor, you probably don't need to define one.

jujugui/static/gui/src/app/components/icon-list/icon-list.js
hatched: Not long ago I expanded the url lib to add this: https://github.com/juju/juju-gui/blob/develop/jujugui/static/gui/src/app/jujulib/urls.js#L342

jujugui/static/gui/src/app/components/icon-list/icon-list.js
hatched: Yeah this would be nice....I'm not sure how that would work in practice though as we frequently have just dealt with strings which are returned from the api.

hatched:
@huwshimi This still needs updating and it's now conflicted - just fyi :)

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8465/

hatched:
:shipit:

jujugui:

PR #3456, opened 2018-02-05
Reload application config on controller login

### LINKED ISSUE #3444: shell icon not visible right after logging in
On staging:
- log out;
- log in again;
- switch to a model and the shell icon is not present;
- refresh the page and the shell icon appears.

hatched:
Fixes #3444 

Because the values in the application config may change after the server knows who the user is we want to refresh the config after logging in.

### DIFF
--- a/jujugui/static/gui/src/app/test-init.js
+++ b/jujugui/static/gui/src/app/test-init.js
@@ -1102,6 +1102,7 @@ describe('init', () => {
       app._renderUserMenu = sinon.stub();
       app.maskVisibility = sinon.stub();
       app.loginToAPIs = sinon.stub();
+      app.reloadConfigFile = sinon.stub();
     };
 
     beforeEach(() => {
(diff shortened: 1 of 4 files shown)


hatched:
#### To QA

To properly QA this requires a full environment but there is a workaround that's a little bit of a mess...

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8468/

jujugui/static/gui/src/app/test-init.js
```diff
@@ -1242,6 +1243,11 @@ describe('init', () => {
         const args = app.state.changeState.args[0];
         assert.equal(args.length, 1, 'changeState num args');
         assert.deepEqual(args[0], {root: null}, 'changeState args');
+        // We are testing this here because this gets called when successfully
+        // logging into a controller and this is typically done from an anonymous
+        // state, even if the user is logging in on the initial load, the app
+        // execution is similar.
+        assert.equal(app.reloadConfigFile.callCount, 1);
```
hatched: Unfortunately this seems like the best we can do without going full uitest route here as it requires so many moving parts to truly test that this functions properly

hatched:
Thanks @huwshimi 

:shipit:

jujugui:

PR #3457, opened 2018-02-05
Ignore owned models in credential list without matching credential

### LINKED ISSUE #3435: New Profile: credential list fails if models credential has been removed
If a model is created with a key, then the key is removed. It is no longer able to fetch from the map and fails.

https://github.com/juju/juju-gui/blob/develop/jujugui/static/gui/src/app/components/profile/credential-list/credential-list.js#L28

This needs to validate that a credential has been returned before trying to assign a model name to it.

hatched:
Fixes #3435

### DIFF
--- a/jujugui/static/gui/src/app/components/profile/credential-list/credential-list.js
+++ b/jujugui/static/gui/src/app/components/profile/credential-list/credential-list.js
@@ -35,8 +35,14 @@ class ProfileCredentialList extends React.Component {
       const clouds = await this._listClouds();
       const credentialMap = await this._getCloudCredentialNames(props.username, clouds);
       const credentialToModel = await this._fetchAndFilterModelsByCredential();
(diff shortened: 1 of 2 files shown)


hatched:
@cassiocassio could you QA this please, with this branch you should be able to view your credential list in the new profile.

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8469/

cassiocassio:
functionality good
- shows creds, including those no longer existing on models shared from others (was this the bug?)


- can add creds
- can see which models creds are used on
- can count when 2 or more models

niggle/quibble:
- alignment of add credentials button?

2 models using creds
creds added
add creds
i have three credentials

model list showing newly added creds

hatched:
Thanks @cassiocassio if you can view your credentials then it's ok! We will address these other concerns in follow-ups.

:shipit:

jujugui:

PR #3464, opened 2018-02-06
Update terminal message for when the connection is closed unexpectedly

### LINKED ISSUE #3461: Shell pane should close automatically when the connection is closed by the server
There is a notification saying "Terminal connection unexpectedly closed", but the pane is left open and unuseful. If users want to re-establish a jujushell connection they have to close the pane and open it again, which is not a great user experience.

We could also consider to changing the message with something like "Shell connection closed by the server".

hatched:
Fixes #3461 

This is a very small first step to make the handling of terminated connections more friendly to the users. Additional updates will be coming in future releases as needed.


jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8475/

hatched:
Thanks @frankban 
:shipit:

jujugui:

Issue #3512, opened 2018-02-15
Profile login link does nothing


huwshimi:
If you visit your charms/bundles list in your profile when you're not logged into the charmstore you see the message to log in. The login button works but the "login" link in the text does nothing.

PR #3524, opened 2018-02-15
Make login and store links work

huwshimi:
Fixes: #3512.


mitechie:
QA is ok on the login link but I'm not sure the charmstore link is helpful. I'm in a message that I need to login to the charmstore but that just opens the browse page and there's on login path forward there at all. I'm not sure that the folks are looking to perform a search from that context of the charms/bundles in their profile.

hatched:
Thanks @mitechie that was the design and this is making sure the functionality is correct. We can go back and modify the design in a follow-up :shipit:

jujugui:

Issue #3533, opened 2018-02-16
TypeError when a changeState triggers a user change


frankban:
For instance, from `/u/frankban/#bundles`, expand the first mediawiki-single bundle, the. click on yellow, the yellow profile is not loaded and the console shows a `TypeError: profileUser is undefined` error.

PR #3535, opened 2018-02-16
Fix switching from one profile to another.

frankban:
Re-evaluate the profile sections on render, rather than only initially when the profile component is initialized. This way we can handle changes on the currently visualized profile in a more solid way.
Fixes https://github.com/juju/juju-gui/issues/3533

### DIFF
--- a/jujugui/static/gui/src/app/components/profile/profile.js
+++ b/jujugui/static/gui/src/app/components/profile/profile.js
@@ -17,121 +17,7 @@ const Panel = require('../panel/panel');
 
 /** Profile React component used to display user details. */
 class Profile extends React.Component {
-  constructor(props) {
-    super(props);
-    const isActiveUsersProfile = props.controllerUser.split('@')[0] === props.userInfo.profile;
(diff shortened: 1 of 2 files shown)


jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8521/

mitechie:
This does work to load that profile but since I clicked it from my own controller logged in as admin, clicking on rharding's list of charms/bundles (which I was logged into the charmstore as) the UX didn't actually change for me in a noticeable way. The username changed and I lost the models/credentials options but the list didn't move/change which is the main body. At first I thought it didn't do anything. 

I think we should look to update the gravatar and such as follow up to this to make the changes more visually different for users. 

Maybe it's worth bringing up to design changing something like a bg color or something when going from the profile of a user on the GIJOE vs from a charmstore profile.

mitechie:
👍 review and QA ok with notes.

frankban:
Thanks for the review and QA Rick!
:shipit:

jujugui:

Issue #3436, opened 2018-01-23
new profile gravatar icon should have tooltip


cassiocassio:
How to set one's (gr)avatar is unguessable in new profile.

We should provide a tooltip, with a link, to help the user know how to do this. 

1. to set their gravatar if they have none
2. to edit their gravatar, if they have one

Designs at:
https://github.com/CanonicalLtd/juju-design/blob/master/Profile/01-Header.png

Initally: link users to www.gravatar.com (which usually redirects to en.gravatar.com)

Later: we should explain they have to create a Wordpress account that uses the same email as they use for USSO. ( link to a page, or provide a pop-up? )

It seems: gravatar usernames are not related to usso usernames.

PR #3549, opened 2018-02-21
Profile header tooltips

huwshimi:
Fixes: #3436.

### DIFF
--- a/jujugui/static/gui/src/app/components/profile/header/_header.scss
+++ b/jujugui/static/gui/src/app/components/profile/header/_header.scss
@@ -19,6 +19,7 @@
     float: left;
     width: $size;
     height: $size;
+    margin-top: 7px;
     border-radius: 50%;
 
     &--hidden {
@@ -68,10 +69,13 @@
   &__username {
     color: $color-dark;
     $size: 45px;
-    margin: 0 0 12px;
-    padding: 20px 0 0 0;
-    font-size: $size;
-    line-height: 58.5px;
+    margin: 20px 0 12px;
+
+    h1 {
+      padding: 0;
+      font-size: $size;
+      line-height: 58.5px;
+    }
   }
 
   &__meta {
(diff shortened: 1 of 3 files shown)


huwshimi:
QA:
- open your profile
- hover your avatar and username
- they should show tooltips
- visit another user's profile
- hovering the avatar/username should not show tooltips

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8534/

huwshimi:
@frankban good catch. Is there any way to tell what the provider is? I think the point of the tooltip is to hint at where the details are coming from. So for local it could read "Your local account details" and for SSO it could remain the same, but as you say we'd want to be able to configure that to show the correct SSO provider.

cassiocassio:
+1 to “username at juju controller x.x.x.x” or dns name if we know it.
“Local”  means local to the Idm, not using a plugin that could be the usso
connector for Jaas or ldap who knows what else. Local is a bit overloaded -
local host etc. The tooltip should let the user where  the identity shown
ultimately resides.

frankban:
We don't have an easy way to tell the id provider. We have the `@local` vs `@external` postfix on the user name to tell at least whether the user is a local controller user or an external one, authenticated through the idm.

cassiocassio:
when adding users in the share dialog, we assume @external, and describe
the users so-added as "Ubuntu SSO user"

but we don't use @external in the juju path in the gui, and we assume it in
the share dialog if you provide a username

The idea of the tooltip for the username was to (i) help people discover
and use the gravatar functionality, and to help them remember that the JAAS
identity lives at login.ubuntu.com, which is somewhat hard to
remember/access otherwise (I know we'll send you there if you logout)

"Your Ubuntu One *account details*" -> links to login.ubuntu.com in new tab

In the local case... we *do* show @local in the GUI path, so perhaps the
tooltip could show

"Your *local username* at the juju controller" -> links to
https://jujucharms.com/docs/stable/tut-users#create-a-user

On Thu, Feb 22, 2018 at 8:40 AM, Francesco Banconi <notifications@github.com

huwshimi:
I've updated this to remove the username tooltip until such time as we know exactly how to handle all those cases.

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8561/

hatched:
:shipit:

jujugui:

Issue #3133, opened 2017-08-11
Check model name conflicts


huwshimi:
In the deployment flow when naming a model we could check the model name against the user's existing models and validate for conflicts or even append a suffix to make it unique or maybe even give some suggestions to the user.

huwshimi:
The suggested names cold be something like "kubernetes**-test**" "kubernetes**-prod**" "kubernetes**-qa**" etc.

cassiocassio:
1.

We already check in the flow if the name is duplicate, but we do so at the point of clicking "Deploy"

<img width="1680" alt="screen shot 2017-08-31 at 14 28 53" src="https://user-images.githubusercontent.com/4593556/29925625-d18641e6-8e58-11e7-9b14-425b22b8009b.png">

We check validity, capitals and spaces etc when you tab out/onward from the field.

<img width="817" alt="screen shot 2017-08-31 at 14 30 35" src="https://user-images.githubusercontent.com/4593556/29925719-25619ebe-8e59-11e7-85d8-675630d7d196.png">

Could/should we also check for namespace-unique at that moment?

2. 

I think it's going to be very hard to usefully guess what users want to name their models. 

It's more important to teach them that models are not rename-able.

3.

The right solution would be be to make models rename-able.

anthonydillon:
Also I spoke to John about the limits on the model names. They say there is
no requirement to block capitals letters. They only did it because they
thought people might get confused with closely named models. Such as,
`mymodel` and `Mymodel`.

I think we should allow the user to name there models what they like. I
came across this issue when creating models on my phone. A phone keyboard
auto sets the first letter to a capital which is not valid.

It would be nice to reduce the limits and validation.

frankban:
+1 on reducing the limits and on validating the field onChange.

PR #3567, opened 2018-03-01
Validate model name conflicts in the deployment flow

huwshimi:
Fixes: #3133.

### DIFF
--- a/jujugui/static/gui/src/app/components/deployment-flow/model-name/model-name.js
+++ b/jujugui/static/gui/src/app/components/deployment-flow/model-name/model-name.js
@@ -7,12 +7,29 @@ const React = require('react');
 const GenericInput = require('../../generic-input/generic-input');
 
 class DeploymentModelName extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      modelNames: []
+    };
+  }
 
   componentDidMount() {
     const modelName = this.refs.modelName;
     if (modelName) {
       modelName.focus();
     }
+    this._getModels();
+  }
+
+  /**
+    Get the user's models
+  */
+  _getModels() {
+    this.props.listModelsWithInfo((error, models) => {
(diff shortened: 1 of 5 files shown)


huwshimi:
QA:
- start a new model and add an app
- open the deployment flow
- enter a model name that is the same as one of your existing models and blur the field
- there should be a validation message
- enter a unique model name and blur the field
- there should be no validation message

huwshimi:
[Check for model name conflicts](https://trello.com/c/1fUIgPnM/376-check-for-model-name-conflicts)

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8553/

jujugui/static/gui/src/app/init/component-renderers-mixin.js
```diff
@@ -1017,6 +1017,7 @@ Browser: ${navigator.userAgent}`
         gtmEnabled={this.applicationConfig.GTM_enabled}
         listBudgets={this.plans.listBudgets.bind(this.plans)}
         listClouds={controllerAPI.listClouds.bind(controllerAPI)}
+        listModelsWithInfo={this._bound.listModelsWithInfo}
```
hatched: We probably don't need the model info for this.

jujugui/static/gui/src/app/components/deployment-flow/model-name/model-name.js
```diff
@@ -7,12 +7,29 @@ const React = require('react');
...
     if (modelName) {
       modelName.focus();
     }
+    this._getModels();
+  }
+
+  /**
+    Get the user's models
+  */
+  _getModels() {
```
hatched: Shouldn't this fetch the list of names on blur as well?

huwshimi:
I'm going to close this because you're probably right, but that's a bigger issue that needs to be resolved in the future.

PR #3586, opened 2018-03-15
Update deployment flow tests to enzyme

huwshimi:


huwshimi:
[Update deployment flow tests to use Enzyme](https://trello.com/c/JJjcpwPi/404-update-deployment-flow-tests-to-use-enzyme)

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8574/

jujugui/static/gui/src/app/components/deployment-flow/credential/add/add.js
```diff
@@ -166,6 +166,9 @@ class DeploymentCredentialAdd extends React.Component {
     if (!info || !info.forms) {
       return;
     }
+    if (!info.forms[this.state.authType]) {
```
hatched: Was this a bug?

huwshimi: It appeared to be, yes. I don't think we ever hit the bug due to the props we had, but I suspect we would have hit it down the track depending on the cloud configs.

huwshimi:
Thanks @hatched :shipit:

jujugui:

PR #3646, opened 2018-05-10
Fix a race generating weird app placement in the canvas

frankban:
We lived in the assumption that the deploy callbacks are executed before the mega-watcher handlers, which is not always the case.
When the mega-watcher messages arrive before the corresponding ghost applications were resurrected, then a duplication occurred, in which app names clashed and also mega-watcher entities had no x/y annotations.

### DIFF
--- a/jujugui/static/gui/src/app/d3-components/test-d3-components.js
+++ b/jujugui/static/gui/src/app/d3-components/test-d3-components.js
@@ -145,20 +145,6 @@ describe('d3-components', function() {
       assert.strictEqual(state.dbldbl, true);
     });
 
-  it('should correctly handle synthetic event bindings', function(done) {
-    comp = new Component({container: container});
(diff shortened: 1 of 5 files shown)


jujugui/static/gui/src/app/d3-components/test-d3-components.js
```diff
@@ -145,20 +145,6 @@ describe('d3-components', function() {
       assert.strictEqual(state.dbldbl, true);
     });
 
-  it('should correctly handle synthetic event bindings', function(done) {
```
frankban: This test violates isolation and generates errors in subsequent test cases randomly.

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8666/

hatched:
:shipit:

jujugui:

PR #3696, opened 2018-06-08
Migrate model tests to the new suite

huwshimi:


huwshimi:
[Migrate model tests to the new suite](https://trello.com/c/sT5VSWqP/678-migrate-model-tests-to-the-new-suite)

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8736/

jujugui/static/gui/src/app/models/test-models.js
```diff
@@ -697,7 +706,7 @@ describe('test_model.js', function() {
 
       // XXX - We no longer use relation_errors but this test should remain
       // until it's completely removed from the codebase.
-      it.skip('should reset relation_errors',
+      xit('should reset relation_errors',
```
jaycee: Have we removed `relation_errors` from the code base yet? It looks like we just set it as an empty object but it otherwise doesn't exist in anything but tests. Would be nice to go ahead and finish removing it.

Not necessarily in this branch mind you, but given we're migrating a test for something we don't need, I thought it worth bringing up.

karma-mocha-old.conf.js.tmpl
```diff
@@ -43,9 +43,6 @@ module.exports = function(config) {
       'jujugui/static/gui/src/test/test_controller_api.js',
       'jujugui/static/gui/src/test/test_env.js',
       'jujugui/static/gui/src/test/test_login.js',
-      'jujugui/static/gui/src/test/test_model_bundle.js',
-      'jujugui/static/gui/src/test/test_model_handlers.js',
-      'jujugui/static/gui/src/test/test_model.js',
```
jaycee: How do tests get loaded/excluded from the new suite? I'm curious since I just seem them being removed from old conf here but not added to the new one, so I'm assuming something automatic is happening?

hatched: It uses a few different wildcards...

```
      'jujugui/static/gui/src/app/test-*.js': ['browserify'],
      'jujugui/static/gui/src/app/d3-components/test-*.js': ['browserify'],
      'jujugui/static/gui/src/app/init/**/test-*.js': ['browserify'],
```

hatched:
:shipit:

jujugui:

jujugui/static/gui/src/app/models/test-models.js
huwshimi: I had a quick look last week and I think this will get removed when @frankban's jujulib work lands then we can remove this.

PR #3701, opened 2018-06-13
Remove old test files

huwshimi:


huwshimi:
[Remove old test files](https://trello.com/c/w2Y9P8Qy/687-remove-old-test-files)

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8742/

frankban:
It would be nice if you provide some context about why those files can be safely removed, why they are no longer used in tests, what's used instead.  Otherwise this looks great thanks.

huwshimi:
@frankban ah yep sorry, these are all files from the old test suite that were not being used any more. Old test assets, test setup and test data. Most of probably hasn't been used in a long long time.

PR #3711, opened 2018-06-19
Only show ssh buttons under a flag and make terminal state more resilient.

### LINKED ISSUE #3706: jujushell state change requires array as value
Clicking the button fails with an error in the console as we are passing true to state instead of an array of commands. Changing it to an empty array fixes the error but the terminal does not open, this time for a possible issue in the falsey value as detected by state.

### LINKED ISSUE #3708: Put new jujushell interaction behind the expert flag
New extensions are:
- ssh to machine in the machine view;
- ssh to unit in the unit detail page in the inspector;
- tail logs and debug hooks in the unit in error in the inspector.

hatched:
Fixes #3706
Fixes #3708

### DIFF
--- a/jujugui/static/gui/src/app/components/inspector/inspector.js
+++ b/jujugui/static/gui/src/app/components/inspector/inspector.js
@@ -152,7 +152,7 @@ class Inspector extends React.Component {
               destroyUnits={nextProps.destroyUnits}
               previousComponent={previousComponent}
               service={service}
-              showSSHButtons={nextProps.showSSHButtons}
+              showSSHButtons={nextProps.showSSHButtons && window.juju_config.flags.expert}
               unit={unit}
               unitStatus={unitStatus} />,
           backState: {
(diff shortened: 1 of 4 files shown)


hatched:
#### To QA

Switch to a model and open the machine view, you should not see any 'ssh-to-machine' buttons.
Enable the `expert` flag and you should then see them.

Click the Terminal button on the top of the canvas once viewing a model. It should open the terminal as expected.

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8752/

jujugui/static/gui/src/app/components/inspector/inspector.js
```diff
@@ -152,7 +152,7 @@ class Inspector extends React.Component {
               destroyUnits={nextProps.destroyUnits}
               previousComponent={previousComponent}
               service={service}
-              showSSHButtons={nextProps.showSSHButtons}
+              showSSHButtons={nextProps.showSSHButtons && window.juju_config.flags.expert}
```
huwshimi: I think we normally try and avoid accessing these globals inside the components, but if this is temporary I guess it should be OK.

hatched:
Thanks @huwshimi 

:shipit:

PR #3741, opened 2018-07-11
Surface errors when deleting models

### LINKED ISSUE #3732: Errors destroying models aren't surfaced to user
On staging, there's an intermittent bug with juju unrelated to the GUI that prevents destroying a model:

```
{"type":"ModelManager","request":"DestroyModels","params":{"models":[{"model-tag":"model-31fea4e9-36eb-41c7-82f3-0cae8e6e259b"}]},"version":4,"request-id":8}
{"request-id":8,"response":{"results":[{"error":{"message":"this Juju controller requires destroyStorage to be true","code":""}}]}}
```

When destroying this model in the GUI, no message is shown to the user, and it should be.

makyo:
Fixes #3732

### DIFF
--- a/jujugui/static/gui/src/app/components/profile/model-list/model-list.js
+++ b/jujugui/static/gui/src/app/components/profile/model-list/model-list.js
@@ -74,9 +74,18 @@ class ProfileModelList extends React.Component {
     });
   }
 
+  /**
+    Destroys a model once confirmed by the user.
+    @param {Object} string The UUID of the model to destroy.
+  */
   _confirmDestroy(modelUUID) {
     this.setState({notification: null});
     this.props.destroyModel(modelUUID, (error, data) => {
+      // There are two possible error states for this call:
+      // An error can be returned from the API call itself, in which case error
+      // is populated or there can be errors in destroying individual models,
+      // in which case data will have errors. Potentially, both will have errors
+      // as well. We will display all errors returned by the API call.
       if (error) {
         this.props.addNotification({
           title: 'Error destroying model',
@@ -84,7 +93,7 @@ class ProfileModelList extends React.Component {
           level: 'error'
         });
       }
-      if (data != {}) {
+      if (Object.keys(data).length) {
         for (let key in data) {
           this.props.addNotification({
             title: 'Error destroying model',
--- a/jujugui/static/gui/src/app/components/profile/model-list/model-list.js
+++ b/jujugui/static/gui/src/app/components/profile/model-list/model-list.js
@@ -90,7 +90,7 @@ class ProfileModelList extends React.Component {
             title: 'Error destroying model',
             message: `Could not destroy model: ${data[key]}`,
             level: 'error'
-          })
+          });
         }
       }
       this._fetchModels(this.props.facadesExist);
--- a/jujugui/static/gui/src/app/components/profile/model-list/model-list.js
+++ b/jujugui/static/gui/src/app/components/profile/model-list/model-list.js
@@ -80,27 +80,15 @@ class ProfileModelList extends React.Component {
   */
   _confirmDestroy(modelUUID) {
     this.setState({notification: null});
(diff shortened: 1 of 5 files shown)


jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8909/

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8910/

jujugui/static/gui/src/app/components/profile/model-list/model-list.js
```diff
@@ -74,6 +74,29 @@ class ProfileModelList extends React.Component {
     });
   }
 
+  _confirmDestroy(modelUUID) {
```
hatched: docblock missing.

jujugui/static/gui/src/app/components/profile/model-list/model-list.js
```diff
@@ -74,6 +74,29 @@ class ProfileModelList extends React.Component {
...
+    this.setState({notification: null});
+    this.props.destroyModel(modelUUID, (error, data) => {
+      if (error) {
+        this.props.addNotification({
+          title: 'Error destroying model',
+          message: error,
+          level: 'error'
+        });
```
hatched: `return;` after this msg else you could give two notifications for the same error....or maybe not....see comment below about multiple error locations.

jujugui/static/gui/src/app/components/profile/model-list/model-list.js
```diff
@@ -74,6 +74,29 @@ class ProfileModelList extends React.Component {
...
+      }
+      if (data != {}) {
```
hatched: I believe here you're checking to see if the data object returned is not empty. I'm a little confused here as to why an error destroying model wouldn't be returned as an error above. If the `destroyModel` call returns error values in the data then maybe we need to document that, or fix it in the api.

In any event, if you want to check if an object is empty you should use something like `if (Object.keys(data).length) {` as it'll only report keys for enumerable properties.

makyo: You can have an error from the API call, or models can return an error on destroy (remember that the actual call is destroyModels and takes more than one; the object is a map of model uuids to responses), or potentially both, so we need to display all errors. I'll comment to that effect.

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8911/

makyo:
# QA

Deploy http://localhost:8042/canonical-kubernetes/bundle/76

Attempt to destroy the model once deployment is finished. Ensure that you get an error about storage.

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8913/

jujugui/static/gui/src/app/store/env/controller-api.js
```diff
@@ -856,15 +856,21 @@ window.yui.add('juju-controller-api', function(Y) {
...
+          if (result.error) {
+            prev[id] = result.error.message;
+            errors.push(result.error.message);
+          } else {
+            prev[id] = null;
+          }
           return prev;
         }, {});
-        callback(null, results);
+        callback(errors, results);
```
hatched: `errors` here would be truthy (`[]` is truthy) so I'd prefer it to stay as null so consumers of the api could still go `if (!error) { ... }`

What do you think about: 
`callback(errors.length ? errors : null, results);`

makyo: This is such a ridiculously small win :P If you say so...

hatched: It's just about creating a consistent api for those who are consuming it.

jujugui:
http://ci-gce.jujugui.org:8080//job/juju-gui/8914/

hatched:
Thanks @makyo :shipit:

jujugui:

PR #3762, opened 2018-08-02
Bundle export module

huwshimi:
Move the model exporting code outside the models (this code will need to be retained when we replace the db).


jujugui/static/gui/src/app/init/bundle-exporter.js
```diff
@@ -0,0 +1,351 @@
...
+    this.db = cfg.db;
+  }
+
+  /**
+    Generate a service list for the exported yaml file based on the list of
+    services passed in.
+    @param {Object} serviceList The service list.
+    @return {Object} The services list for the export.
+  */
+  _generateServiceList(serviceList) {
```
huwshimi: These methods are pretty much just copied over from the models file.

jujugui/static/gui/src/app/init/utils.js
```diff
@@ -105,6 +107,7 @@ utils.exportEnvironmentFile = db => {
   @param {Date} date object
 */
 utils._generateBundleExportFileName = (envName, date=new Date()) => {
+  envName = envName || 'untitled-model';
```
huwshimi: This is a fix for the generated filename that was introduced when we changed the new model name to be blank.

huwshimi:
[Bundle export module](https://trello.com/c/6655uc7Y/839-bundle-export-module)

huwshimi:
QA:
- start a new model
- add a bundle
- click export in the header
- the file should export correctly

PR #3764, opened 2018-08-03
Remove link colour override

huwshimi:
There was a temporary override to Vanilla link colours but this branch removes that so we can just use the Vanilla styles.


huwshimi:
QA:
- have a poke around the GUI
- the link colours should be as before (look for obviously out of place blue links)

huwshimi:
[Remove link colour override (use Vanilla defaults)](https://trello.com/c/ID1h4Sa2/843-remove-link-colour-override-use-vanilla-defaults)

Issue #3659, opened 2018-05-18
Destroy model silently fails with external storage


cassiocassio:
Destroying a model in the GUI silently fails

Clicking on the < trash> icon: 
<img width="848" alt="screen shot 2018-05-18 at 11 03 27" src="https://user-images.githubusercontent.com/4593556/40229237-8f787fd0-5a8b-11e8-90f1-0120f3f26522.png">

... leads to the confirm dialog:

<img width="430" alt="screen shot 2018-05-18 at 11 03 24" src="https://user-images.githubusercontent.com/4593556/40229259-a42dc5ca-5a8b-11e8-9236-f2980f3e9643.png">

Clicking <Destroy>

--

With @mhilton - Initially I thought this was when Juju controller requires destroyStorage to be true:

```
cassio@benbecula-2:~ $ juju destroy-model testk8s
WARNING! This command will destroy the "testk8s" model.
This includes all machines, applications, data and other resources.

Continue [y/N]? y
Destroying model
ERROR failed to destroy model "testk8s"
ERROR cannot destroy model: this Juju controller requires destroyStorage to be true
cassio@benbecula-2:~ $ juju destroy-model testk8s --destroy-storage
WARNING! This command will destroy the "testk8s" model.
This includes all machines, applications, data and other resources.

Continue [y/N]? y
Destroying model
```

but I fear also seems to be a problem in the general case. This needs further investigation.

**Proposals: (in the case of the storage failure)**

1.  Improve the error language for the juju CLI

'ERROR cannot destroy model: use --destroy-storage option to force.

2. If attached storage is also destroyed, make it clear in the dialog.

![destroy all](https://user-images.githubusercontent.com/4593556/40230615-0ab8356a-5a90-11e8-8f22-f7bdc5a83c76.png)

frankban:
Isn't this also a problem in current API client system always selecting most recent facade version?

hatched:
I think it's a combination of two things. We always select the latest facade version, and that we aren't sending that configuration option when destroying. We'll need to revisit this once the new jujulib stuff has been implemented.

PR #3769, opened 2018-08-08
Work around the destroy storage error

frankban:
When using a new client (jimm) with an old server (like staging) the client returned that error when the deploy-storage key is missing or false in the API request.
See https://github.com/juju/juju/blob/develop/api/modelmanager/modelmanager.go#L390
Fixes https://github.com/juju/juju-gui/issues/3659 (for now)
We will improve this later, when using the new jujulib.

### DIFF
--- a/jujugui/static/gui/src/app/store/env/controller-api.js
+++ b/jujugui/static/gui/src/app/store/env/controller-api.js
@@ -883,7 +883,12 @@ window.yui.add('juju-controller-api', function(Y) {
         tagKey = 'tag';
       }
       const entities = ids.map(function(id) {
-        return {[tagKey]: tags.build(tags.MODEL, id)};
+        return {
+          [tagKey]: tags.build(tags.MODEL, id),
+          // TODO(frankban): allow for selecting not to destroy storage.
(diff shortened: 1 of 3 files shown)


frankban:
QA: point guiproxy to this branch on staging, destroy a model, the confirmation dialog should mention that also storage will be destroyed, and then it should work.

jujugui/static/gui/src/app/store/env/controller-api.js
```diff
@@ -883,7 +883,12 @@ window.yui.add('juju-controller-api', function(Y) {
         tagKey = 'tag';
       }
       const entities = ids.map(function(id) {
-        return {[tagKey]: tags.build(tags.MODEL, id)};
+        return {
+          [tagKey]: tags.build(tags.MODEL, id),
+          // TODO(frankban): allow for selecting not to destroy storage.
+          // This will be available when switching to jujulib.
```
hatched: Issue created to track this: https://github.com/juju/juju-gui/issues/3770

hatched:
Thanks @frankban

PR #3778, opened 2018-08-21
Set up Jest

huwshimi:


jujugui/static/gui/src/app/components/help/test-help.js
```diff
@@ -22,132 +20,7 @@ describe('Help', () => {
...
-                target="_blank">
-                The Juju project
-              </a>
-            </p>
-          </div>
-        </footer>
-      </div>
-    );
-    assert.compareJSX(wrapper.find('.help__body'), expected);
+    expect(wrapper).toMatchSnapshot();
```
huwshimi: I've updated this one component as an example.

frankban: Nice

barrymcgee: 💥🙌🏻💪🏻

.babelrc
```diff
@@ -13,6 +13,15 @@
       "plugins": [
         "rewire"
```
hatched: Do we still need rewire in development if it's in the test?

Makefile
```diff
@@ -293,6 +295,18 @@ test-js-old: gui
...
 
+.PHONY: test-jest
+test-jest: gui
+	$(JEST)
+
+.PHONY: watch-jest
+watch-jest: gui
+	$(JEST) --watchAll
+
+.PHONY: watch-jest
```
hatched: This phony label is incorrect.

jujugui/static/gui/src/app/components/help/test-help.js
hatched: 😆

package.json
```diff
@@ -64,10 +68,12 @@
     "keysim": "2.0.1",
     "mkdirp": "0.5.1",
     "mocha": "5.1.1",
+    "mutation-observer": "^1.0.3",
```
hatched: I see a number of these are being added but not used? Can you give me an idea of why these are required?

huwshimi:
[Set up Jest](https://trello.com/c/SLw5xjIy/872-set-up-jest)

.babelrc
huwshimi: Unfortunately we rewire in files that are tested by Karma and Jest and Karma uses "development" and Jest uses "test".

jujugui/static/gui/src/test/jest-setup.js
```diff
@@ -0,0 +1,6 @@
+'use strict';
+
+global.assert = require('chai').assert;
+global.sinon = require('sinon');
+global.enzyme = require('enzyme');
+global.MutationObserver = require('mutation-observer');
```
huwshimi: This is where the mutation observer gets attached.

package.json
huwshimi: This gets attached to the global test object as jsdom doesn't have a MutationObserver implementation by default. I've added a comment to this PR for where this happens.

huwshimi:
Thanks everyone :shipit:

huwshimi:
Oops, not the squirrel!

PR #3784, opened 2018-09-04
Send analytics ping when exporting bundle.

hatched:


jujugui/static/gui/src/app/init/utils.js
```diff
@@ -76,8 +76,10 @@ utils.unloadWindow = function() {
   Export the YAML for the current model, including uncommitted changes.
 
   @param {Object} db The application database.
+  @param {Function} [sendAnalytics] Function to send analytics about the use of
+    the bundle export.
 */
-utils.exportEnvironmentFile = db => {
+utils.exportEnvironmentFile = (db, sendAnalytics)=> {
```
hatched: This fn has no tests, and should have them. It uses so many globals that makes it quite difficult to properly test. Because we may remove this in favour of the Juju export I didn't feel it valuable to do that work at this time.

jujugui/static/gui/src/app/init/component-renderers-mixin.js
```diff
@@ -129,7 +129,7 @@ const ComponentRenderersMixin = superclass => class extends superclass {
         changeState={this._bound.changeState}
         displayTerminalButton={this._enableTerminal()}
         exportEnvironmentFile={
-          initUtils.exportEnvironmentFile.bind(initUtils, this.db)}
+          initUtils.exportEnvironmentFile.bind(initUtils, this.db, this.sendAnalytics)}
```
frankban: Is sendAnalytics already bound?

hatched: Everything it needs is either a global or passed in on call so it doesn't need to be bound.

hatched:
Thanks!

PR #3787, opened 2018-09-07
Update ProfileNavigation to Vanilla v1 markup

barrymcgee:


barrymcgee:
@hatched I've updated the `.eslintrc` rules in this change to include;

`"react/jsx-max-props-per-line": [1, { "when": "always" }],`

The reason for this is props alignment is currently largely inconsistent across the codebase and Prettier was making changes as you highlighted in a previous PR.

By default, Prettier will allow inline props until the set line width (80 chars) before breaking them onto separate lines. 

This will result in mixed cases where some components have props inline whereas others components have props multi-line. This has [proved controversial](https://github.com/prettier/prettier/issues/3101).

However, as with our setup, Eslint trumps Prettier so we can add the above rule for consistency across the board. You will, however, see regular corrections as I/we touch different files and the consistency is enforced.

jujugui/static/gui/src/app/components/profile/navigation/_navigation.scss
```diff
@@ -1,30 +1,30 @@
...
     cursor: pointer;
+    list-style: none;
 
     &:nth-child(4) {
-      border-top: 1px solid #cdcdcd;
-      padding-top: 12px;
-      margin-right: 24px;
+      border-top: 1px solid $color-mid-light;
+      margin-top: 1rem;
+      padding-top: 1rem !important;
```
anthonydillon: I think adding an element to the list such as: `<li class="p-list__item"><hr class="u-no-margin--bottom"></li>` is clearer then tarenting `nth-child` in css.

jujugui/static/gui/src/app/components/profile/navigation/_navigation.scss
```diff
@@ -1,30 +1,30 @@
 .profile-navigation {
   color: $color-dark;
+  padding: 0 1rem;
 
-  &__list-item {
-    list-style: none;
+  .p-list__item {
     cursor: pointer;
+    list-style: none;
```
anthonydillon: This should be inherited via `p-list`

jujugui/static/gui/src/app/components/profile/navigation/_navigation.scss
barrymcgee: This was my preferred approach also but as these links are generated from a section map object, that would involve splitting the section map and populating two different lists etc which is, in my opinion, adding too much complexity just to accommodate a stylistic keyline. 

However, relying on `nth:child` is rather brittle so as an improvement I have instead added a unique class to each list item and refactored the SCSS to target the `Cloud Credentials` item instead.

jujugui/static/gui/src/app/components/profile/profile.js
```diff
@@ -164,19 +165,15 @@ class Profile extends React.Component {
       sectionsMap.set('invoices', {
         label: 'Invoices',
         getComponent: () => {
-          return (
-            <ProfileInvoiceList
-              baseURL={props.baseURL}
-              user={props.userInfo.external} />);
+          return (<ProfileInvoiceList baseURL={props.baseURL}
```
hatched: I think the old layout for this was easier to read

jujugui/static/gui/src/app/components/profile/profile.js
```diff
@@ -164,19 +165,15 @@ class Profile extends React.Component {
...
         }
       });
       sectionsMap.set('revenue-statement', {
         label: 'Revenue Statements',
         getComponent: () => {
-          return (
-            <ProfileInvoiceList
-              baseURL={props.baseURL}
-              user={props.userInfo.external} />);
+          return (<ProfileInvoiceList baseURL={props.baseURL}
```
hatched: here as well.

jujugui/static/gui/src/app/components/profile/profile.js
```diff
@@ -191,28 +188,25 @@ class Profile extends React.Component {
...
-          <div className="twelve-col">
-            <div className="profile__content inner-wrapper">
-              <ProfileNavigation
-                // Use supplied activeSection or the key from the first map entry.
-                activeSection={this._getSectionInfo().active || mapEntry[0]}
-                changeState={props.changeState}
-                sectionsMap={sectionsMap} />
-              {section.getComponent()}
-            </div>
+      <Panel instanceName="profile"
```
hatched: The inconsistency where sometimes an attribute it on the same line as the tag name and sometimes not actually makes things harder to read.

.eslintrc
```diff
@@ -43,6 +43,7 @@
     "react/jsx-tag-spacing": ["error", { "beforeSelfClosing": "always" }],
     "react/jsx-uses-react": 2,
     "react/jsx-uses-vars": 2,
+    "react/jsx-max-props-per-line": [1, { "when": "always" }],
```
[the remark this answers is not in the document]
huwshimi: I'm OK with 1 per line. We do that in probably 90+% of cases at the moment.

jujugui/static/gui/src/app/components/profile/navigation/navigation.js
```diff
@@ -24,29 +23,28 @@ class ProfileNavigation extends React.Component {
...
+      );
     });
 
     return (
-      <div className="profile-navigation">
-        <ul>
-          {links}
-        </ul>
-      </div>);
+      <div className="profile-navigation v1">
```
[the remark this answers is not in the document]
barrymcgee: This is how we namespace new Vanilla.

jujugui/static/gui/src/app/components/profile/navigation/_navigation.scss
```diff
@@ -1,30 +1,29 @@
...
-  &__list-item {
-    list-style: none;
+  .p-list__item {
     cursor: pointer;
 
-    &:nth-child(4) {
-      border-top: 1px solid #cdcdcd;
-      padding-top: 12px;
-      margin-right: 24px;
+    .profile__content &.cloud-credentials {
```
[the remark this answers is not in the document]
barrymcgee: The extra class is needed for adequate specificity otherwise this rule is overwritten - this is because adding the `v1` namespace gives the default Vanilla styling added clout. 

The alternative is an `!important` - while not idea, I consider this approach the lesser of two evils.

jujugui/static/gui/src/app/components/profile/profile.js
barrymcgee: Perhaps but bear in mind if we make the max props 0 - you'll also get;

```
<div
	className="foo">
Baz
</div>
```

I think it looks a little off on first glance because you're not used to it - give it a week and you'll forget it was ever anything else.

hatched: looking at the eslint rule it looks like it should have allowed the original layout so I'm not sure why the change.

barrymcgee: @hatched Prettier will run and inline all props/attributes until the character limit before breaking onto a new line - then it will pass off to Eslint where the rule above will kick in and it'll make the change above.

PR #3788, opened 2018-09-07
Remove inline copyright notices and update LICENCE date

barrymcgee:


barrymcgee:
These notices are rarely updated when the respective files are so most are outdated. Timestamps for when a file has last been updated is provided by Git. There is also a dedicated `LICENCE` file at the root.

hatched:
As discussed offline I'm going to close this for now, we can revisit opening it in the future.

barrymcgee:
@hatched Can we keep this open please until I hear back from legal please? Isn't that what you asked me to do?

PR #3796, opened 2018-09-13
Align body and keylines in Profile and use snapshot tests

barrymcgee:


hatched:
There are so many changes here for changes sake that I can't even tell what has actually changed. Can you please turn off prettier as this is a tremendous waste of time to review.

barrymcgee:
@hatched I've split out indentation tweaks to a single commit and then added the subsequent actual changes on a one-commit-per-component basis. 

Ignoring the snapshots and the actual changes are very minimal.

PR #3800, opened 2018-09-14
Fix disabling import when in read only mode. Fix bundle exporting.

frankban:


jujugui/static/gui/src/app/components/model-actions/model-actions.js
```diff
@@ -129,7 +129,7 @@ class ModelActions extends React.Component {
             </span>
           </span>
           <span className="model-actions__import model-actions__button"
-            onClick={!isReadOnly && this._handleImportClick.bind(this)}
+            onClick={isReadOnly ? undefined : this._handleImportClick.bind(this)}
```
hatched: It seems a little odd that this logic is in here. I feel like it might be better placed in the handler function itself.

PR #3801, opened 2018-09-17
Split out row and cell components from the basic table

huwshimi:


huwshimi:
QA:
- open the status and your profile
- the tables should be unchanged

huwshimi:
[Split out row and cell components from the basic table](https://trello.com/c/Reln3DMF/938-split-out-row-and-cell-components-from-the-basic-table)

PR #3802, opened 2018-09-17
Update the basic table to expand content without ExpandingRow

huwshimi:


huwshimi:
[Update the basic table to expand content without ExpandingRow](https://trello.com/c/Iwz0Xi8s/939-update-the-basic-table-to-expand-content-without-expandingrow)

PR #3806, opened 2018-09-18
Make a presentation only (shareable) status wrapper component

huwshimi:


huwshimi:
[Make a presentation only (shareable) status wrapper component](https://trello.com/c/czEVGsA2/942-make-a-presentation-only-shareable-status-wrapper-component)

jujugui/static/gui/src/app/components/shared/status/status/status.js
```diff
@@ -0,0 +1,244 @@
...
+class Status extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      highestStatus: utils.STATUSES.OK,
+      statusFilter: null
+    };
+  }
+
+  componentWillReceiveProps() {
```
hatched: This is deprecated.

jujugui/static/gui/src/app/components/shared/status/status/status.js
```diff
@@ -0,0 +1,244 @@
...
+        applications={this.props.entities.applications}
+        generateApplicationURL={this.props.generateApplicationURL}
+        onApplicationClick={this.props.navigateToApplication}
+        relations={relations}
+        statusFilter={this.state.statusFilter} />);
+  }
+
+  render() {
+    let content;
+    if (!this.props.entities.model.modelUUID) {
```
hatched: This should check that model is defined as it may not be if we aren't connected.

jujugui/static/gui/src/app/components/status/__snapshots__/test-status.js.snap
```diff
@@ -1,139 +1,41 @@
...
-          "region": "neutral zone",
-          "sla": "advanced",
-          "version": "2.42.47",
-        }
-      }
-      statusFilter={null}
-    />
-  </div>
-</Panel>
+exports[`Status renders 1`] = `
```
hatched: This doesn't really test much

PR #3809, opened 2018-09-20
Deploy with pay flag but no pay

hatched:
If the pay flag is enabled but the remanning values aren't set up, still allow deployment.


jujugui/static/gui/src/app/components/deployment-flow/deployment-flow.js
```diff
@@ -1189,7 +1190,7 @@ class DeploymentFlow extends React.Component {
       return false;
     }
     // Can't deploy if there is no user.
```
frankban: Expand the comment?

frankban:
QA ok.

PR #3812, opened 2018-09-21
If a charm cannot be found in the charm store, do not fail when trying to read it from the endpoints map.

hatched:


hatched:
The tests were not updated because there are no tests to update 🤦‍♂️  This code will be removed with the new datastore shortly.

jaycee:
QA OK

jujugui/static/gui/src/app/init/endpoint-utils.js
```diff
@@ -90,6 +90,15 @@ const getEndpoints = (application, controller) => {
     // application is deployed. The GUI only needs to concern itself with
     // client/server relations.
     if (targetId === appId) { return; }
+    // XXX hatch At the time of writing...When the inspector renders we generate
```
jaycee: Please throw a date on this XXX

PR #3816, opened 2018-09-26
Shared panel component

huwshimi:


huwshimi:
QA, really this should work if CI passes, but to be sure:
- load the gui
- open the status, profile, deployment flow
- they should all appear unchanged

huwshimi:
[Shared panel component](https://trello.com/c/UVDCixUb/947-shared-panel-component)

package-lock.json
```diff
@@ -11127,6 +11127,7 @@
               "version": "1.1.11",
               "bundled": true,
               "dev": true,
+              "optional": true,
```
hatched: Ahh the good ol `"optional": true,` changes every PR....

PR #3819, opened 2018-09-27
Export bundles with series information for applications.

hatched:
When a model has a default series, say `bionic` and it contains a number of multi-series applications, on export those applications wouldn't report the series they were deployed as, so sometimes you'd end up with a machine and application being `xenial` but when re-importing that bundle it would try and deploy `bionic` application on a `xenial` machine. This branch makes it explicit what the series of the application is without inferring from the machine.


hatched:
#### To QA

- Deploy Canonical Distribution of Kubernetes and then export the bundle.
- The bundle yaml should now contain a `series: xenial` for all of the applications.
- In a new model, re-import and deploy the bundle.
- It should work as expected.

PR #3820, opened 2018-09-28
Shared button components

huwshimi:


huwshimi:
QA:
- load the gui and have a cursory look around, the buttons should be unchanaged

huwshimi:
[Shared button components](https://trello.com/c/q1f4pwri/960-shared-button-components)

PR #3829, opened 2018-10-03
Update GUI dependencies.

hatched:
Update the remaining possible dependencies and resolve any test issues.


jujugui/static/gui/src/app/components/shared/basic-table/row/__snapshots__/test-row.js.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`BasicTableRow can display rows with expandable content 1`] = `
 <li
-  className="twelve-col  basic-table__row basic-table__row--expandable basic-table__row--clickable"
```
hatched: These were from updating the `classNames` package

karma.conf.js.tmpl
```diff
@@ -94,6 +94,12 @@ module.exports = function(config) {
 
     // Continuous Integration mode
     // if true, Karma captures browsers, runs the tests and exits
-    singleRun: false
+    singleRun: false,
+
+    client: {
```
hatched: The default for jasmine now is to run the tests in a random order. This makes sense however, we we know, the `environment.js` tests are a mess and randomly fail. These tests will all be being deleted with the jujulib change so at that time we'll want to remove this.

package.json
```diff
@@ -9,32 +9,32 @@
     "url": "https://github.com/juju/juju-gui.git"
   },
   "dependencies": {
-    "classnames": "2.2.5",
-    "clipboard": "1.7.1",
+    "classnames": "^2.2.6",
+    "clipboard": "^1.7.1",
```
hatched: This is not the latest version of clipboard. The latest version 2.x does not actually function because it tries to access a file which doesn't exist in its package. 

re: https://github.com/zenorocha/clipboard.js/issues/558

karma.conf.js.tmpl
```diff
@@ -94,6 +94,12 @@ module.exports = function(config) {
...
+      jasmine: {
+        random: false
```
frankban: Please add a TODO and a bug to re-enable this when possible, also explaining why it's not possible for now (basically your review comment).

package.json
frankban: Didn't we decide for not using ^ in deps?

hatched: Yeah I can go through and remove those, however npm now installs from the package-lock so even though this has ^, it'll still only install the version specified in the lock.

karma.conf.js.tmpl
hatched: done

package.json
huwshimi: I've never seen it respect the package lock, but maybe that's a new thing...

hatched:
Thanks @huwshimi

hatched:
thanks @frankban

PR #3833, opened 2018-10-09
Add parcel building

huwshimi:


huwshimi:
The remaining bits to do here are for building the dist and getting it working with gij/gisf.

.babelrc
```diff
@@ -1,5 +1,8 @@
 {
-  "presets": [],
+  "presets": [
+    "es2015",
```
hatched: Can we try and isolate the issues requiring this tranpilation to be required? 1. So that we can keep an eye on what needs to be updated in their codebase so that we can remove these and 2. So we can remove any that aren't necessary.

.babelrc
```diff
@@ -10,13 +13,14 @@
...
-        "rewire"
-      ]
-  },
+      "plugins": []
+    },
     "test": {
       "plugins": [
-        "rewire"
+        "rewire",
+        ["babel-plugin-transform-require-ignore", {
```
hatched: Why is this required for tests but not for other builds?

Makefile
```diff
@@ -102,14 +105,14 @@ qa-server: gui
 	bin/pserve qa.ini
 
 .PHONY: run
-run: gui
+run: gui-deps
 	@echo
```
hatched: I'm not sure this message makes sense here any longer?

Makefile
```diff
@@ -102,14 +105,14 @@ qa-server: gui
 	bin/pserve qa.ini
 
 .PHONY: run
-run: gui
+run: gui-deps
```
hatched: This isn't really building any deps, it's just starting the parcel dev server?

jujugui/static/gui/src/app/assets/js/config.js
```diff
@@ -0,0 +1,23 @@
+module.exports = {
```
hatched: should this be `staging-config.js` ? or the like?

jujugui/templates/index.html
```diff
@@ -0,0 +1,218 @@
...
+    <script type="text/javascript">
+      var _gaq = _gaq || [];
+      window._gaq = _gaq;
+      // Do not call _setAccount as we need to get the key from
+      // the config file which is not yet available.
+    </script>
+  </head>
+  <body>
+    <div style="display: none;" id="svg-sprite"></div>
+    <script src="../static/gui/src/app/assets/js/svg.js"></script>
```
hatched: Should we instead generate this template with this in it? Especially now that we have 3 templates :)

jujugui/static/gui/src/app/components/store/store.js
```diff
@@ -153,7 +155,7 @@ class Store extends React.Component {
           </div>
         </div>
         <div>
-          <img src={this._generateLocalImagePath('write-your-own.png')} />
+          <img src={images['write-your-own.png']} />
```
hatched: I'm curious what this gets generated as? Is it just the path on disk relative for the parcel root?

jujugui/static/gui/src/app/components/svg-icon/svg-icon.js
```diff
@@ -71,7 +71,8 @@ class SvgIcon extends React.PureComponent {
       <svg
         className={this._generateClasses()}
         style={this._generateStyles()}
-        viewBox={this._generateViewbox()}>
+        viewBox={this._generateViewbox()}
+        xlinkHref="http://www.w3.org/1999/xlink">
```
hatched: Why are these now required?

jujugui/static/gui/src/app/init.js
```diff
@@ -1658,6 +1660,10 @@ class GUIApp {
     // Fetch the url for the config file from the DOM. This is necessary because
     // the config path is defined on render by the various deployment environments.
     const oldScript = doc.querySelector('script[data=config]');
+    if (!oldScript) {
+      // We must be using Parcel so no need to do anything.
```
hatched: Is this true? Don't we still need to get the config values and refresh them to merge in changes supplied by the user?

jujugui/static/gui/src/test/proxy-module.js
```diff
@@ -0,0 +1,4 @@
+'use strict';
+
+// This module is used as a proxy for css modules in the jest tests.
```
hatched: Is this still required with the transform in the babelrc? 
Also, why shouldn't they be loaded?

.babelrc
huwshimi: This is for the regeneratorRuntime polyfill.

.babelrc
huwshimi: Parcel handles the scss/image imports but babel doesn't know how to do that (for the karma test suites). We also don't need those to be loaded for the tests so we ignore them.

Makefile
huwshimi: You do still need to point guiproxy at the parcel runner.

jujugui/static/gui/src/test/proxy-module.js
huwshimi: We don't do anything with the css that is imported so we ignore it (rather than adding a lib to handle it). You're right this is no longer required.

.babelrc
hatched: Ahh ok, this is how I resolved it in the past: 

`parcel-bundler/parcel/issues/916#issuecomment-419154118`

Add this to the package.json
```json
  "browserslist": [
    "last 1 Chrome version"
  ]
```

huwshimi: This does work for running parcel, but parcel build now dies with `Maximum call stack size exceeded` in `...node_modules/parcel-bundler/src/utils/objectHash.js`. It must be causing something to be unhappy. No idea where though.

PR #3843, opened 2018-10-24
Increase width of owner col in models table

barrymcgee:


barrymcgee:
@hatched I've increased the width of the owner column in Models table but obviously, usernames are variable as is the width of this table as screen-width decreases so there may be still some usernames that break onto two lines. This is fine as the username is still easily readable.

PR #3844, opened 2018-10-24
Over override styles for destroy model PopUp

barrymcgee:


barrymcgee:
Destroy dialog should now look like:

screenshot 2018-10-24 at 12 42 05

PR #3845, opened 2018-10-24
Fix spacing around credentials action dropdown

barrymcgee:


barrymcgee:
Dropdown should now look like:

screenshot 2018-10-24 at 12 53 55

PR #3849, opened 2018-10-25
Due to recent changes in React this technique to scroll to an element no longer works.

### LINKED ISSUE #3848: Cannot view machine view if machines exist
If you view the machine view when you have deployed machines there is a js error which breaks the app.

hatched:
Fixes #3848

### DIFF
--- a/jujugui/static/gui/src/app/components/machine-view/machine-view.js
+++ b/jujugui/static/gui/src/app/components/machine-view/machine-view.js
@@ -233,7 +233,7 @@ class MachineView extends React.Component {
       is selected.
   */
   selectMachine(id) {
-    if (!id || id === null) {
+    if (!id) {
       return;
     }
     this.props.changeState({gui: {machines: id}});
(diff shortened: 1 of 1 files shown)


hatched:
#### To QA

Open a model with machines and then switch to the machine view.

jujugui/static/gui/src/app/components/machine-view/machine-view.js
```diff
@@ -229,42 +226,16 @@ class MachineView extends React.Component {
...
     Handle selecting a machine.
 
     @param {String} id The machine id to select. If the id is null, no machine
       is selected.
-    @param {Boolean} scrollToMachine Whether to scroll to the selected machine.
   */
-  selectMachine(id, scrollToMachine=true) {
-    if (id === null) {
+  selectMachine(id) {
+    if (!id || id === null) {
```
frankban: null evals to false so the second part of the OR is redundant.

hatched: You are correct!

frankban:
QA is ok.
Looks good with one suggestion.

Issue #3852, opened 2018-10-26
ports list is duplicated... sometimes


buggtb:
Ports list in the juju status page is a bit dodgy

<img width="1623" alt="screenshot 2018-10-26 at 13 29 19" src="https://user-images.githubusercontent.com/103544/47577159-24004f80-d93e-11e8-9263-d5bc9f96350a.png">

hatched:
While I can't reproduce exactly what is there I was able to reproduce the duplication easily by just deploying and exposing `apache-drill`.

![screen shot 2018-10-26 at 10 16 54 am](https://user-images.githubusercontent.com/532033/47579105-568f5580-d908-11e8-8e5f-ff173d305afd.png)

hatched:
The data from the parseUnit in maraca is returning the proper value but the data in the valueStore is incorrect. It appears to be caused by the deepmerge in maraca's _updateStore duplicating entries in the port range array.

PR #3854, opened 2018-10-29
Fix array replace

huwshimi:
Fixes: #3852.


huwshimi:
[Fix status port arrays](https://trello.com/c/9QvjY2yw/1036-fix-status-port-arrays)

huwshimi:
QA:
- on prod add 'apache-drill' to a new model
- deploy and open the status
- once the unit has deployed you should only see two ports in the list

Issue #3856, opened 2018-10-29
Borked Google logo


buggtb:
On FF I'm seeing this both in my usual session and incognito:

<img width="1168" alt="screenshot 2018-10-29 at 12 54 12" src="https://user-images.githubusercontent.com/103544/47651593-8eea9a00-db7b-11e8-8350-4fc9e96f4cf6.png">

PR #3861, opened 2018-10-31
Fixed Google Cloud Platform logo anomaly on Firefox

barrymcgee:
Fixes https://github.com/juju/juju-gui/issues/3856


barrymcgee:
## QA

- Go to Credentials table **in Firefox**
- Click 'Add credentials'
- Logos should look as follows;

screenshot 2018-10-31 at 10 04 53

package.json
```diff
@@ -31,6 +31,7 @@
     "react-dnd-html5-backend": "^5.0.1",
     "react-dom": "^16.5.2",
     "shapeup": "0.2.5",
+    "svg-sprite": "^1.3.6",
```
hatched: We actually left this out intentionally so that we didn't install it unless you needed to build the svg's

barrymcgee: Ok, will remove...

PR #3863, opened 2018-11-01
Format all .scss files using Prettier 

barrymcgee:
After speaking with @hatched last week, we agreed that I would push up a branch having enabled Prettier so we could all see the tangible changes it would make to the codebase.

Please leave pre-conceptions behind and review with an open mind 🙏🏻 Vanilla conversion, pulling out react components, implementing Jujulib are all on our roadmap and require large-scale refactoring of the codebase - [in my humble opinion](https://docs.google.com/document/d/1m62KWot5dxN4dOx1-7CuvmWm7fU-if_GhItcBLn8GC0/edit), Prettier is an incredibly valuable tool to aid that process.

It's also important that you all review and give an informed decision either way or we're just back where we're started with two strong opinions and everyone else on the fence. :)

For clarity, I'm going to split this task into two PRs, one for Sass and one for Javascript. This PR is Sass only.

Each branch will use the same [Prettier config file](https://prettier.io/docs/en/configuration.html). That is;

```
{
  "bracketSpacing": false,
  "singleQuote": true,
  "printWidth": 95
}
```

* I've stuck with a print width of 95 as that is what is currently enforced by Eslint but I would recommend increasing that to 120 across the board. 95% of the code never hits the 95 print limit but anything that does would benefit from a little more room before wrapping. I'm guessing a print width of 120 would cover 99% of all code and is still viewable on most if not all modern, common resolutions.

* Restricting the line width to 95 meaning removing the 2 tab rule from `.scss-lint.yml` as Prettier will indent a wrapping line by two tabs. Prettier will enfore all indentation regardless making this `.sass-lint` rule arguably redundant.

* I've removed all references to `// scss-lint ignore` as that tool is now deprecated and is no longer used by this project. 

* As a result, we have less significantly less 'ignore' comments than we had previously.

*  There's only two instances in the whole codebase where I've used `/* prettier-ignore */` and they are very edge case.
	- 1. One is the Sass map we used to z-index order the panels. Purely for the use case of this map, it makes it easier to visualise as a stack rather than a comma separated inline list.
	- 2. In `_notifications.scss` we interpolate Sass variables inside a base64 SVG to use as a background URL. Prettier didn't break them but it did insert wierd whitespace as it tried to them figure out.

----------------

Next PR: [Running Prettier over Javascript files](https://github.com/juju/juju-gui/pull/3864)


barrymcgee:
Addendum: Prettier has now been adopted by [Vanilla](https://github.com/vanilla-framework/vanilla-framework/pull/2012) and [Snapcraft](https://github.com/canonical-websites/snapcraft.io/pull/1082)

jujugui/static/gui/src/app/assets/css/canvas/_service.scss
```diff
@@ -191,12 +190,20 @@
     }
 
     @-webkit-keyframes serviceBlockHaloSelected {
-      0% { opacity: 0; }
-      100% { opacity: .3; }
+      0% {
```
hatched: I can see how this would be an improvement if you had multiple styles applied, but when there is just one, it's actually easier to glance with just the one 🤷‍♂️

jujugui/static/gui/src/app/components/added-services-list/_added-services-list.scss
```diff
@@ -2,7 +2,8 @@
   @extend %fade-in;
   @extend %floating-panel;
   @extend %scroll-children;
-  $extra-space: $navbar-height + $panel-space + $environment-header-height + $panel-space + 30px + 44px; //44px for the beta logo
+  $extra-space: $navbar-height + $panel-space + $environment-header-height + $panel-space +
```
hatched: Is this where it's enforcing a line limit?

I didn't know we had a line limit for scss?

barrymcgee: Prettier enforces a line limit, I believe it's core to how it works - at the very least there is no config to disable it.

By default, it's 80.  In the config, I've set it to the same as what we already have in Eslint, 95.

The above is a good example of why I'd be in favour of pushing it out to 120. Such is the block nature of CSS, these sorts of rules are relatively rare anyway.

I note your comment on the other PR about enforcing 95 to have two files side-by-side but even with 95 as is, I can't find two JS files in the codebse that side by side on my 15" MBP without some amount of horizontal scrolling.

jujugui/static/gui/src/app/assets/css/canvas/_service.scss
barrymcgee: Yes, I would agree here. But, at the risk of sounding like a broken record, nobody has to like all of the things...

barrymcgee:
@hatched Sass-lint does not have a `--fix` option like Eslint. There are some tools around that [try to do it](https://github.com/sasstools/sass-lint/issues/904), but none as wide supported or adopted as Prettier.

anthonydillon:
For my 2 cents there is nothing concerning to me about this branch. In fact it fixes a number of styling problems.

barrymcgee:
I have just had a call with @hatched and a conclusion has been reached - we should adopt Prettier for SCSS files now but wait until Jeff has concluded `jujulib` refactoring before adopting Prettier for JS files which is expected to be about one month from now.

Actions:

- Close this PR and re-open another after rebasing on latest `develop`
- Add a glob to the Prettier ignore file to ignore JS files for now 
- Add a precommit hooks for `.scss` files only

Thank you all for your time and patience with these reviews. 🙏🏻

PR #3864, opened 2018-11-01
Format all .js files using Prettier

barrymcgee:
After speaking with @hatched last week, we agreed that I would push up a branch having enabled Prettier so we could all see the tangible changes it would make to the codebase.

For clarity, I'm going to split this task into two PRs, one for Sass and one for Javascript. 

This PR is JS only.

Each branch will use the same Prettier config file. That is;

```
{
  "bracketSpacing": false,
  "singleQuote": true,
  "printWidth": 95
}
```

* I've stuck with a print width of 95 as that is what is currently enforced by Eslint but I would recommend increasing that to 120 across the board. 95% of the code never hits the 95 print limit but anything that does would benefit from a little more room before wrapping. I'm guessing a print width of 120 would cover 99% of all code and is still viewable on most if not all modern, common resolutions.

* I've haven't used any `// prettier-ignore` flags in this PR.


hatched:
The rational behind the line length limit is to keep the lines short enough so you can have multiple files side by side on laptop screens. You don't increase that and reduce the legibility to serve some formatting tool :)

jujugui/static/gui/src/app/components/added-services-list/label/test-label.js
```diff
@@ -60,10 +72,13 @@ describe('AddedServicesLabel', () => {
...
-    component.find('.inspector-view__label-link-list li').at(1).simulate('click');
+    component
+      .find('.inspector-view__label-link-list li')
+      .at(1)
+      .simulate('click');
     assert.equal(changeState.callCount, 1);
-    assert.deepEqual(
-      changeState.args[0][0], {postDeploymentPanel: 'wiki-simple/bundle/4'});
+    assert.deepEqual(changeState.args[0][0], {
+      postDeploymentPanel: 'wiki-simple/bundle/4'
```
jaycee: Prettier is making some choices here that I personally find less legible, much like this choice for wrapping.

jujugui/static/gui/src/app/components/budget-table/row/test-row.js
```diff
@@ -278,9 +285,27 @@ describe('BudgetTableRow', function() {
       acl,
       plansEditable: true
     });
-    assert.equal(wrapper.find('Button').at(0).prop('disabled'), true);
-    assert.equal(wrapper.find('Button').at(1).prop('disabled'), true);
-    assert.equal(wrapper.find('Button').at(2).prop('disabled'), true);
+    assert.equal(
```
hatched: In this case you actually want them in a single line (like I thought it would have done) as it reads from left to right.

jujugui/static/gui/src/app/components/create-payment-user/create-payment-user.js
```diff
@@ -116,9 +110,9 @@ class CreatePaymentUser extends React.Component {
       name: address.name,
       email: refs.emailAddress.getValue(),
       addresses: [address],
-      vat: business && refs.VATNumber.getValue() || null,
+      vat: (business && refs.VATNumber.getValue()) || null,
```
hatched: I don't get this one....

jujugui/static/gui/src/app/components/deployment-flow/budget/test-budget.js
```diff
@@ -53,84 +60,90 @@ describe('DeploymentBudget', function() {
...
-              <div className="two-col">
-                Credit limit: $100
-              </div>
-              <div className="ten-col last-col">
-                Available credit: $500
-              </div>
-              <div className="one-col">
-                Increase
-              </div>
+              <div className="two-col">Credit limit: $100</div>
```
hatched: It isn't consistent here....sometimes it puts them in a line, sometimes it breaks them over multiple.... even if the line length isn't hit....maybe if the content IN the html is a statement `{}`

scripts/transpile.js
```diff
@@ -66,43 +70,51 @@ if (spliceLength == 0) {
...
-      fs.writeFile(fullPath, full.code, writeHandler);
-      const min = babel.transform(data, {
-        presets: ['babel-preset-babili'],
-        plugins,
-        compact: true,
-        comments: false
-      });
-      fs.writeFile(`${fullPath.replace('.js', '-min.js')}`, min.code, writeHandler);
-    });
+    fs.readFile(
```
hatched: no no no, stay out of my nodejs prettier!

jujugui/static/gui/src/test/test_controller_api.js
```diff
@@ -294,30 +308,28 @@ describe('Controller API', function() {
...
+        }
+      });
       assert.strictEqual(controllerAPI.get('controllerAccess'), 'add-model');
     });
 
     it('stores user information (legacy addmodel)', function() {
       controllerAPI.login();
       // Assume login to be the first request.
-      conn.msg({'request-id': 1,
+      conn.msg({
```
hatched: This actually is nicer

jujugui/static/gui/src/test/test_controller_api.js
```diff
@@ -294,30 +308,28 @@ describe('Controller API', function() {
...
+        'request-id': 1,
         response: {
-          facades: [
-            {name: 'Client', versions: [0]},
-            {name: 'ModelManager', versions: [4]}
-          ],
+          facades: [{name: 'Client', versions: [0]}, {name: 'ModelManager', versions: [4]}],
```
hatched: and then it has to go do this?!?!?!? wut?

jujugui/static/gui/src/test/test_controller_api.js
```diff
@@ -69,9 +74,15 @@ describe('Controller API', function() {
   });
 
   afterEach(function() {
-    cleanups.forEach(function(action) {action();});
-    if (controllerAPI && controllerAPI.destroy) {controllerAPI.destroy();}
-    if (conn && conn.destroy) {conn.destroy();}
+    cleanups.forEach(function(action) {
```
hatched: this is nicer.

jujugui/static/gui/src/app/components/profile/model-list/model-list.js
```diff
@@ -157,11 +165,7 @@ class ProfileModelList extends React.Component {
     @return {Object} The model list as JSX.
   */
   _generateModels() {
-    const icons = new Map([
-      ['read', 'show_16'],
-      ['write', 'edit_16'],
-      ['admin', 'user_16']
-    ]);
+    const icons = new Map([['read', 'show_16'], ['write', 'edit_16'], ['admin', 'user_16']]);
```
hatched: nobody should ever write a map or multi-dimensional array like this...

jujugui/static/gui/src/app/components/added-services-list/label/test-label.js
barrymcgee: Sure, I'd probably agree in this case. There are other parts I wouldn't personally choose either but in my opinion, the ability to completely delegate formatting to a computer and never have to go-round and fix a linting error that broke a build when I'm already at dinner far outweighs these personal preferences that I'll invariably forget I even ad in a couple of weeks.

hatched: It's a person who has to read this in the future so prioritizing consistency and legibility is a priority

anthonydillon: Consistency is a major concern with Prettier. This always for no inconsistency to creep in over time that does not match a linting rule. Personally I feel this code is no less legible but of course that is a personal preference.

hatched: But it's not consistent within itself...

Compare this example to the one here: https://github.com/juju/juju-gui/pull/3864/files/55ffb22776559cbe5947331352b95bc46a38f245#diff-c92c703841e3a00e72786a3a0ad5cf9bR100

On this line it moved the first argument onto the line with the function call, but in the link it has moved it below...

anthonydillon: Can you link to the line please?

hatched: That is....I don't think GH handles 60k line diffs well ;)

scroll down and see for where I tell prettier to stay out of my nodejs ;) in transpile.js

anthonydillon: Interesting, I can't explain why they are different. But should we care to that level?

jaycee: > Sure, I'd probably agree in this case. There are other parts I wouldn't personally choose either but in my opinion, the ability to completely delegate formatting to a computer and never have to go-round and fix a linting error that broke a build when I'm already at dinner far outweighs these personal preferences that I'll invariably forget I even had in a couple of weeks.

That's a little dismissive, isn't it? 

I'm seeing multiple places @hatched has highlighted where I feel like the code has been made less legible--and for me legibility is a primary concern. I don't care about random formatting preferences, tbh, but when code is actually getting harder to read I don't value the time tradeoff of automatic formatting.

I agree that I don't want to argue over styling stuff, but much of what I'm seeing doesn't lead to arbitrary code style, it's doing stuff that again, I find harder to read. And like, 90% of what I'm doing in the gui is reading the code.

I still don't really care here--if the consensus is go with prettier, fine. But I'm not won over that this is *positive* change.

barrymcgee: @jaycee It was in no way intended to be dismissive which is why I was careful to state it was only my opinion but I'm sorry if you felt it was dismissive or that's how it read.

If you've reviewed the proposal and you're not won over that this is a positive change, then for the purposes of this exercise you're a nay. Thanks for taking the time to review.

jujugui/static/gui/src/app/components/accordion-section/accordion-section.js
```diff
@@ -59,24 +59,18 @@ class AccordionSection extends React.Component {
...
-      className="accordion-section__title"
-      onClick={onClick}
-      role={role}>
-      <span className="accordion-section__title-content">
-        {this.props.title}
-      </span>
-      {icon}
-    </div>);
+    return (
+      <div className="accordion-section__title" onClick={onClick} role={role}>
```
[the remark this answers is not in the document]
barrymcgee: I respectfully disagree, HTML has been written this way since day zero. But again, subjective.

jujugui/static/gui/src/app/components/accordion-section/test-accordion-section.js
```diff
@@ -8,14 +8,15 @@ const AccordionSection = require('./accordion-section');
...
-      title={options.title || 'My title!'}>
-      {options.children === undefined ? (<span>Hello</span>) : options.children}
-    </AccordionSection>
-  );
+  const renderComponent = (options = {}) =>
+    enzyme.shallow(
+      <AccordionSection
+        startOpen={options.startOpen === undefined ? false : options.startOpen}
+        title={options.title || 'My title!'}
+      >
```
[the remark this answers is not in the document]
barrymcgee: ..hoisted closing brackets are more difficult for me to read. Again, subjective. Either way, this is a Prettier config option.

jujugui/static/gui/src/app/components/added-services-list/label/test-label.js
barrymcgee: @hatched @anthonydillon Prettier is not inconsistent - it respects the print width. Here is the two examples cited side-by-side at char width 95: https://bit.ly/2QcOjFf - they are identical.

(...which make me wonder if this code was formatted with the default setting of 80)

jujugui/static/gui/src/app/components/accordion-section/accordion-section.js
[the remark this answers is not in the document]
Lukewh: I also prefer the prettier version here. For me, it's easier to see the child `span` at a glance. When scanning a file I generally don't care about all the props of every element, I want to find the element I need to investigate/ modify further.

jujugui/static/gui/src/app/components/accordion-section/accordion-section.js
```diff
@@ -85,12 +79,11 @@ class AccordionSection extends React.Component {
...
   */
   _generateContent() {
-    return (<div
-      className="accordion-section__content"
-      ref="content"
-      style={this._getStyle()}>
-      {this.props.children}
-    </div>);
+    return (
+      <div className="accordion-section__content" ref="content" style={this._getStyle()}>
```
Lukewh: Personal preference, but I much prefer the 
```
(
  </>
);
```

format to
```
(<
/>)
```

The indentation points to the important part of the return

jujugui/static/gui/src/app/components/accordion-section/accordion-section.js
```diff
@@ -101,18 +94,12 @@ class AccordionSection extends React.Component {
...
   }
-};
+}
 
 AccordionSection.propTypes = {
-  children: PropTypes.oneOfType([
-    PropTypes.string,
-    PropTypes.object
-  ]),
+  children: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
```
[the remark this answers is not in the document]
Lukewh: That's one thing I have noticed - it does value line length quite highly. I don't however, think it affects my ability to read and understand the code

jujugui/static/gui/src/app/components/accordion-section/test-accordion-section.js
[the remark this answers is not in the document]
Lukewh: I prefer the dingle-dangle, as previously mentioned is helps with differentiating props and children.

jujugui/static/gui/src/app/components/added-services-list/added-services-list.js
```diff
@@ -37,7 +36,9 @@ class AddedServicesList extends React.Component {
         lastInList={lastInList}
         ref={'AddedServicesListItem-' + service.get('id')}
         service={service}
-        serviceModule={this.props.serviceModule} />);
+        serviceModule={this.props.serviceModule}
+      />
+    );
```
Lukewh: I find this much easier to read and understand.

Indentation is a pointer to the structure of the code, having things open at different indentations and close all on one confuses my simple mind :face_with_head_bandage:

jujugui/static/gui/src/app/components/budget-chart/budget-chart.js
```diff
@@ -17,14 +17,14 @@ class BudgetChart extends React.Component {
   _generateStyles(limit, allocated, newAllocations) {
     // Set the width of the new allocations bar to the percent of the limit.
     // If the limit is zero the percent will equal NaN, hence the || 0.
-    var newWidth = ((newAllocations / limit * 100) || 0) + '%';
+    var newWidth = ((newAllocations / limit) * 100 || 0) + '%';
```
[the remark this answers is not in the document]
jaycee: It did, but in this case order of operations means there was no change. This seems fine, albeit unnecessary.

jujugui/static/gui/src/app/components/added-services-list/label/test-label.js
jaycee: > @jaycee It was in no way intended to be dismissive which is why I was careful to state it was only my opinion but I'm sorry if you felt it was dismissive or that's how it read.
> 
> If you've reviewed the proposal and you're not won over that this is a positive change, then for the purposes of this exercise you're a nay. Thanks for taking the time to review.

I think I slightly misread your original reply, so I'll apologize for taking offense. :stuck_out_tongue_closed_eyes: 

I'm actually still going through this; it's quite a large diff, and the things that stick out are obviously the bits I disagree with. I'm trying to go through and find the bits I like.

jaycee: > @hatched @anthonydillon Prettier is not inconsistent - it respects the print width. Here is the two examples cited side-by-side at char width 95: https://bit.ly/2QcOjFf - they are identical.

Ok, I actually am happier looking at this set. Thanks for providing that example @barrymcgee

scripts/transpile.js
jaycee: @barrymcgee @hatched I assume we can tell prettier to stay out of certain paths? I'm not sure this needs to be treated differently from the rest of the code, but if it's a problem perhaps that's something we can do?

jujugui/static/gui/src/app/components/added-services-list/item/item.js
```diff
@@ -107,29 +101,20 @@ class AddedServicesListItem extends React.Component {
...
-        <span className="inspector-view__item-name">
-          {service.name}
-        </span>
-        <span className="inspector-view__status-block">
-          {statusIndicator}
-        </span>
+        <span className="inspector-view__item-count">{service.unit_count}</span>{' '}
+        <span className="inspector-view__item-name">{service.name}</span>
+        <span className="inspector-view__status-block">{statusIndicator}</span>
       </li>
```
jaycee: This does look better to me; I like where possible having a span be a single line.

jujugui/static/gui/src/app/components/boolean-config/boolean-config.js
```diff
@@ -93,22 +90,21 @@ class BooleanConfig extends React.Component {
...
+              <div className="boolean-config--handle" />
             </label>
           </div>
         </div>
         <div
           className="boolean-config--description"
-          dangerouslySetInnerHTML={{__html: this.props.option.description}}>
-        </div>
+          dangerouslySetInnerHTML={{__html: this.props.option.description}}
+        />
```
jaycee: Ok, I really like that it's catching empty sections and converting them to self-closing bits.

jujugui/static/gui/src/app/components/budget-table/row/row.js
```diff
@@ -362,17 +339,10 @@ class BudgetTableRow extends React.Component {
...
-        </div>
-        <div className={editableWidth + ' no-margin-bottom'}>
-          $1
-        </div>
-        <div className={editableWidth + ' no-margin-bottom'}>
-          {this._generateAllocation()}
-        </div>
-        <div className={
-          'one-col no-margin-bottom' + (plansEditable ? '' : ' last-col')}>
+        <div className="three-col no-margin-bottom">{this._generateSelectedPlan()}</div>
```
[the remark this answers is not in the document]
jaycee: I actually need to disagree; this reads more clearly to me.

scripts/transpile.js
barrymcgee: @jaycee Yes, there is a `.prettierignore` file in root which will take any file or folder glob/path and respect it.

jujugui/static/gui/src/app/components/budget-table/row/test-row.js
jaycee: Here I agree; this reads oddly to me.

jujugui/static/gui/src/app/components/budget-chart/budget-chart.js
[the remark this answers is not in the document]
hatched: It was a rhetorical comment ;) The issue here is that it's trying to be too smart, it's now gone beyond modifying style and into modifying how a program may operate.

jujugui/static/gui/src/app/components/create-payment-user/create-payment-user.js
barrymcgee: See: https://prettier.io/blog/2018/05/27/1.13.0.html#insert-more-parentheses-in-math-expressions-4413-and-4407-by-duailibe

jujugui/static/gui/src/app/components/added-services-list/item/item.js
```diff
@@ -69,14 +66,11 @@ class AddedServicesListItem extends React.Component {
...
     var service = props.service;
-    return classNames(
-      'inspector-view__list-item',
-      {
-        'visibility-toggled': service.get('highlight') || service.get('fade'),
-        hover: props.hovered,
-        'inspector-view__list-item__last': this.props.lastInList
-      }
-    );
+    return classNames('inspector-view__list-item', {
```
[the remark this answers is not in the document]
barrymcgee: These objects are formatted differently as they are being used differently. 

In this example, the object is a passed as a param to the `classNames` function. 

In the other file below, `test-item.js`, the object is part of an array of objects.

You can see these differences in isolation in this playground I made: https://bit.ly/2AIEAB6

PR #3868, opened 2018-11-02
Exclude grid template from expanded credential row

### LINKED ISSUE #3866: Editing credentials UI should use all the space
When editing credentials in the profile table to UI does not use all the available width inside the expanded panel.
screenshot_2018-11-02 juju gui

barrymcgee:
Fixes #3866


barrymcgee:
## QA

- Go to Credentials form on Profile
- Add an AWS credential if you do not already have one
- Click to edit it from the actions drop down in the table
- Verify the edit form layout is correct, like the following screenshot;

screenshot 2018-11-02 at 14 43 39

Issue #3817, opened 2018-09-26
Trash icon has incorrect padding in profile table


hatched:
To reproduce, hover over a model in your user profile.

<img width="44" alt="screen shot 2018-09-26 at 2 47 19 pm" src="https://user-images.githubusercontent.com/532033/46108373-26cf1f80-c19b-11e8-9351-8625eb170c7e.png">

PR #3869, opened 2018-11-02
Center align trash icon in Profile Models table

barrymcgee:
Fixes https://github.com/juju/juju-gui/issues/3817


barrymcgee:
## QA

- Run feature branch
- Verify trash can icon on the right-hand side of Models Profile table is centre aligned, both in the table view and the expanded content view, as per screenshots;

screenshot 2018-11-02 at 17 56 08
screenshot 2018-11-02 at 17 56 31

PR #3873, opened 2018-11-06
Add styling override for "Remove credentials” dialog box

### LINKED ISSUE #3871: Remove credential dialog box mis-styled
screen shot 2018-11-05 at 4 20 23 pm

barrymcgee:
Fixes #3871

### DIFF
--- a/jujugui/static/gui/src/app/components/profile/credential-list/_credential-list.scss
+++ b/jujugui/static/gui/src/app/components/profile/credential-list/_credential-list.scss
@@ -136,7 +136,8 @@
     button {
       width: 100%;
     }
-     .button--destructive {
+
+    .button--destructive {
       background: $color-negative;
       color: $color-light;
     }
--- a/jujugui/static/gui/src/app/components/profile/credential-list/_credential-list.scss
+++ b/jujugui/static/gui/src/app/components/profile/credential-list/_credential-list.scss
@@ -123,4 +123,22 @@
   .dropdown-menu__list-item-link {
     padding: 0 0 0.5rem;
   }
+
+  // XXX These override styles are required until the <PopUp> components is converted to Vanilla v1
+  .button-row {
+    width: 100%;
+    margin: 0;
+    padding: 0;
+    display: grid;
+    grid-template-columns: 1fr 1fr;
+    grid-gap: 1rem;
+
+    button {
+      width: 100%;
+    }
+     .button--destructive {
+      background: $color-negative;
+      color: $color-light;
+    }
+  }
 }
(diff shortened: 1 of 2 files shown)


barrymcgee:
## QA

- Run feature branch
- Go to Profile
- Click 'Credentials' from left hand sidebar
- Click 'Delete' from the actions icon in the table row
- the Dialog box should look like;

screenshot 2018-11-06 at 09 33 54

spencerbygraves:
@clagom and I think that 'Remove' would be a better word for the red button.

'Continue' makes it sound like there's another step which I think there isn't?

jujugui/static/gui/src/app/components/profile/credential-list/_credential-list.scss
```diff
@@ -124,6 +124,25 @@
   .dropdown-menu__list-item-link {
     padding: 0 0 0.5rem;
   }
+
+  // XXX These override styles are required until the <PopUp> components is converted to Vanilla v1
+  .button-row {
+    width: 100%;
+    margin: 0;
+    padding: 0;
+    display: grid;
```
anthonydillon: Why grid instead of flex?

jujugui/static/gui/src/app/components/profile/credential-list/_credential-list.scss
```diff
@@ -124,6 +124,25 @@
...
+    grid-template-columns: 1fr 1fr;
+    grid-gap: 1rem;
+
+    button {
+      width: 100%;
+    }
+
+    .button--destructive {
```
anthonydillon: Can you use p-button--negative?

jujugui/static/gui/src/app/components/profile/credential-list/_credential-list.scss
barrymcgee: No particular reason - I think it's six of one, half a dozen of the other in this case.

jujugui/static/gui/src/app/components/profile/credential-list/_credential-list.scss
barrymcgee: No, as `.button--destructive` is part of the <PopUp> component which is used elsewhere in the app in places that are not encapsulated in the `v1` namespace. Changing the markup of the button would fix it in Profile but break it everywhere else, hence the comment above this which is that these override styles are necessary until `<PopUp>` has been converted.

anthonydillon:
> @clagom and I think that 'Remove' would be a better word for the red button.
> 
> 'Continue' makes it sound like there's another step which I think there isn't?

I would be verbose and say 'Remove credential'

PR #3876, opened 2018-11-08
Edit plans card in entity content

barrymcgee:


barrymcgee:
## QA

- Run feature branch
- Go to `/u/spiculecharms/apache-drill/24`
- Check that a card now appears in the right-hand sidebar displaying pricing info

Before:

screenshot 2018-11-08 at 10 16 14

After:

screenshot 2018-11-08 at 10 16 30

jujugui/static/gui/src/app/components/entity-details/content/content.js
```diff
@@ -538,39 +538,30 @@ class EntityContent extends React.Component {
     var plansList = [];
     plans.forEach((plan, i) => {
       var classes = classNames(
-        'entity-content__plan',
-        'four-col',
-        {'last-col': (i + 1) % 3 === 0});
+        'entity-content__plan'
```
hatched: If there is only a single class you can just set that in the JSX.

spencerbygraves:
Thanks @barrymcgee looks much better 👍

barrymcgee:
@hatched Made that requested change.

PR #3877, opened 2018-11-09
Require SCSS files

huwshimi:


jujugui/static/gui/src/app/assets/css/_mixins.scss
```diff
@@ -1,3 +1,6 @@
+@import './plugins/cloud-vanilla-theme/scss/global-settings';
```
hatched: Are imports deduped?

jujugui/static/gui/src/app/components/budget-chart/budget-chart.js
```diff
@@ -4,6 +4,8 @@
 const PropTypes = require('prop-types');
 const React = require('react');
 
+require('./_budget-chart.scss');
```
hatched: I don't think we have any duplicate requires, but are they deduped as well?

jujugui/static/gui/src/app/assets/css/_mixins.scss
huwshimi: The imports only import variables, so when compiled it only uses the values as required.

jujugui/static/gui/src/app/components/budget-chart/budget-chart.js
huwshimi: They are deduped in the browserify scss builder! But I think you're right that we don't include anything twice anyway.

hatched:
Can you update the changelog plz.

hatched:
QA Issue:

screen shot 2018-11-14 at 7 00 06 pm

The size of the CSS has also increased quite substantially 
From 356KB -> 61.2KB gz
To 396KB -> 63.3KB gz

hatched:
Thanks @huwshimi

PR #3880, opened 2018-11-16
Add error boundaries to Profile components

### LINKED ISSUE #3865: Identify and implement Error Boundaries in a top level component
React [Error Boundaries](https://reactjs.org/docs/error-boundaries.html) allow components to fail gracefully without bringing down other components around it. We should be utilising them to increase the resilience of our product. 

Action: Identify a top-level component and implement Error Boundaries.

barrymcgee:
## Done 

- Added error boundaries to Profile section of the GUI to improve resilience.

## QA

- Run this feature branch
- Click each link in the side nav of Profile to ensure each view loads
- Open `credential-list.js` and add `throw new Error('Error!')` to the last line of the `Constructor()` function.
- Reload the GUI and you should see an error message where the Credentials table should be.

## Demo

Before - Without error boundaries: https://cl.ly/8788d419815b
After - With error boundaries: https://cl.ly/9fd93f8d66ba

Fixes #3865


barrymcgee:
@hatched First pass - need to fix those tests and fix a styling bug but wanted to give visibility on WIP.

barrymcgee:
## Done 

- Added error boundaries to Profile section of the GUI to improve resilience.

## QA

- Run this feature branch
- Click each link in the side nav of Profile to ensure each view loads
- Open `credential-list.js` and add `throw new Error('Error!')` to the last line of the `Constructor()` function.
- Reload the GUI and you should see an error message where the Credentials table should be.

Fixes #3865

hatched:
Looks like there are a couple small conflicts after changing the base.

barrymcgee:
@hatched Resolved conflicts, tests will fail but I assume it's due to your ongoing work as doesn't seem related to these changes?

hatched:
That's right, there are failing tests from the jujulib conversion.

jujugui/static/gui/src/app/components/error-boundary/test-error-boundary.js
```diff
@@ -0,0 +1,44 @@
...
+        <h1>Buggy component</h1>
+      </div>
+    );
+  }
+}
+
+describe('Error boundary', function() {
+
+  const renderComponent = (options = {}) => enzyme.shallow(
+    <ErrorBoundary />
```
hatched: These tests are no good because you're not actually passing the content into the component ;)

jujugui/static/gui/src/app/components/error-boundary/test-error-boundary.js
```diff
@@ -0,0 +1,44 @@
...
+      content: (<span>Children content!</span>)
+    });
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('should show error message if available', () => {
+    const renderComponentWithError = (options = {}) => enzyme.shallow(
+      <ErrorBoundary />
+    );
+    const wrapper = renderComponentWithError({
```
hatched: This can just use the same `renderComponent` above

jujugui/static/gui/src/app/components/error-boundary/error-boundary.js
```diff
@@ -0,0 +1,53 @@
...
+
+  render() {
+    const error = this.state.error;
+    if (error) {
+      return (
+        <div className="v1">
+          <div className="row">
+            <div className="col-12">
+              <div className="p-notification--negative">
+                {error.message &&
```
hatched: This message will be a JS error and won't be very helpful to the user. I think it should go under the friendly message, maybe in a collapsed box.

jujugui/static/gui/src/app/components/error-boundary/error-boundary.js
```diff
@@ -0,0 +1,53 @@
...
+                  <p className="p-notification__response">
+                    <strong>{error.message}</strong>
+                  </p>
+                }
+                <p className="p-notification__response">
+                  Hmm.. something has went wrong with this part of the application.
+                  If this problem persists, please <a
+                    className="p-link--external"
+                    href="https://github.com/CanonicalLtd/jujucharms.com">
+                    raise an issue.
```
hatched: "... with the contents of the error message below."

jujugui/static/gui/src/app/components/error-boundary/test-error-boundary.js
barrymcgee: I pass the content object in the next code block, line 30 ;) Why? This is a pattern Huw uses elsewhere which I replicated for consistency.

jujugui/static/gui/src/app/components/error-boundary/test-error-boundary.js
barrymcgee: As above, these twp components will receive different content - the reason I've created a new component here is to ensure a freshly mounted component for this test block as they are still within the same test suite. Perhaps this is an unnecessary concern?

jujugui/static/gui/src/app/components/error-boundary/error-boundary.js
barrymcgee: I've move it below and placed it inside a Vanilla `<code>` block.

barrymcgee:
@hatched The reason why there is no `<ErrorBoundary>` on `<Profile>` itself is because it's required to wrap where the component where it is instantiated meaning, in this case, I'd need to hoist the component into `App.js` and wrap the `<Profile ... />` component there.. I considered that as scope creep for this task given it was intially a spike to see how feasible Error Boundaries were to implement.

barrymcgee:
@hatched 
 I've added a `_error_boundary.scss` file to add a custom top margin on the code block in the nofifcation pattern but it doesn't seem to be applied.. AFAIK I've included it correctly using require but am I missing something? 

screenshot 2018-11-22 at 11 17 08

jujugui/static/gui/src/app/components/error-boundary/test-error-boundary.js
barrymcgee: @hatched Right.. gotcha. I've updated the tests but the snapshot for the bug component still doesn't throw the fall back ui? Can you see what I'm missing?

hatched: Because the broken component is child component of the wrapper it's probably not being executed because of the shallow render. There are hacks to make it render but you should probably render it to the dom. 

See: https://github.com/juju/juju-gui/blob/use-jujulib/jujugui/static/gui/src/app/components/profile/credential-list/test-credential-list.js#L67

And then be sure to clean up afterwards: https://github.com/juju/juju-gui/blob/use-jujulib/jujugui/static/gui/src/app/components/profile/credential-list/test-credential-list.js#L107

barrymcgee:
@hatched Using `enzyme.mount()` instead of `enzyme.shallow()` has fixed up the snapshots to show the correct content.

jujugui/static/gui/src/app/components/error-boundary/test-error-boundary.js
```diff
@@ -0,0 +1,44 @@
...
+      <div>
+        <h1>Buggy component</h1>
+      </div>
+    );
+  }
+}
+
+describe('Error boundary', function() {
+
+  const renderComponent = (options = {}) => enzyme.mount(
```
hatched: Can you add an `aftereach` which unmounts this component?

jujugui/static/gui/src/app/components/error-boundary/test-error-boundary.js
```diff
@@ -0,0 +1,44 @@
...
+
+  it('should not show if no error thrown', () => {
+    const wrapper = renderComponent({
+      content: (<span>Children content!</span>)
+    });
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('should show fallback UI if error thrown', () => {
+
```
hatched: extra line

jujugui/static/gui/src/app/components/error-boundary/error-boundary.js
```diff
@@ -0,0 +1,54 @@
...
+          <div className="row">
+            <div className="col-12">
+              <div className="p-notification--negative">
+                <p className="p-notification__response">
+                  Hmm.. something has went wrong with this part of the application.
+                  If this problem persists, please <a
+                    className="p-link--external"
+                    href="https://github.com/CanonicalLtd/jujucharms.com">
+                    raise an issue</a> with the contents of the error message below.
+                  {error.message &&
```
hatched: We should also have a default message as we're pointing them to an error message 'below'

jujugui/static/gui/src/app/assets/css/base.scss
```diff
@@ -102,12 +102,16 @@ a {
...
+  @include vf-p-buttons;
   @include vf-p-grid;
   @include vf-p-strip;
+  @include vf-p-links;
   @include vf-p-lists;
   @include vf-p-media-object;
   @include vf-p-divider;
   @include vf-p-tooltips;
+  @include vf-p-notification;
+  @include vf-p-icons;
```
hatched: Is this used in here somewhere?

jujugui/static/gui/src/app/assets/css/base.scss
```diff
@@ -102,12 +102,16 @@ a {
   // Include base Vanilla styles
   @include vf-base;
   // Include components as they are used
+  @include vf-p-buttons;
```
hatched: I don't see this being used...

jujugui/static/gui/src/app/components/error-boundary/test-error-boundary.js
```diff
@@ -0,0 +1,47 @@
...
+  });
+
+  it('should show fallback UI if error thrown', () => {
+    const wrapper = renderComponent({
+      content: (<ComponentWithError />)
+    });
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  afterEach(() => {
```
hatched: At a risk of being a little pedantic, would you mind moving this up above any of the `it()` tests. It's kind of a convention, then you can easily see the test setup and teardowns before you get to the tests.

hatched:
Thanks for this @barrymcgee

PR #3881, opened 2018-11-20
Update HeaderBreadcrumb to use jujulib

hatched:
This will fail CI

#### To QA
Load the GUI and it'll open your profile.
Close the profile and the header should have your username and the model switcher.
Opening the model switcher should list your models. (Model switching does not work)
Click the user menu icon and it should open with a list.
The profile link and help links should work.


hatched:
Great, thanks @huwshimi

PR #3886, opened 2018-11-23
Connect to, and display, models using jujulib.

### LINKED ISSUE #3883: Connect to models using jujulib

hatched:
Fixes #3883

### DIFF
--- a/jujugui/static/gui/src/app/init.js
+++ b/jujugui/static/gui/src/app/init.js
@@ -739,7 +739,6 @@ class GUIApp {
             console.log('cannot watch model:', err);
             process.exit(1);
           }
-          console.log(result);
           this._processDeltas(result);
           this.modelConnectionStatus = 'ready';
         });
--- a/jujugui/static/gui/src/app/init.js
+++ b/jujugui/static/gui/src/app/init.js
@@ -375,14 +375,29 @@ class GUIApp {
     */
     this.jujuClient = null;
 
+    /**
+      A list of require statements for the required jujulib model facades.
+      @type {Array}
+    */
+    this._defaultModelFacades = [
+      // Sort facades alphabetically.
+      require('@canonical/jujulib/api/facades/application-v6.js'),
+      require('@canonical/jujulib/api/facades/all-watcher-v1.js'),
+      require('@canonical/jujulib/api/facades/charms-v2.js'),
+      require('@canonical/jujulib/api/facades/client-v1.js'),
+      require('@canonical/jujulib/api/facades/pinger-v1.js')
+    ];
+
+    this._defaultControllerFacades = [
+      // Sort facades alphabetically.
+      require('@canonical/jujulib/api/facades/cloud-v2.js'),
+      require('@canonical/jujulib/api/facades/model-manager-v4.js'),
+      require('@canonical/jujulib/api/facades/pinger-v1.js')
+    ];
+
     const connectionOptions = {
       debug: true,
-      facades: [
-        // Sort facades alphabetically.
-        require('@canonical/jujulib/api/facades/cloud-v2.js'),
-        require('@canonical/jujulib/api/facades/model-manager-v4.js'),
-        require('@canonical/jujulib/api/facades/pinger-v1.js')
-      ],
+      facades: this._defaultControllerFacades,
       wsclass: WebSocket,
       bakery: this.bakery
     };
@@ -718,14 +733,7 @@ valid values are [${validValues.map(i => String(i)).join(', ')}]`;
 
     const connectionOptions = {
       debug: true,
-      facades: [
-        // Sort facades alphabetically.
-        require('@canonical/jujulib/api/facades/application-v6.js'),
-        require('@canonical/jujulib/api/facades/all-watcher-v1.js'),
-        require('@canonical/jujulib/api/facades/charms-v2.js'),
-        require('@canonical/jujulib/api/facades/client-v1.js'),
-        require('@canonical/jujulib/api/facades/pinger-v1.js')
-      ],
+      facades: this._defaultModelFacades,
       wsclass: WebSocket,
       bakery: this.bakery
     };
@@ -775,7 +783,7 @@ valid values are [${validValues.map(i => String(i)).join(', ')}]`;
       deltas.push([kind + 'Info', operation, entityInfo]);
     });
     deltas.sort((a, b) => {
-      // Sort items not in our hierarchy last.
+      // Sort items that are not not in our hierarchy last.
       if (!cmp[a[0]]) {
         return 1;
       }
--- a/jujugui/static/gui/src/app/init.js
+++ b/jujugui/static/gui/src/app/init.js
@@ -195,8 +195,8 @@ class GUIApp {
     */
     this.modelController = new ModelController({
       db: this.db,
-      env: this.modelAPI,
-      charmstore: this.charmstore
+      charmstore: this.charmstore,
+      getModelConnection: () => this.modelConnection
     });
     /**
       Application instance of the endpointsController.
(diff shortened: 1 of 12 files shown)


jujugui/static/gui/src/app/init.js
```diff
@@ -333,11 +333,45 @@ class GUIApp {
...
+      Once login is complete this will hold the connection to the model.
+      @type {Object}
+    */
+    this.modelConnection = null;
+
+    /**
+      The model uuid that is currently connected to.
+      @type {string}
+    */
+    this.activeModelUUID = null;
```
huwshimi: Is there a reason you added this instead of  using `this.modelUUID`?

jujugui/static/gui/src/app/init.js
```diff
@@ -333,11 +333,45 @@ class GUIApp {
...
+      the application until the model connection is completely ready and
+      native promises do not allow external inspection on its status.
+
+      Allowed values are:
+        null
+        'ready'
+        'connecting'
+      @type {String}
+    */
+    this.modelConnectionStatus = null;
```
huwshimi: Should we use a getter/setter for this to enforce the allowed values?

jujugui/static/gui/src/app/init.js
```diff
@@ -659,13 +693,97 @@ class GUIApp {
...
-    next();
+
+    const modelURL = jujulib.generateModelURL(
+      this.applicationConfig.apiAddress, newModelUUID);
+
+    const connectionOptions = {
+      debug: true,
+      facades: [
+        // Sort facades alphabetically.
+        require('@canonical/jujulib/api/facades/application-v6.js'),
```
huwshimi: This is going to do a require every time we change models. There's nothing dynamic going on here so would it be better to store this array at the top of the file with the other requires?

jujugui/static/gui/src/app/init.js
```diff
@@ -659,13 +693,97 @@ class GUIApp {
...
+      machineInfo: 4,
+      annotationInfo: 5,
+      remoteapplicationInfo: 100
+    };
+    result.deltas.forEach(delta => {
+      const [kind, operation, entityInfo] = delta;
+      deltas.push([kind + 'Info', operation, entityInfo]);
+    });
+    deltas.sort((a, b) => {
+      // Sort items not in our hierarchy last.
```
huwshimi: "last" or "list "?

jujugui/static/gui/src/app/init.js
hatched: I wanted to leave the old code in until it was all done so there was a name collision. I also wanted the name to be a little more explicit.

jujugui/static/gui/src/app/init.js
hatched: This was copied over and updated....I'm guessing it should be list :)

jujugui/static/gui/src/app/init.js
hatched: Yes good point, it should cache, but I'll move it out to be explicit.

jujugui/static/gui/src/app/init.js
hatched: good call

jujugui/static/gui/src/app/init.js
huwshimi: OK, we'll just have to remember to consolidate these afterwards if we can, would be good to not have too many places to get out of sync.

hatched: Yeah that's a good point, maybe I'll remove some of that stuff in this branch too so we don't have too much cruft left around.

jujugui/static/gui/src/app/init.js
hatched: It actually is "last" as in "to put the items, not in the list, in the last position" I'll update the comment to be clearer.

jujugui/static/gui/src/app/init.js
hatched: I have modified all of the uses of  `modelUUID` to instead use the `activeModelUUID` property.

jujugui/static/gui/src/app/init.js
huwshimi: Oh, hah, thanks.

jujugui/static/gui/src/app/init.js
huwshimi: Thanks!

jujugui/static/gui/src/app/init.js
hatched: done

jujugui/static/gui/src/app/init.js
hatched: done

jujugui/static/gui/src/app/init.js
hatched: done

jujugui/static/gui/src/app/state/handlers.js
```diff
@@ -0,0 +1,221 @@
...
+      });
+  }
+
+  /**
+    Process the deltas that are returned from the megawatcher. This code
+    was moved over from the old model api handler and can be improved once we
+    remove the YUI db and replace it with maraca.
+    @param {Object} result The data result from the watch jujulib calls.
+  */
+  _processDeltas(result) {
```
hatched: This was copied over and is not tested. This will soon be handled by maraca.

jujugui/static/gui/src/app/state/test-handlers.js
```diff
@@ -0,0 +1,126 @@
...
+'use strict';
+
+const StateChangeHandlers = require('./handlers');
+const State = require('./state');
+
+const jujulibTestHelper = require('@canonical/jujulib/api/test-helpers');
+const AllWatcherResponse = require('@canonical/jujulib/tests/data/allwatcher-response');
+
+function setupApp(options = {}) {
+  return {
```
hatched: The goal is to have as small of a surface area to interact with app over time but a lot of these are required simply for legacy interactions and will be removed.

jujugui/static/gui/src/app/state/test-handlers.js
```diff
@@ -0,0 +1,126 @@
...
+
+    it('calls next and does nothing if already connected', () => {
+      const {state, stateChangeHandlers} = setupStateInstance({
+        app: setupApp({
+          activeModelUUID
+        })
+      });
+      stateChangeHandlers.modelConnectionStatus = 'ready';
+      const next = jest.fn();
+      const value = state._dispatchers.model[0][0]({
```
hatched: Grab the model dispatcher from the instance itself.

jujugui/static/gui/src/app/state/test-handlers.js
```diff
@@ -0,0 +1,126 @@
...
+        }
+      }, next);
+      expect(next).toHaveBeenCalledTimes(0);
+      expect(jujuPromise instanceof Promise).toBe(true);
+      ws.queueResponses(new Map([
+        [2, AllWatcherResponse.watchAll],
+        [3, AllWatcherResponse.next]
+      ]));
+      ws.open();
+      jujulibTestHelper.replyToLogin(ws, {
```
hatched: Because this test relies on this new method from `jujulibTestHelper` this branch can't be qa'd until https://github.com/juju/js-libjuju/pull/32 lands, or it can be checked out and `npm link @canonical/jujulib`

package.json
```diff
@@ -9,7 +9,7 @@
     "url": "https://github.com/juju/juju-gui.git"
   },
   "dependencies": {
-    "@canonical/jujulib": "^1.1.0",
+    "@canonical/jujulib": "^1.2.0",
```
hatched: Once jujulib is updated this will also be updated so this branch can be qa'd

hatched:
@huwshimi the changes we discussed have been completed, can you please take another look. Thanks.

jujugui/static/gui/src/app/state/handlers.js
```diff
@@ -0,0 +1,221 @@
...
+      ['hash'],
+      ['special.dd']
+    ]);
+    return {state, stateChangeHandlers: handlers};
+  }
+
+  set modelConnectionStatus(value) {
+    const validValues = [null, 'ready', 'connecting'];
+    if (!validValues.includes(value)) {
+      throw `Invalid value "${value}" set to modelConnectionStatus, \
```
huwshimi: Nice!

jujugui/static/gui/src/app/state/test-handlers.js
```diff
@@ -0,0 +1,126 @@
...
+        })
+      });
+      stateChangeHandlers.modelConnectionStatus = 'ready';
+      const next = jest.fn();
+      const value = state._dispatchers.model[0][0]({
+        model: {
+          uuid: activeModelUUID
+        }
+      }, next);
+      expect(next).toHaveBeenCalled();
```
huwshimi: Any reason you're not using `assert` for all this code?

jujugui/static/gui/src/app/state/__snapshots__/test-handlers.js.snap
```diff
@@ -0,0 +1,94 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`StateChangeHandlers setupState can return a setup state instance 1`] = `
+Object {
+  "*": Array [
+    Array [
+      [Function],
+      undefined,
```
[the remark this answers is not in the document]
hatched: It is! It took me a while to confirm, but basically the callbacks are collected under the keys and the first function is the 'on change' and the second is 'on clear' which is why they are undefined as we no longer use that functionality and let react deal with it.

jujugui/static/gui/src/app/state/test-handlers.js
hatched: Yeah the failure diffs with assert are totally useless with Jest so I had to switch to expect so that I could get intelligent diffs on objects. I also used the jest stub functions instead of sinon.

PR #3888, opened 2018-11-26
Add webpack

huwshimi:


huwshimi:
QA:
- make and run the gui and everything should be as it was before
- make dist should build the dist
- test with gij
- test with the storefront

jujugui/static/gui/src/app/maraca/maraca.js
```diff
@@ -2,7 +2,11 @@
 'use strict';
 
 const clonedeep = require('lodash.clonedeep');
-const deepmerge = require('deepmerge');
+let deepmerge = require('deepmerge');
+// Handle differences between how the tests load this code.
```
hatched: Can you expand on this? I have seen these issues before when not using the `import` syntax with webpack but can you elaborate in this comment for future us.

hatched: Now that we're using webpack should we switch to the `import` syntax? Can we `import` the sass?

package.json
```diff
@@ -58,37 +58,42 @@
...
     "husky": "^1.1.3",
+    "inject-loader": "^4.0.1",
     "jasmine-core": "^3.2.1",
     "jest": "^23.6.0",
     "karma": "^3.0.0",
-    "karma-browserify": "^5.3.0",
     "karma-chai-sinon": "0.1.5",
     "karma-chrome-launcher": "2.2.0",
-    "karma-jasmine": "^1.1.2",
+    "karma-jasmine": "2.0.1",
```
hatched: Is there a reason this is an explicit dep vs using ^ like the rest?

package.json
```diff
@@ -58,37 +58,42 @@
...
+    "webpack-cli": "^3.1.2",
+    "webpack-merge": "^4.1.4"
   },
   "husky": {
     "hooks": {
       "pre-commit": "pretty-quick --staged"
     }
-  }
+  },
+  "browserslist": [
```
hatched: Ahh the same 'hack' from parcel :-)

webpack.common.js
```diff
@@ -0,0 +1,54 @@
...
+module.exports = {
+  entry: './jujugui/static/gui/src/app/init.js',
+  output: {
+    path: path.resolve(__dirname, 'jujugui/static/gui/build/app'),
+    filename: 'init-pkg.js',
+    library: 'JujuGUI',
+    libraryTarget: 'var'
+  },
+  module: {
+    rules: [
```
hatched: Can you please add code comments to each rule explaining what a rule does and why it's there. Webpack configs in projects get unwieldy fast and are quite difficult to read. Also in the prod, dev, test versions.

hatched:
Looks like there is a small issue with the css for the machine info in the DF. This does not happen in the develop dist. Note the padding around the contents.

screen shot 2018-11-26 at 1 45 27 pm

hatched:
It is also tremendously slow to build...`make clean-all && time make gui` with:

Browserify:
```
real    1m17.899s
user    1m3.448s
sys     0m23.517s
```


Webpack:
```
real    5m10.090s
user    5m16.042s
sys     0m25.660s
```

Are you seeing the same results?

package.json
huwshimi: We're about 50/50 with having the `^` in our package.json. We've run into issues with it not actually locking the version down so I tend to use a specific version number.

package.json
huwshimi: Yep! Thanks for that :).

huwshimi:
> ```
> Webpack:
> 
> ```
> real    5m10.090s
> user    5m16.042s
> sys     0m25.660s
> ```
> Are you seeing the same results?

Mine wasn't quite so bad, but still a lot worse than browserify:

real	3m10.956s
user	3m16.798s
sys	0m26.442s

But when I ran it just then it looked like webpack was being run twice. I'll take a look at that.

huwshimi:
I've fixed the Makefile so it only builds the GUI once and the time is now:

real	1m5.939s
user	0m56.340s
sys	0m22.167s

hatched:
Big improvement! 👍 

real    1m6.244s
user    0m57.236s
sys     0m16.668s

hatched:
Thanks for this @huwshimi

PR #3894, opened 2018-12-07
Store the extra-info supported values in the y.models.

### LINKED ISSUE #3893: Pass extra-info through for 'supported' value support

hatched:
Fixes #3893

### DIFF
--- a/jujugui/static/gui/src/app/models/models.js
+++ b/jujugui/static/gui/src/app/models/models.js
@@ -268,9 +268,15 @@ window.yui.add('juju-models', function(Y) {
     }
   }, {
     ATTRS: {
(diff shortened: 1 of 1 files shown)


jujugui/static/gui/src/app/models/models.js
```diff
@@ -268,6 +268,9 @@ window.yui.add('juju-models', function(Y) {
     }
   }, {
     ATTRS: {
+      supported: {},
```
huwshimi: Should this default to false?

hatched: Good call, added some defaults to all 3.

Issue #3904, opened 2019-03-19
Search not working in the store overlay


spencerbygraves:
If I click either of the buttons or any of the links highlighted below, no search results are returned. The loading icon appears but starts over and over again.

![image](https://user-images.githubusercontent.com/9105999/54625443-3bfc7b00-4a67-11e9-9951-6e7d30b1a408.png)

hatched:
The request to the charmstore is made over and over.

PR #3905, opened 2019-03-20
Handle searches without query text

huwshimi:
Fixes: #3904. Search URLs that did not include query text would infinitely reload e.g. listing by tag/entity type.


huwshimi:
QA:
- open the store and click on "View all the charms" or "View all the bundles" or on one of the tags
- the listing should appear
- do a search for a charm or bundle
- the listing should appear

hatched:
Thanks @huwshimi

Issue #3902, opened 2019-03-18
Mailing list link in GUI is broken


mitechie:
Copied from https://bugs.launchpad.net/juju/+bug/1820558

# Steps to reproduce:

* Open the Juju GUI
* Click on the question mark button in the top right
* Under "Mailing lists", click "The Juju Project"

# Expected Behavior

It takes you to the mailing list

# Actual behavior

It takes you here: https://lists.ubuntu.com/mailman/listinfo/juju

Which says:

> No such list juju

# Other info:

Version of Juju: 2.5.2-bionic-amd64

There are more bugs with this menu:

2.
"Get Support" takes the user to a page which does not get them support.

3.
"FAQs" takes the user to a page which is just a general overview of Juju. The `#frequently-asked-questions` anchor is broken.

4.
There is no link to Launchpad for issue creation. There should be.

hatched:
These links need to be updated.

We should add a link to https://discourse.jujucharms.com/

PR #3906, opened 2019-03-25
Update Discourse link

huwshimi:
Fixes: #3902.


huwshimi:
QA:
- open the help screen
- there should now be a discourse link and no mailing list link

jujugui/static/gui/src/app/components/help/__snapshots__/test-help.js.snap
```diff
@@ -170,15 +170,15 @@ exports[`Help renders 1`] = `
           </a>
         </p>
         <p>
-          Mailing lists
+          Forums
```
hatched: "Forum" ?

Same below

huwshimi: Done.

Issue #3898, opened 2019-02-01
Charm detail page on mobile, page jumps up


clagom:
Got this feedback from a user through Usabilla, on jujucharms.com:
"On mobile android chrome -Samsung s8+, when you open a charm or bundle from store to read more information. Example: https://jujucharms.com/canonical-kubernetes/ The page behaviour when I scroll down and the top ribbon expands to show the extra information makes the page suddenly jump up and down, making top paragraphs of information inaccessible."

I couldn't replicate on my smartphone (the top sticky element does not create jumpy page on iPhone), but we might want to investigate this issue.

@barrymcgee @hatched @anthonydillon

hatched:
We discussed this offline and we'll remove the sticky headers from the GUI & jujucharms.com.

PR #3907, opened 2019-03-25
Don't use sticky headers on charms/bundles

huwshimi:
Fixes: #3898.


huwshimi:
QA:
- view a charm/bundle
- scroll down the page
- the header should scroll out of view, and not become sticky

CHANGELOG.md
```diff
@@ -13,12 +13,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed unused svg assets from the build/dist.
 
 ### Fixed
-- Styling issue with Plans selector in sticky header on entity pages.
 - Broken GCP cloud logo in Firefox.
 - Repeating port values on Status pane.
 - Broken styles on the Status pane.
 - Searches without query text would reload forever.
 
+### Removed
```
hatched: 👍 on keeping this updated

jujugui/static/gui/src/app/components/entity-details/header/header.js
```diff
@@ -14,15 +13,6 @@ const SvgIcon = require('../../svg-icon/svg-icon');
 require('./_header.scss');
 
 class EntityHeader extends React.Component {
```
hatched: This may be able to be converted to a functional component now.

PR #3908, opened 2019-03-26
Serve the GUI from flask instead of pyramid.

hatched:
This branch enables serving the GUI from a flask app instead of a pyramid one. 

This is the first step in converting the file structure of the GUI over.

Fixes https://github.com/ubuntudesign/juju-squad/issues/562


.env
```diff
@@ -0,0 +1,4 @@
+PORT=8000
```
nottrobin: We should pick a unique port for this project and then add it to [our ports document](https://canonical-webteam.github.io/practices/project-structure/ports.html). E.g. `8032`.

package.json
```diff
@@ -8,6 +8,16 @@
...
     "url": "https://github.com/juju/juju-gui.git"
   },
+  "scripts": {
+    "build": "webpack --config webpack.dev.js && cp -r node_modules/yui static && rm -rf static/yui/node_modules && ./scripts/version.sh",
+    "clean": "rm -rf static/yui static/build node_modules webapp/__pycache__",
+    "watch": "webpack --config webpack.dev.js --watch",
+    "lint-js": "eslint --rulesdir eslint-rules jujugui/static/gui/src",
+    "lint": "npm run lint-js",
+    "serve": "./entrypoint 0.0.0.0:$PORT",
+    "test-old": "./scripts/test-js.sh && ./scripts/test-js-old.sh",
```
nottrobin: Why do we need this?

templates/_base/base.html
```diff
@@ -0,0 +1,76 @@
+<!DOCTYPE html>
```
nottrobin: These "base" templates are used by `404.html` and `500.html` but not `index.html`? That seems odd...

templates/404.html
```diff
@@ -0,0 +1,21 @@
+{% extends "_base/base.html" %} {% block title %}404: Page not found | Juju{% endblock %} {%
```
nottrobin: I feel maybe `_includes/` or `_layout/` is a more appropriate name than `_base` for the folder. The repetition of `_base/base.html` is a bit jarring.

templates/index.html
```diff
@@ -0,0 +1,269 @@
...
+    <script type="text/javascript">
+      var _gaq = _gaq || [];
+      window._gaq = _gaq;
+      // Do not call _setAccount as we need to get the key from
+      // the config file which is not yet available.
+    </script>
+  </head>
+  <body>
+    <div style="display: none;">
+      ## Include the svg sprite image inside the body. {% include "./sprite.css.svg" %}
```
nottrobin: Would you consider including the contents of `sprite.css.svg` inline here? It would slightly improve performance, and remove an endpoint.

templates/404.html
nottrobin: I see now that you will have copied this from jp.ubuntu.com. But I still don't like it and suggest changing it ;)

templates/_base/base.html
nottrobin: Oh again I guess this is because you copied these pages from jp.ubuntu.com.

So these includes should be replaced with the actual markup from `index.html`.

templates/index.html
```diff
@@ -0,0 +1,269 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr">
```
nottrobin: This markup should be split up into includes, so it can be shared with error pages.

webapp/handlers.py
```diff
@@ -0,0 +1,93 @@
...
+    def utility_processor():
+        """
+        This defines the set of properties and functions that will be added
+        to the default context for processing templates. All these items
+        can be used in all templates
+        """
+
+        return {
+            # Functions
+            "format_date": template_utils.format_date,
```
nottrobin: Since they're not used in this app, I would remove `format_date`, `replace_admin` and `truncate_chars`, along with their accompanying functions from `template_utils.py`.

`get_year` and `versioned_static` are presumably only used because you copied the markup for error pages from jp.ubuntu.com, and so when you split out `index.html` markup into includes so it can be shared with error pages, that old markup will go and you'll no longer be using those context processors either.

In which case you could delete this whole context_processor function, and all of `templare_utils.py`.

webapp/handlers.py
```diff
@@ -0,0 +1,93 @@
...
+
+        return (
+            flask.render_template("500.html", error_name=error_name),
+            status_code,
+        )
+
+    # Global tasks for all requests
+    # ===
+    @app.before_request
+    def clear_trailing():
```
nottrobin: You probably don't want this function. URLs in the GUI appear to in fact have trailing slashes.

webapp/handlers.py
```diff
@@ -0,0 +1,93 @@
...
+        """
+
+        response.headers["X-Hostname"] = socket.gethostname()
+
+        if response.status_code == 200:
+            if flask.session:
+                response.headers["Cache-Control"] = "private"
+            else:
+                # Only add caching headers to successful responses
+                response.headers["Cache-Control"] = ", ".join(
```
nottrobin: I'm guessing this was copied from jp.ubuntu.com. You might want to consider this carefully, in case it might put your app at risk.

nottrobin:
Really nice quick work so far @hatched, thanks!

I left a few comments inline, just push back on or ignore any of them that don't make sense.

One QA issue - static images don't seem to be loading properly:

static-error

There's another issue with the "view all the charms" endpoint at http://localhost:8000/q/store/?type=charm, which seems to end up in an infinite loop, but the same issue seems to be there on live, so it's probably not Flask's fault: https://jujucharms.com/q/store/?type=charm.

package.json
hatched: It's a long story, but we effectively need these old test runners for now. I'll be happy to explain it to you offline.

templates/_base/base.html
hatched: Yeah this was copied from jp but the index is just the one copied from the GUI project with a few minor changes. I'll be updating the index file a lot in the coming PR's

templates/404.html
hatched: I figure we should probably use the same wrapper as jaas.ai in the end. I'll see what we can do about bringing these projects together on the wrapper.

templates/index.html
hatched: Definitely, this was just copied over from the old GUI template and will be synced with the bases as mentioned above.

webapp/handlers.py
hatched: Heh I had originally done this but they were required by the base templates. I'll see what we can do about syncing them with jaas.ai and then see what we can remove. 👍

webapp/handlers.py
hatched: The GUI handles all of the URL parsing but I think we might like the slashes removed pre-gui.

webapp/handlers.py
hatched: It was, I'll investigate.

.env
hatched: Updated, and PR created for the ports document.

hatched:
Thanks for the review @nottrobin !

Good find with those images, I'll update those and the rest of them with the same issue.

The looping request has been fixed in develop, so you're right not part of this branch :)

templates/404.html
nottrobin: Sorry what do you mean? jaas.ai doesn't have a `templates/_base` folder. That's what I was suggesting you rename in this case.

But if you mean the markup, I still think that while the GUI and jaas.ai are separate domains/subdomains, the GUI should have error pages that appear consistent with the GUI's look and feel. Maybe that just means a blank page with a message in the middle? Similar to how the loading screen looks. That would be fine.

webapp/handlers.py
nottrobin: See https://github.com/juju/juju-gui/pull/3908#discussion_r269739820. We should ask a designer, but I don't agree that simply using jaas.ai's layout is a good move.

webapp/handlers.py
nottrobin: Well, the GUI handles the URL parsing after the first load. But if you reload, then it's back to Flask, initially. So if you GUI crafts a URL like https://example.com/somewhere/, which works initially, but then you reload, then Flask redirects to remove that slash before allow JS to take over again, it could easily confuse the GUI.

If you're doing URL routing in the GUI SPA, I would imagine you want Flask messing with your URLs as little as possible.

nottrobin: https://jujucharms.com/q/store/?type=charm is an example of a URL with a slash on the end

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Browserify has been replaced with Webpack which also now builds the SCSS files.
 - Update the canvas to use the existing svg sprites.
 - Removed unused svg assets from the build/dist.
+- GUI is now being served by Flask.
```
nottrobin: I'm not sure whether we should keep this CHANGELOG.md. Besically, it implies that this is a versioned package, which I know it used to be.

However, going forward, if it's to follow the pattern of our other websites (which I assume it will), we'll remove any `setup.py` that would imply this might be a package, and slim down `package.json` so it doesn't have any version info, to indicate that this is now purely a live application project without versioning.

In that world, we will be doing releases as often as there are changes, which might be multiple times a day. We might even move to a more automated release process where every merge to master gets pushed straight to at least staging automatically. In that world I don't think it's feasible to keep a changelog. The best way to see changes in that world is probably just to look at the list of closed pull requests.

Thoughts?

hatched: That may be the case for the JAAS release but the Juju core version requires versioned releases and extensive QA on each one.

webapp/handlers.py
hatched: Yeah good call. I'll standardize this between the two parsers.

CHANGELOG.md
nottrobin: Okay yeah let's leave this here for now, and we should discuss down the line how we can best facilitate a decent speed of development while also satisfying what Juju needs.

package.json
```diff
@@ -8,6 +8,16 @@
     "type": "git",
     "url": "https://github.com/juju/juju-gui.git"
   },
+  "scripts": {
+    "build": "webpack --config webpack.dev.js && cp -r node_modules/yui static && rm -rf static/yui/node_modules && ./scripts/version.sh",
+    "clean": "rm -rf static/yui static/build node_modules webapp/__pycache__",
+    "watch": "webpack --config webpack.dev.js --watch",
+    "lint-js": "eslint --rulesdir eslint-rules jujugui/static/gui/src",
```
nottrobin: Would you mind adding:

```
    "lint-py": "flake8 --exclude '*env*,node_modules' && black --line-length 79 --check --exclude '(node_modules/.*|[^/]*env[0-9]?/.*)' .",
    "test-javascript": "jest",
    "test": "yarn run lint-python && yarn run lint-js && yarn run test-javascript"
```

?

nottrobin: https://github.com/canonical-websites/blog.ubuntu.com/blob/master/package.json is a good example here

package.json
```diff
@@ -8,6 +8,16 @@
...
+    "lint": "npm run lint-js",
```
nottrobin: We use `yarn`

package.json
hatched: Done

webapp/template_utils.py
```diff
@@ -0,0 +1,54 @@
+import hashlib
```
nottrobin: This file can now be removed, right?

hatched:
I'm going to stop working on getting this travis build to pass and instead, in a follow-up, move it over to circle.

@nottrobin would you mind taking another look so that this can land?

PR #3921, opened 2019-04-11
Convert the GUI to use the ./run architecture. 

hatched:
Converts the GUI from pyramid to flask.
Adds the ability to run it using ./run.
Adds Juju core dist build system.

Fixes ubuntudesign/juju-squad#542

#### To QA

`./run` and the GUI should work as expected
Bootstrap a new controller `./run exec yarn run build-dist` and then `juju upgrade-gui path/to/gui` and the GUI should work as expected.


squidsoup:
QA okay 👍

huwshimi:
@hatched there are a few things that I assume we can clean up now like removing the Makefile, Vagrant stuff etc. but I guess that can happen at our leisure.

hatched:
@huwshimi Yep you're right. There are still a few targets I'd like to export into scripts from the makefile.

Thanks @huwshimi @squidsoup

PR #3940, opened 2019-04-24
Set login cookie and handle JAAS redirects based on said cookie.

hatched:
Fixes canonical-web-and-design/juju-squad#590

- Switches cookie handling to third party library.
- Removes package-lock.json as we now use yarn.
- Sets `loggedin` cookie when logged into the controller.
- Removes `loggedin` cookie on log out.
- Redirects the user to the JAAS storefront or GUI based on `loggedin` cookie.

#### To QA
- Clear your cookies for the GUI domain.
- visit:
  - `/` it should take you to jaas.ai
  - `/home` it should take you to jaas.ai
  - `/u/hatch/failtester` it should take you to jaas.ai/u/hatch/failtester
  - `/new` should take you to the GUI
- Now log into the GUI and visit:
  - `/` it should take you to the GUI
  - `/home` it should take you to jaas.ai
  - `/u/hatch/failtester` it should take you to the GUI


webapp/blueprint.py
```diff
@@ -7,11 +7,40 @@
     "gui", __name__, template_folder="/templates", static_folder="/static"
 )
 
+JAAS_URL = "https://jaas.ai"
+INDEX = "index.html"
+
+
+def loggedIn():
+    return "loggedin" in flask.request.cookies
```
nottrobin: It occurs to me, the most important audience for these redirects are people who are logged in *currently*, and would expect to still see the GUI after the switch-over. Those people won't have a `loggedin` cookie, as we will have only just released this code, but they will have an `auth_tkt` cookie. So I think we should probably check for the existence of either. Do you agree?

(Also, I think it's likely to surprise people less if you treat `loggedin` as an actual boolean value - `return flask.request.cookies.get('loggedin') == 'true'` or something)

hatched: I had originally thought about including the `auth_tkt` but decided against it because I am trying to avoid adding in legacy patches for something that is only a minor speedbump for the actively logged in users. Users get logged out all the time :) 

I can add this back in if you feel strongly about it.

I've updated the cookie value check to be more explicit in the meantime.

nottrobin: I'm worried about seriously confusing existing users. Currently. how used do you think people are to going to https://jujucharms.com (rather than https://jujucharms.com/new) to access the GUI?

We're changing the structure here quite a bit. On the day we switch over https://jujucharms.com to point to this version of the app, anyone who has been used to going to https://jujucharms.com to access the GUI will now go to https://jaas.ai, and the "log in" button will have disappeared. I think that's significantly jarring, don't you?

I agree that preserving `auth_tkt` is legacy mess, which is unfortunate, but it seems to me that the better user experience *vastly* outweighs that particular downside (as it usually does). Our first responsibility is to the users. Arguable this whole redirect system is legacy mess anyway - one more cookie name in our conditional is the least of our problems =)

gui/src/app/init.js
```diff
@@ -820,16 +821,10 @@ class GUIApp {
...
-    // logged into the charmstore.
     if (this.applicationConfig.gisf) {
-      this._sendGISFPostBack(() => {
-        // We can only request the updated config file after the storefront has
-        // been notified of the users login information.
-        this.reloadConfigFile(document);
-      });
+      // If this is the JAAS GUI then we need to set a cookie to say that we're
+      // logged in to correctly redirect requests to jaas.ai if necessary.
+      cookie.set('loggedin', 'true');
```
nottrobin: Did you consider this format? Is naming cookies lowercase without spaces a common standard? If not I think we should consider it. We should either follow the precedent in this app of lowercase_with_underscores (`auth_tkt`) or come up with a new standard deliberately.

Personally, I would choose lowercase-with-hyphens - so `logged-in`. This [seems like it should work everywhere](https://stackoverflow.com/a/1969339/613540), and is the standard we use for most things that aren't actually programming - URL paths, class names and IDs, etc.

nottrobin:
Happy to consider this out of scope for this PR, but this just occurred to me - presumably there are some URLs that are only available in the GUI, not in bluesbrowser. Is that right?

Apart from private charms, are there any other standard URLs that we should be explicitly whitelisting here on jujucharms.com? So they don't redirect ever?

With regard to private charms, I forgot what we agreed, but I imagine the flow would be something like this:

- A non-logged-in user goes to https://jujucharms.com/my-private-charm, but Flask has no way of knowing it's a private charm at this point
- They get redirected to https://jaas.ai/my-private-charm. Hopefully https://jaas.ai is able to tell it's a private charm?
- They get redirected back to https://jujucharms.com/login?forward-url=my-private-charm or something

Is that about right?

hatched:
Unfortunately due to the original URL requirements there is no way for us to distinguish between GUI specific routes, like a model `u/username/mymodel` and a charm `u/username/mycharm` without the user first logging in and then us querying the various api's.

If a non-logged-in user, or one without permissions, tries to access a private charm url it fails with a 404 as it would if it didn't exist at all. This is to avoid leaking names. So what this means is that you can only view private charms in the GUI after logging in.

nottrobin:
@hatched so does that mean that for a user whose login has expired, if they visit `jujucharms.com/u/username/private` they will get redirected to `jaas.ai/u/username/private` which will 404? And then they'll need to be savvy enough to click "Getting started", and then log in and then go back to their private charm?

That's a horrible journey. Again I think we need to make it clearer how to get to the GUI from jaas.ai. Again, hopefully @clagom can give us some guidance.

hatched:
@nottrobin that's correct. That was the requirement at the time.

clagom:
I had problems with GitHub notifications from this repository for some reasons, without getting any email notifications... anyway, @nottrobin pointed out the discussion so here I am. 
@hatched, @nottrobin and I discussed briefly about this today, thinking that maybe a call to action like "Your models" could help -partially- solve the issue. The main problem here is that I don't think any label could cover all the use cases we have due of the missing login. E.g. "Your models" works for users who are aware of what a model is, new/potential users would probably wonder what it is.
We might start with this for now, instead of "get started" (which is more useful for people who don't know about JAAS), but I think we need to explore other solutions as well. 
I would setup a meeting next week to discuss this part and the implementation of the future login system with also @anthonydillon. I think we need to map out what we need to do, estimante, and plan this work for the next cycle. Thanks

hatched:
Ok the requested changes have been made, this is now ready for QA.

webapp/blueprint.py
```diff
@@ -7,11 +7,48 @@
...
+
+
 @gui.route("/<path:path>")
-def homepage(path=""):
-    return flask.render_template("index.html")
+def entity(path=""):
+    if loggedIn():
+        return flask.render_template(INDEX)
+    else:
+        return flask.redirect(JAAS_URL + path)
```
nottrobin: I believe this needs an extra `/` in it. E.g. with `os.path.join(JAAS_URL, path)`. Or just `f"{JAAS_URL}/{path}"`, or `JAAS_URL + '/' + path`. Your choice.

hatched:
Thanks @nottrobin

PR #3941, opened 2019-04-30
Update prod dockerfile to actually build the app.

hatched:
Updates the docker file to actually build the app for production.

#### To QA
docker build -t prod:test .
docker run -ti -p 8999:80 prod:test

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,13 +10,17 @@ RUN apt-get update && \
     apt-get install curl --yes && \
     curl -sL https://deb.nodesource.com/setup_10.x | bash - && \
     apt-get update && \
-    apt-get install --yes build-essential python3-dev python3-pip git nodejs
+    apt-get install --yes build-essential python3-dev python3-pip
 
 # Import code, install code dependencies
-ADD . .
+ADD templates/ templates/
+ADD static/assets/ static/assets/
+ADD static/build/ static/build/
+ADD static/gui/ static/gui/
+ADD webapp/ webapp/
+ADD requirements.txt requirements.txt
+ADD entrypoint entrypoint
 RUN pip3 install -r requirements.txt
-RUN npm install --global yarn
-RUN yarn build-prod
 
 # Setup commands to run server
 ENTRYPOINT ["./entrypoint"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,10 @@ ADD requirements.txt requirements.txt
 ADD entrypoint entrypoint
 RUN pip3 install -r requirements.txt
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -5,16 +5,18 @@ ENV LANG C.UTF-8
...
+    apt-get install curl --yes && \
+    curl -sL https://deb.nodesource.com/setup_10.x | bash - && \
+    apt-get update && \
+    apt-get install --yes build-essential python3-dev python3-pip git nodejs
 
 # Import code, install code dependencies
 ADD . .
 RUN pip3 install -r requirements.txt
+RUN npm install --global yarn
+RUN yarn build-prod
```
nottrobin: Any building should ideally happen outside of the image itself, because the production image should only contain what it needs to run the application in production. In this case, I don't think we need `node` or `yarn` to actually run the app (as we usually don't with our sites).

So normally, we will expect the person building the image to run e.g. `./run build` first, and the we use an `ADD` instruction inside the image to copy in the built assets.

Could that also work here?

Dockerfile
```diff
@@ -5,16 +5,18 @@ ENV LANG C.UTF-8
 WORKDIR /srv
 
 # System dependencies
-RUN apt-get update && apt-get install --yes python3-pip net-tools
-
-# Set git commit ID
-ARG COMMIT_ID
```
nottrobin: We do still need to pass a revision ID through to the image. Although the more modern way to do this is to use `TALISKER_REVISION_ID`, which avoids us having to write it to a file. E.g.:

https://github.com/canonical-web-and-design/docs.snapcraft.io/blob/b3e01feba8446ce74c9384bae416ffd9a6ce10bd/Dockerfile#L15-L17

Dockerfile
hatched: Yep, I agree, updated.

Dockerfile
```diff
@@ -10,13 +10,17 @@ RUN apt-get update && \
     apt-get install curl --yes && \
     curl -sL https://deb.nodesource.com/setup_10.x | bash - && \
     apt-get update && \
-    apt-get install --yes build-essential python3-dev python3-pip git nodejs
+    apt-get install --yes build-essential python3-dev python3-pip
 
 # Import code, install code dependencies
-ADD . .
+ADD templates/ templates/
```
nottrobin: I think you could do all this in one statement - https://stackoverflow.com/questions/30256386/how-to-copy-multiple-files-in-one-layer-using-a-dockerfile

(Consolidating layers in Docker helps with performance)

nottrobin: This went badly. I retract this suggestion.

PR #3943, opened 2019-05-01
Require protocol be defined in JAAS_API_BASE in the gui config view.

### LINKED ISSUE #3942: templated api values should not include the protocol
`bundleServiceURL="https://" + JAAS_API_BASE + "/bundleservice",` Creates invalid urls  because the protocol is defined in `JAAS_API_BASE`
https://github.com/juju/juju-gui/blob/develop/webapp/blueprint.py#L73

hatched:
Fixes #3942 

This fixes an issue where it would generate invalid URL's if the env value wasn't provided correctly.

#### To QA
Check out the demo link, does the GUI work as expected?


webapp/blueprint.py
```diff
@@ -70,18 +70,18 @@ def config():
             "config.js.jinja",
             apiAddress=JIMM_WSS_URL,
             baseUrl="/",
-            bundleServiceURL="https://" + JAAS_API_BASE + "/bundleservice",
-            charmstoreURL="https://" + JAAS_API_BASE + "/charmstore",
+            bundleServiceURL=JAAS_API_BASE + "/bundleservice",
```
huwshimi: It might be nice to append a protocol to `JAAS_API_BASE` in this file if it doesn't include one in .env. But this could be a followup.

PR #3952, opened 2019-05-09
Add redirects for the docs paths and fix the redirects for the root paths

hatched:


webapp/blueprint.py
```diff
@@ -1,4 +1,5 @@
 import os
+import urllib.parse
```
mitechie: make life easy? 

    from urllib.parse import urljoin

webapp/blueprint.py
```diff
@@ -31,6 +32,11 @@ def root():
         return flask.redirect(JAAS_URL)
 
 
+@gui.route("/docs/<path:path>")
+def docs(path=""):
+    return flask.redirect(urllib.parse.urljoin("https://docs.jujucharms.com", path))
```
mitechie: will this hit any other redirects that help with the version to non-versioned urls as well?

webapp/blueprint.py
```diff
@@ -48,7 +54,10 @@ def root():
 @gui.route("/store")
 @gui.route("/support")
 def jaas():
-    return flask.redirect(JAAS_URL)
+    if flask.request.path == "/home":
+        return flask.redirect(JAAS_URL)
+    else:
+        return flask.redirect(urllib.parse.urljoin(JAAS_URL, flask.request.path))
```
mitechie: are there any tests in the codebase around the urls that we care about to where they end up going?

hatched: There aren't, we should add them though - I'll create a follow-up issue for this

webapp/blueprint.py
hatched: done thanks

Issue #3955, opened 2019-05-13
Actions SVG for credentials on Profile is broken


barrymcgee:
<img width="1440" alt="Screenshot 2019-05-13 at 14 44 06" src="https://user-images.githubusercontent.com/505570/57622639-57c75d80-758e-11e9-8558-db6d6d43befc.png">

barrymcgee:
<img width="166" alt="Screenshot 2019-05-13 at 13 45 40" src="https://user-images.githubusercontent.com/505570/57622663-6746a680-758e-11e9-8b66-e4a5b66dc49e.png">

barrymcgee:
@huwshimi Jeff thinks this is probably related to your recent SVG changes?

PR #3956, opened 2019-05-14
Fix profile context menu icon

huwshimi:
Fixes the context icon on the profile credentials list. Fixes: #3955.


huwshimi:
QA:
- visit /u/<your name>/#credentials
- check that the ... icon in the action column has three clear dots, not like: https://user-images.githubusercontent.com/505570/57622663-6746a680-758e-11e9-8b66-e4a5b66dc49e.png

PR #3967, opened 2019-05-14
Remove old Google Analytics events

huwshimi:
These are being removed so they can be replaced with new events.


huwshimi:
QA:
- Run the GUI
- There shouldn't be any console errors.

PR #3969, opened 2019-05-16
Add component analytics collection/sending

huwshimi:


gui/src/app/init/analytics.js
```diff
@@ -1,75 +1,59 @@
...
+  }
+
+  addComponent(component) {
+    let categories = this.category ? [this.category] : [];
+    categories.push(component._reactInternalFiber.elementType.name);
+    // Return a new analytics instance with the new data. This is so that we
+    // don't overwrite a top level instance with every fork as the analytics is
+    // passed into nested components.
+    return new Analytics({
+      category: categories.join(' > '),
```
hatched: Maybe this should stay as an easier to manipulate data structure, like an array, until the event is actually sent. It would make it a little easier to tweak down the line if we wanted. And arguably a micro perf upgrade ;)

gui/src/app/init/analytics.js
```diff
@@ -1,75 +1,59 @@
...
-    valueStr = valueArr.join('|');
-
-    // Emit a google tag manager event registering the state change.
-    dataLayer.push({
+  sendEvent(action, value) {
+    if (!action) {
+      console.error('Analytics action not provided');
+      return;
+    }
+    let label = this.label;
```
hatched: `label` should probably be a setter, which every time you pass a label to it, it appends to the end. You'll see this type of logging in golang error messages:

`cannot fetch foo; cannot parse json; some other chained message`

huwshimi:
> If you are required to `addComponent` in every call then you might as well just have it as an argument, positional or otherwise, in the `sendEvent` call.

You only addComponent when you want to add something to the hierarchy, which might not be the component you sendEvent from (e.g. the sendEvent might be in a button component in which you don't necessarily care about).

> 
> ```js
> this.props.analytics.sendEvent(this, analytics.commonEvents.VIEW);
> ```
> 
> I'm also not keen on having to include a separate lib for the common events, those should be statics on the instance.

Good call.

> 
> ```js
> const analytics = this.props.analytics;
> analytics.sendEvent(this, analytics.VIEW);
> ```
> 
> Having to do this setup everywhere in the component _might_ get tedious if it turns out that it needs to be done in multiple places frequently.
> 
> ```js
> componentDidMount() {
>   this.analytics = this.props.analytics.addComponent(this);
> }
> 
> somemethod() {
>   this.analytics.sendEvent(this.analytics.VIEW);
> }
> ```

Nice. I think it's flexible enough that you can do that where it makes sense.

> 
> Then adding to the label while sending the event
> 
> ```js
>   this.analytics.sendEvent(this.analytics.VIEW, 'cs:my-awesome-charm');
>   // label value = model committed; user authenticated; cs:my-awesome-charm
> ```

I like it, I'll just have to figure out a way to make that flexible enough to work in the case where this is not being used inside a component e.g. on the canvas.

gui/src/app/init/analytics.js
huwshimi: I did have that, but I went for uniformity of flexibility as it didn't seem to give us much (just made the code more complicated).

gui/src/app/init/analytics.js
huwshimi: hmm... this needs to cover cases when you call sendEvent from not inside components so I'm a bit hesitant to make this that specific.

huwshimi:
QA:
- pull down this branch
- `console.log(event);` here: https://github.com/juju/juju-gui/pull/3969/files#diff-b56ece7a328af261e599c4e03a3b0d4bR82
- then `./run`
- click around (plus circle, inspector config, search series filter etc.)
- you should get nice events in the console

gui/src/app/init/analytics.js
```diff
@@ -1,75 +1,88 @@
...
+    something like: Charmbrowser > Search Results > Filter
+    @param {Object|String} categoryOrComponent - A category name or React
+      component.
+    @returns {Object} A new Analytics instance including the old and new data.
+  */
+  addCategory(categoryOrComponent) {
+    let name;
+    if (typeof categoryOrComponent === 'string') {
+      name = categoryOrComponent;
+    } else {
```
hatched: Do we want to be extra resilient here and check for the react prop?

gui/src/app/init/analytics.js
```diff
@@ -1,75 +1,88 @@
...
-  };
-};
+      'eventValue': options.value
+    };
+    this.dataLayer.push(event);
+  }
+}
 
-module.exports = {sendAnalyticsFactory};
+module.exports = {
```
hatched: You can just export the class, then you don't need `{}` on the imports
```javascript
module.exports = Analytics;
```

gui/src/app/init.js
```diff
@@ -7,7 +7,7 @@ const ReactDOM = require('react-dom');
 const mixwith = require('mixwith');
 
 const acl = require('./store/env/acl');
-const analytics = require('./init/analytics');
+const {Analytics} = require('./init/analytics');
```
hatched: See comment below about export.

gui/src/app/init/analytics.js
```diff
@@ -1,75 +1,88 @@
...
+    Set up the analytics instance.
+    @param {String} dataLayer - The Google Tag Manager dataLayer instance.
+    @param {Object} data - Optional parameters:
+      - categories: an array of categories;
+      - getLabel: a function that will be used to generate the label string.
+  */
+  constructor(dataLayer, data={}) {
+    this.dataLayer = dataLayer;
+    this.categories = data.categories || [];
+    this.getLabel = data.getLabel;
```
hatched: `getLabel` is a bit weird....you can only really set it on the initial instance at the app level. I could see some use case where you may want to have multiple `getLabel` calls as you traverse down the tree.

Maybe if we want to leave it as-is for now it should be renamed like 'globalLabels' or something. Something to indicate that it'll be prepended to every event.

gui/src/app/init/analytics.js
```diff
@@ -1,75 +1,88 @@
...
+    Append a category. This will be displayed in the analytics as
+    something like: Charmbrowser > Search Results > Filter
+    @param {Object|String} categoryOrComponent - A category name or React
+      component.
+    @returns {Object} A new Analytics instance including the old and new data.
+  */
+  addCategory(categoryOrComponent) {
+    let name;
+    if (typeof categoryOrComponent === 'string') {
+      name = categoryOrComponent;
```
hatched: We should validate that it doesn't contain the string ">" as that could throw off parsing of event trees.

gui/src/app/state/state.js
```diff
@@ -440,6 +442,10 @@ class State {
     this._appStateHistory.push(purgedState);
     this._pushState();
 
+    this.analytics.addCategory('Navigation').sendEvent('State change', {
```
hatched: Nice 👍

gui/src/app/init/analytics.js
huwshimi: Do you think we should replace it with something else?

PR #3970, opened 2019-05-17
Enable tag manager

huwshimi:
Enable the tag manager for prod. Fixes: #3966.

### DIFF
--- a/templates/config.js.jinja
+++ b/templates/config.js.jinja
@@ -7,7 +7,7 @@ var juju_config = {
     "container": "#main",
     "controllerSocketTemplate": "{{controllerSocketTemplate}}",
     "gisf": {{gisf}},
-    "GTM_enabled": {{ GTM_enabled }},
+    "GTM_enabled": {{GTM_enabled}},
     "interactiveLogin": true,
     "jujuEnvUUID": "{{uuid}}",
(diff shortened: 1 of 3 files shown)


huwshimi:
QA:
- get the output of `juju_config.GTM_enabled` in your browser console
- on the demo service this should be `false`
- if you change `FLASK_DEBUG=false` in .env then `juju_config.GTM_enabled` should become `true`

templates/config.js.jinja
```diff
@@ -7,6 +7,7 @@ var juju_config = {
     "container": "#main",
     "controllerSocketTemplate": "{{controllerSocketTemplate}}",
     "gisf": {{gisf}},
+    "GTM_enabled": {{ GTM_enabled }},
```
hatched: probably should use consistent spacing with the curlies

webapp/blueprint.py
```diff
@@ -104,6 +105,7 @@ def config():
             controllerSocketTemplate="wss://$server:$port/api",
             flags="{terminal: true, support: true, anssr: true, expert: true}",
             gisf="true",
+            GTM_enabled="false" if FLASK_DEBUG == "true" else "true",
```
hatched: 😮

PR #3971, opened 2019-05-17
Add analytics to the header

huwshimi:
Add analytics events to the GUI header. Fixes: #3959.

### DIFF
--- a/gui/src/app/init.js
+++ b/gui/src/app/init.js
@@ -216,6 +216,12 @@ class GUIApp {
       Generated send analytics method. Must be setup before state is set up as
       it is used by state and relies on the controllerAPI instance.
     */
+    if (!window.dataLayer) {
+      // If the dataLayer is not available (e.g. when in development mode) then
+      // store the events in our own array. This is useful for debugging the
+      // events.
+      window.dataLayer = [];
+    }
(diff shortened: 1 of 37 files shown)


huwshimi:
QA:
- do some different things with the header (export, open the model switcher, open profile/help etc.)
- in your console look at the contents of the `dataLayer` variable
- there should be sensible looking events for the header.

gui/src/app/init.js
```diff
@@ -216,6 +216,12 @@ class GUIApp {
       Generated send analytics method. Must be setup before state is set up as
       it is used by state and relies on the controllerAPI instance.
     */
+    if (!window.dataLayer) {
+      // If the dataLayer is not available (e.g. when in development mode) then
+      // store the events in our own array. This is useful for debugging the
+      // events.
+      window.dataLayer = [];
```
hatched: Because this would usually indicate a problem, we should add a console log here to indicate that this is not available and that it's using a 'local' store or something.
```suggestion
      console.log('dataLayer not available, using local array.');
      window.dataLayer = [];
```

huwshimi: I think this will actually only exist on jaas.ai. So dev and embedded would both output that to a console.

Maybe I should leave a comment to that affect. Would that be sufficient?

PR #3972, opened 2019-05-17
Add profile analytics

huwshimi:
Add analytics to the profile. Fixes: #3958.


huwshimi:
QA:
- do some different things in the profile (open the different lists, add things to the canvas, add creds, create a new model etc.)
- in your console look at the contents of the `dataLayer` variable
- there should be sensible looking events for the profile.

PR #3973, opened 2019-05-18
Add canvas analytics

huwshimi:
Fixes: #3957.

### DIFF
--- a/gui/src/app/init/topology/relation.js
+++ b/gui/src/app/init/topology/relation.js
@@ -31,9 +31,10 @@ const RelationMenu = require('../../components/relation-menu/relation-menu');
  * @class RelationModule
  */
 class RelationModule {
-  constructor() {
+  constructor(options={}) {
     this.name = 'RelationModule';
     this.relations = [];
+    this.analytics = options.analytics.addCategory('Relation');
     this.events = {
       scene: {
         '.rel-indicator': {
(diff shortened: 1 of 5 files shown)


gui/src/app/init/topology/relation.js
```diff
@@ -31,9 +31,10 @@ const RelationMenu = require('../../components/relation-menu/relation-menu');
  * @class RelationModule
  */
 class RelationModule {
-  constructor() {
+  constructor(options={}) {
```
hatched: driveby fix?

gui/src/app/init/topology/relation.js
```diff
@@ -648,6 +650,7 @@ class RelationModule {
     // If we landed on a rect, add relation, otherwise, cancel.
     if (rect) {
       self.ambiguousAddRelationCheck(endpoint, self, rect);
+      this.analytics.sendEvent(self.analytics.ADD);
```
hatched: should `this` be `self`? 

doesn't this need a label or a category?

gui/src/app/init/topology/relation.js
huwshimi: This is require to pass in the analytics instance below (`this.analytics = options.analytics.addCategory('Relation');`).

gui/src/app/init/topology/relation.js
huwshimi: This will result in `Canvas > Relation` action: Add. Do you think it needs something else?

hatched: Nope that will be fine

PR #3975, opened 2019-05-18
Add analytics to the store

huwshimi:
Fixes: #3974.

### DIFF
--- a/gui/src/app/components/charmbrowser/charmbrowser.js
+++ b/gui/src/app/components/charmbrowser/charmbrowser.js
@@ -25,7 +25,7 @@ class Charmbrowser extends React.Component {
     // Setting a default state object.
     var state = this.generateState(this.props);
     this.state = state;
-    this.analytics = this.props.analytics.addCategory('Charmbrowser');
+    this.analytics = this.props.analytics.addCategory('Store');
   }
 
   /**
--- a/gui/src/app/components/charmbrowser/charmbrowser.js
+++ b/gui/src/app/components/charmbrowser/charmbrowser.js
@@ -25,6 +25,7 @@ class Charmbrowser extends React.Component {
     // Setting a default state object.
     var state = this.generateState(this.props);
     this.state = state;
+    this.analytics = this.props.analytics.addCategory('Charmbrowser');
   }
 
   /**
@@ -101,6 +102,7 @@ class Charmbrowser extends React.Component {
       case 'store':
         activeChild = (
           <Store
+            analytics={this.analytics}
             changeState={changeState}
             charmstoreURL={this.props.charmstoreURL}
             gisf={this.props.gisf}
@@ -115,7 +117,7 @@ class Charmbrowser extends React.Component {
           <SearchResults
             acl={this.props.acl}
             addToModel={this.props.addToModel}
-            analytics={this.props.analytics.addCategory(this)}
+            analytics={this.analytics}
             changeState={changeState}
             charmstoreSearch={this.props.charmstore.search}
             generatePath={appState.generatePath.bind(appState)}
@@ -138,6 +140,7 @@ class Charmbrowser extends React.Component {
           <EntityDetails
             acl={this.props.acl}
             addNotification={this.props.addNotification}
+            analytics={this.analytics}
             changeState={changeState}
             charmstore={shapeup.fromShape(charmstore, EntityDetails.propTypes.charmstore)}
             clearLightbox={this.props.clearLightbox}
--- a/gui/src/app/components/copy-to-clipboard/copy-to-clipboard.js
+++ b/gui/src/app/components/copy-to-clipboard/copy-to-clipboard.js
@@ -19,8 +19,10 @@ class CopyToClipboard extends React.Component {
   componentDidMount() {
     var node = ReactDOM.findDOMNode(this).querySelector('button');
     this.clipboard = new Clipboard(node, {
-      target(trigger) {
-        return trigger.previousElementSibling;
+      target: trigger => {
+        this.props.analytics.addCategory(this).sendEvent(
+          this.props.analytics.CLICK, {label: `entity: ${this.props.value}`});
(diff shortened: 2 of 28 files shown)


gui/src/app/components/copy-to-clipboard/copy-to-clipboard.js
```diff
@@ -19,8 +19,10 @@ class CopyToClipboard extends React.Component {
   componentDidMount() {
     var node = ReactDOM.findDOMNode(this).querySelector('button');
     this.clipboard = new Clipboard(node, {
-      target(trigger) {
-        return trigger.previousElementSibling;
+      target: trigger => {
+        this.props.analytics.addCategory(this).sendEvent(
+          this.props.analytics.CLICK, {label: `entity: ${this.props.value}`});
```
anthonydillon: Entity is not clear here. Also could the label describe where the input is and the value. This would help to determine which ones being used.

gui/src/app/components/expert-contact-card/expert-contact-card.js
```diff
@@ -24,6 +24,8 @@ class ExpertContactCard extends React.Component {
   */
   _showContact() {
     this.setState({showContact: true});
+    this.props.analytics.addCategory(this).addCategory('Show Contact Details').sendEvent(
```
anthonydillon: Did you mean to nest addCategory here?

gui/src/app/components/entity-details/content/content.js
```diff
@@ -256,14 +256,17 @@ class EntityContent extends React.Component {
...
       hash: null,
       search: {
-        tags: e.target.getAttribute('data-id'),
+        tags: tag,
         text: ''
       },
       store: null
     });
+    this.props.analytics.addCategory('Tag').sendEvent(
+      this.props.analytics.CLICK, {label: `tag: ${tag}`});
```
anthonydillon: Don't think we need the prefix of tag: when setting it to the tag category.

gui/src/app/components/entity-details/header/header.js
```diff
@@ -24,10 +24,11 @@ class EntityHeader extends React.Component {
     const props = this.props;
     const entityModel = props.entityModel;
     const entity = entityModel.toEntity();
+    let plan;
```
hatched: can you add a default value here

gui/src/app/components/entity-details/header/header.js
```diff
@@ -60,6 +68,7 @@ class EntityHeader extends React.Component {
       hash: null,
       store: null
     });
+    this.props.analytics.sendEvent(this.props.analytics.CLOSE);
```
hatched: At this point in this file it's not clear what the category will be on this. Is this a bug? Should we maybe require that each component has a top level category defined IN the component file?

gui/src/app/components/search-results/search-results.js
```diff
@@ -489,6 +493,7 @@ class SearchResults extends React.Component {
     let state = this.state;
     state.showCommunity = !this.state.showCommunity;
     this.setState(this._generateState(state));
+    this.analytics.addCategory('Toggle Community').sendEvent(this.props.analytics.VIEW);
```
hatched: maybe add a label here to show if it's opening or closing?

gui/src/app/components/charmbrowser/charmbrowser.js
```diff
@@ -25,6 +25,7 @@ class Charmbrowser extends React.Component {
     // Setting a default state object.
     var state = this.generateState(this.props);
     this.state = state;
+    this.analytics = this.props.analytics.addCategory('Charmbrowser');
```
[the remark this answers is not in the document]
huwshimi: Done.

gui/src/app/components/copy-to-clipboard/copy-to-clipboard.js
```diff
@@ -19,8 +19,10 @@ class CopyToClipboard extends React.Component {
   componentDidMount() {
     var node = ReactDOM.findDOMNode(this).querySelector('button');
     this.clipboard = new Clipboard(node, {
-      target(trigger) {
-        return trigger.previousElementSibling;
+      target: trigger => {
+        this.props.analytics.addCategory(this).sendEvent(
```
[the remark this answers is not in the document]
huwshimi: Passing `this` means that the category will use the component name, in this case `CopyToClipboard`.

gui/src/app/components/expert-contact-card/expert-contact-card.js
huwshimi: This would result in something like `Store > Entity Details > ExpertContactCard > Show Contact Details`. Chaining the `addCategory` calls adds a new section to that list.

gui/src/app/components/entity-details/content/content.js
huwshimi: I've been adding the prefix because there are also some global details appended to the label. The result in this case would be something like: `model committed: false, user authenticated: true, tag: ops1`. The tag prefix would help identify that bit of data.

gui/src/app/components/entity-details/header/header.js
huwshimi: Like `null`?

gui/src/app/components/entity-details/header/header.js
huwshimi: I was trying to think about what would make sense in the data. So in this case it would result in `Store > Entity Details` and then action: `Close`. But happy to add Header into that chain if you think it's clearer.

gui/src/app/components/search-results/search-results.js
huwshimi: Oops, that's a bug, I meant to only send it when those are shown.

hatched:
QA Issue:

`eventLabel: "model committed: true, user authenticated: true, query: apache filters: [object Object]"`

PR #3976, opened 2019-05-18
Add analytics to the status

huwshimi:
Fixes: #3962.

### DIFF
--- a/gui/src/app/components/status/status.js
+++ b/gui/src/app/components/status/status.js
@@ -13,6 +13,14 @@ require('./_status.scss');
 
 /** Status React component used to display Juju status. */
 class Status extends React.Component {
+  constructor(props) {
+    super(props);
+    this.analytics = this.props.analytics.addCategory('Status');
+  }
+
+  componentDidMount() {
+    this.analytics.sendEvent(this.props.analytics.VIEW);
+  }
 
   /**
     Generate the state to navigate to an application.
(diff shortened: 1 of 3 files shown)


gui/src/app/components/status/status.js
```diff
@@ -43,6 +51,8 @@ class Status extends React.Component {
     // Navigate to the app in the inspector, clearing the state so that the
     // app overview is shown.
     this.props.changeState(this._generateApplicationClickState(appId));
+    this.analytics.addCategory('Application').sendEvent(
```
anthonydillon: It will be difficult to know "Application" is part of the Status component. Could this be called "Application status" or something?

gui/src/app/components/status/status.js
```diff
@@ -93,6 +105,8 @@ class Status extends React.Component {
     evt.stopPropagation();
     evt.preventDefault();
     this.props.changeState(this._generateMachineClickState(machineId));
+    this.analytics.addCategory('Machine').sendEvent(
```
anthonydillon: Same for these

gui/src/app/components/status/status.js
huwshimi: This will be have "Status" add from the call above (`this.analytics = this.props.analytics.addCategory('Status');`) so the final path would be "Status > Application". Same for the events below.

PR #3977, opened 2019-05-19
Add machine view analytics

huwshimi:
Fixes: #3963.


gui/src/app/components/machine-view/add-machine/add-machine.js
```diff
@@ -48,7 +53,9 @@ class MachineViewAddMachine extends React.Component {
       return;
     }
     const selectedMachine = state.selectedMachine;
+    let constraints;
```
hatched: lets set default values for `constraints` and `series`

huwshimi: To `null`?

PR #3978, opened 2019-05-19
Add inspector analytics

huwshimi:
Fixes: #3960.

### DIFF
--- a/gui/src/app/components/inspector/expose/expose.js
+++ b/gui/src/app/components/inspector/expose/expose.js
@@ -11,6 +11,15 @@ const InspectorExposeUnit = require('./unit/unit');
 require('./_expose.scss');
 
 class InspectorExpose extends React.Component {
+  constructor(props) {
+    super(props);
+    this.analytics = this.props.analytics.addCategory('Expose');
+  }
+
+  componentDidMount() {
+    this.analytics.sendEvent(this.props.analytics.VIEW);
+  }
+
   /**
     The callable to be passed to the unit items for navigating to the unit
     details.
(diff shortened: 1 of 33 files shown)


gui/src/app/components/inspector/expose/expose.js
```diff
@@ -29,6 +38,7 @@ class InspectorExpose extends React.Component {
         }
       }
     });
+    this.analytics.addCategory('Unit').sendEvent(this.props.analytics.CLICK);
```
hatched: maybe this should be 'Unit Details'?

gui/src/app/components/inspector/relations/relations.js
```diff
@@ -29,6 +34,7 @@ class InspectorRelations extends React.Component {
           id: this.props.service.get('id'),
           activeComponent: 'relate-to'
         }}});
+    this.analytics.addCategory('Build A Relation').sendEvent(this.props.analytics.CLICK);
```
hatched: Maybe just 'Build' or 'Create' ?

gui/src/app/init/relation-utils.js
```diff
@@ -531,7 +531,7 @@ function getAvailableEndpoints(
     return relatableEndpoints;
   }
   // Filter out the existing relations
-  return relatableEndpoints.filter(relation =>
+  return relatableEndpoints && relatableEndpoints.filter(relation =>
```
hatched: What's up with this? Should we add a test for this issue?

huwshimi: This was a bug that I found while testing the events. I'd add a test, but the tests require YUI, so they're no able to be run.

hatched: Ok that's fine you can leve it as is

PR #3979, opened 2019-05-19
Add analytics to the deployment flow

huwshimi:
Fixes: #3961.

### DIFF
--- a/gui/src/app/components/deployment-flow/cloud/cloud.js
+++ b/gui/src/app/components/deployment-flow/cloud/cloud.js
@@ -92,7 +92,7 @@ class DeploymentCloud extends React.Component {
           onClick={() => {
             this.props.setCloud(cloud);
             this.props.analytics.addCategory('Choose cloud').sendEvent(
-              this.props.analytics.CLICK, {label: `cloud: ${cloud.name}`});
+              this.props.analytics.UPDATE, {label: `cloud: ${cloud.name}`});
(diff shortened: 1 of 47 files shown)


gui/src/app/components/deployment-flow/cloud/cloud.js
```diff
@@ -89,7 +89,11 @@ class DeploymentCloud extends React.Component {
         <li
           className={classes}
           key={cloud.name}
-          onClick={this.props.setCloud.bind(null, cloud)}
+          onClick={() => {
+            this.props.setCloud(cloud);
+            this.props.analytics.addCategory('Cloud').sendEvent(
+              this.props.analytics.UPDATE, {label: `cloud: ${cloud}`});
```
anthonydillon: It might be clearer when we analyse the events of the category is called "Set cloud" and the label is just the cloud

gui/src/app/components/deployment-flow/sshkey/sshkey.js
```diff
@@ -178,6 +179,8 @@ class DeploymentSSHKey extends React.Component {
       this.refs.launchpadUsername.setValue(null);
       this.refs.launchpadUsername.focus();
     }
+    this.props.analytics.sendEvent(
```
anthonydillon: Can you not use `this.analytics` assigned above?

Issue #3984, opened 2019-05-21
Can't add credentials


huwshimi:
I don't seem to be able to add new credentials from either the profile or deployment flow. When I click the add button it does nothing (no errors).

Also, I'm sure this has been fixed already, but I'm seeing broken buttons on the form again:
<img width="660" alt="Screen Shot 2019-05-21 at 5 50 48 am" src="https://user-images.githubusercontent.com/361637/58069287-89db4f80-7b8c-11e9-976a-a71905d4b3cb.png">

PR #3987, opened 2019-05-21
Fix adding credentials

huwshimi:
Fixed adding credentials. Fixed scaling up in the inspector. Fix button layout in the credentials form. Fixes: #3984.


huwshimi:
QA:
- open the credential list in your profile
- open the add credentials form, the buttons should not be squashed (like https://github.com/juju/juju-gui/issues/3984)
- add a credential, it should be created
- open the inspector
- go to the unit scale up form
- add some units
- the form should submit

PR #4001, opened 2019-07-03
Ensure that the flags config value is being passed through and set the default value for the jujushellURL.

### LINKED ISSUE #3999: Juju shell icon disappeared from the canvas
The icon for the Juju shell disappeared.

Before

After

@hatched has another bug, as he can see the icon, but then the app breaks when opening the CLI.

hatched:
Fixes #3999 

To QA:
- connect to a model.
- click the shell icon in the header.
- the shell should open in the footer.


anthonydillon:
@hatched hitting errors with the the analytics object when opening the shell.

hatched:
Thanks @anthonydillon

PR #4014, opened 2019-08-12
Upgrade jaaslib to remove the request for `revision-info` on Charmstore requests.

### LINKED ISSUE #4013: Loading revision-info for bundles and charms
We are still loading `revision-info` from the API which is causing the server to struggle. We need to remove that include from the request and remove revision information from the templates.

hatched:
Fixes #4013 

#### To QA:
- Visit a charm details page and look at the request. It should not have `revision-info` in the url.
- The details page should work as expected. It should not have a link for 'latest version'

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -119,6 +119,11 @@
     lodash "^4.17.11"
     to-fast-properties "^2.0.0"
 
+"@canonical/jaaslib@^0.6.0":
+  version "0.6.0"
+  resolved "https://registry.yarnpkg.com/@canonical/jaaslib/-/jaaslib-0.6.0.tgz#7b77ff0a6899a2e58adc1b2077d8971de019047b"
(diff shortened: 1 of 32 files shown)


yarn.lock
```diff
@@ -5382,7 +5387,7 @@ istanbul-reports@^1.5.1:
   dependencies:
     handlebars "^4.0.3"
 
-jaaslib@0.5.0, jaaslib@^0.5.0:
+jaaslib@^0.5.0:
```
nottrobin: This would suggest something still depends on plain old `jaaslib`, no?

hatched: yessir the juju-react-components lib does. It's on deck to get an upgrade :D 

Good catch 😉

hatched:
Thanks @huwshimi

